PR #728, opened 2020-02-19
Problem: 'hctl shutdown' uses wrong hostname

### LINKED ISSUE #704: Problem: 'hctl shutdown' uses wrong hostname
Context:
- LNet was not configured.
- [For want of](https://en.wikipedia.org/wiki/For_Want_of_a_Nail) LNet, hax failed.
- For want of hax, RC Leader didn't get elected. (“RC leader” Consul session requires both hax and confd to be active on any Consul server node.)
- For want of the RC Leader, `hctl bootstrap` did not terminate.

---

`hctl shutdown` command, executed while RC leader election is going on, tries to ssh to nonexistent
`elect<number>` host:
```
[root@eos-democ-161 ~]# consul kv get leader
elect18098
[root@eos-democ-161 ~]# hctl shutdown
Stopping hare-consul-agent at eos-democ-161.mero.colo.seagate.com... done
Stopping hare-consul-agent at eos-democ-162.mero.colo.seagate.com... done
Killing RC Leader at elect18098... ssh: Could not resolve hostname elect18098: Name or service not known
**ERROR**
```

Solution: if the value, returned by `consul kv get leader`, matches `^elect[0-9]+$` regexp, assume the RC leader
has not been elected and make no attempt to kill it.

typeundefined:
If no RC leader was elected, `hctl shutdown` fails.

Solution: analyze `leader` key in Consul KV and don't try to kill the RC
if the leader name looks like a stub (e.g. elect18098)

Closes #704

### DIFF
--- a/utils/hare-shutdown
+++ b/utils/hare-shutdown
@@ -3,18 +3,15 @@
 # :help: stop the cluster
 
 import os
+import re
 from socket import gethostname
 from typing import Dict, List, NamedTuple
 
 from consul import Consul
 
-Process = NamedTuple('Process', [
-    ('node', str),
-    ('consul_name', str),
-    ('systemd_name', str),
-    ('fidk', int),
-    ('status', str)
-])
+Process = NamedTuple('Process', [('node', str), ('consul_name', str),
+                                 ('systemd_name', str), ('fidk', int),
+                                 ('status', str)])
 shutdown_sequence = ('s3service', 'ios', 'confd', 'hax', 'consul')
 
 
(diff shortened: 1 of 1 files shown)


andriytk:
Why does it fail if there is not RC-leader?

knekrasov:
@andriytk Just look into this piece of code:

https://github.com/Seagate/cortx-hare/blob/b43c3620efdcafb75c6f110ccf2fa0e0342c1a90/utils/hare-shutdown#L99-L115

Initially last 3 lines looked like this:
```python
    if leader_node:
        print(f'Killing RC Leader at {leader_node}... ', end='', flush=True)
        exec(ssh_prefix(leader_node) + 'sudo pkill --exact -KILL consul')
```

ssh'ing to a garbage value `elect1234` would obviously fail. Because not all values of `leader` key mean a hostname.

andriytk:
I see. The right solution would probably be to check for session on the leader key: if the session exists (which means the RC leader is elected), then the value is a valid hostname.

PR #1287, opened 2020-09-10
Problem: Hax service stuck while stopping rconfc

rajanikantchirmade:
While shutting down cluster, Hax service stops after IOS & CONFD
stops. In this case when Hax tried to stop rconfc, it stuck in
confd election, since confds already shutdown it should not go
for election and mark as rconfc_failed instead.

Solution:
Currently rconfc only used to fetch filesystem stats, and it only
fetch stats if all Motr services are running. fs-stats-updater thread
start rconfc only when all Motr services are running. So whenever it
it finds that all services are not running it stops rconfc and again
start when all services up and running again.


andriytk:
What if a shutdown happens during fs-status fetch when rconfc is not stopped yet?..

PR #1347, opened 2020-09-30
Problem: S3 resources do not stop when motr stack migrated to another node

azheregelya:
This makes s3server to fail and produce a lot of traces which consume
disk space on shared partition affecting another node.
ClusterIP resource shall also migrate after motr-stack since "failed"
node can't serve requests anymore.

Solution:
* Add hax-running attribute which is set to 1 when hax is
  started on current node and set to 0 when hax is stopped (Exec Post-hooks
  in systemd file).
* Add location constrains for ClusterIP-clone and s3auth-clone resources
  to make them migrate/stop when hax (i.e. motr-stack) is not running on
  current node. Since s3server-cN resources are colocated with s3auth
  and can't work on opposite nodes - they will be stopped on "failed"
  node.

Signed-off-by: Andrei Zheregelia <andrei.zheregelia@seagate.com>

### DIFF
--- a/utils/build-ees-ha
+++ b/utils/build-ees-ha
@@ -442,12 +442,16 @@ hax_systemd_update() {
     sudo sed -e 's/hare-consul-agent.service/hare-consul-agent-c1.service/' \
              -e "/ExecStart=/iExecStartPre=/bin/sh -c 'mountpoint /var/mero || /bin/mount $lvolume /var/mero'" \
              -e "/ExecStart=/aExecStopPost=/bin/sh -c '! mountpoint /var/mero || while ! /bin/umount /var/mero; do lsof +D /var/mero; sleep 1; done'" \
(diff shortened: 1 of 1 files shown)


azheregelya:
Deploy is required to verify that this patch doesn't break anything.
cc: @NeeravChoudhari @Jyoti-cs

andriytk:
@azheregelya was it cherry-picked from somewhere? If so - could you please mention the sha1 in the commit msg? (Note: `git cherry-pick -x` command does this automatically.)

It might be handy to link it with the original PR also (just by mentioning it in some comment).

Thanks!

andriytk:
Ah, sorry - the previous PR was in another (cortx-ha repo)... Nevermind it then.

andriytk (approved):

mssawant (approved):

PR #1348, opened 2020-09-30
EOS-13678: Consul doesn't always invoke service watcher

knekrasov:
Workaround: don't rely on watch mechanism. Instead, poll the service
health status directly from Consul via Health API.


andriytk:
@knekrasov , could you please add a little bit more context into the description? Thanks.

andriytk:
@vvv, @chumakd it would be nice if you could take a look at the Python code here (style, etc.)

andriytk:
@knekrasov , could you please add the context info into the commit msgs also? As we are in the very beginning of the review - it is Ok to force push the updated commit here.

andriytk:
> Problem: consul sends useless requests to hax with serice statuses …

I'd suggest either to merge this patch with the previous one or explain in its commit-msg why it is useless. Otherwise, it's not very clear.

hax/hax/servicemon.py
```diff
@@ -0,0 +1,87 @@
...
+                    delta: List[HAState] = []
+
+                    for name in service_names:
+                        health: HAState = self.consul.get_local_service_health(
+                            name)
+                        if (health.status != known_statuses[name]):
+                            delta.append(health)
+                            known_statuses[name] = health.status
+                            logging.debug('%s is now %s', name, health.status)
+                    self._broadcast(delta)
```
andriytk: Currently, Motr is Ok to receive the full list of all the ha_states when some of the state(s) is/are changed. But maybe it's better to send only the delta indeed and save on the update msg sizes...

andriytk (approved):
LGTM

hax/hax/servicemon.py
knekrasov: @andriytk Please note that this code will send the changed statuses only (otherwise the vector would not be named delta)

andriytk: Yes, I understood that! :)

PR #1352, opened 2020-10-02
Problem: reportbug tool doesn't generate syslog in JSON format

azheregelya:
Syslog in JSON format is required as input for analysis scripts.

Solution: add command to generate reqired log file

Signed-off-by: Andrei Zheregelia <andrei.zheregelia@seagate.com>

### DIFF
--- a/utils/hare-reportbug
+++ b/utils/hare-reportbug
@@ -79,6 +79,9 @@ exec 2> >(tee _reportbug.stderr >&2)
 sudo journalctl --no-pager --full --utc --output short-precise \
      > syslog.txt || true
 
+sudo journalctl --no-pager --full --utc --output=json --unit=pacemaker.service \
+     > syslog-pacemaker.json || true
+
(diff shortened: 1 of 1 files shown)


utils/hare-reportbug
```diff
@@ -79,6 +79,9 @@ exec 2> >(tee _reportbug.stderr >&2)
 sudo journalctl --no-pager --full --utc --output short-precise \
      > syslog.txt || true
 
+sudo journalctl --no-pager --full --utc --output=json --unit=pacemaker.service \
+     > syslog.json || true
```
chumakd, line 83: maybe `s/syslog.json/syslog-pacemaker.json/`?

chumakd (approved):

utils/hare-reportbug
azheregelya: Right. Fixed. Let me squash that fixup

PR #1356, opened 2020-10-07
EOS-13263: Hare utility to mask features and logrotate configuration is missing

mssawant:
- CSM requires Hare to report any unavailable features based on the
   installation type.
- Logrotate settings for hare logs must be configured based on the given installation
   type. Such a utility is missing in hare.
- Cron job for pruning hax m0trace files is also missing based on the installation
   type.

Solution:
    - Add a utility to accept options for configuring logrotate and to report
    masked features based on the installation type fetched from the provisioner.
    Sample provisioner installation information,
    ```
    {
    "ret": {
    "nodes": 1,
    "server_type": "physical",
    "servers_per_node": 2,
    "storage_type": "5u84"
    }
    }
    ```
    - Add a cron job script to prune /var/motr/hax/m0trace.* files based on the
      installation type (virtual or physical).

### DIFF
--- a/provisioning/setup.py
+++ b/provisioning/setup.py
@@ -1,0 +1,99 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+# Setup utility for Hare to configure Hare related settings, e.g. logrotate,
+# report unsupported features, etc.
+
+import argparse
+import asyncio
+import json
+import logging
+import shutil
+import subprocess
+
+from eos.utils.product_features import unsupported_features
+
+
+def get_data_from_provisioner_cli(method, output_format='json') -> str:
+    try:
+        process = subprocess.run(['provisioner', method, f'--out={output_format}'],
+                                 check=True,
+                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+                                 shell=False)
+        stdout = process.stdout
+        rc = process.returncode
+    except Exception as e:
+        logging.error('Failed to fetch data from provisioner (%s)', e)
+        return 'unknown'
+    if rc != 0:
+        return 'unknown'
+    res = stdout.decode('utf-8')
+    return json.loads(res)['ret'] if res else 'unknown'
+
+
+def logrotate_config():
+    try:
+        setup_info = get_data_from_provisioner_cli('get_setup_info')
+        if setup_info != 'unknown':
+            server_type = setup_info['server_type']
+            shutil.copyfile(f'/opt/seagate/cortx/hare/conf/logrotate/{server_type}',
+                            '/etc/logrotate.d/hare')
+    except Exception as error:
+        logging.error('Error setting logrotate values for hare (%s)', error)
+
+
+def _report_unsupported_features(features_unavailable):
+    uf_db = unsupported_features.UnsupportedFeaturesDB()
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(uf_db.store_unsupported_features('hare',
+                            features_unavailable))
+
+
+class UnsupportedFeatures(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        try:
+            features_unavailable = []
+            path = '/opt/seagate/cortx/hare/conf/setup_info.json'
+            with open(path) as hare_features_info:
+                hare_unavailable_features = json.load(hare_features_info)
+                setup_info = get_data_from_provisioner_cli('get_setup_info')
+                if setup_info != 'unknown':
+                    for setup in hare_unavailable_features['setup_types']:
+                        if setup['name'] == setup_info['server_type']:
+                            features_unavailable.extend(
+                                setup['unsupported_features'])
+                            _report_unsupported_features(features_unavailable)
+        except Exception as error:
+            logging.error('Error reporting hare unsupported features (%s)',
+                          error)
+
+
+def main(argv=None):
+    p = argparse.ArgumentParser(description='Configure hare settings')
+    p.add_argument('--report-unavailable-features',
+                   nargs=0,
+                   help='Report unsupported features according to setup type',
+                   action=UnsupportedFeatures)
+    p.parse_args(argv)
+    logrotate_config()
+
+
+if __name__ == '__main__':
+    main()
--- a/utils/m0trace-prune
+++ b/utils/m0trace-prune
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+set -eu -o pipefail
+# set -x
+export PS4='+ [${BASH_SOURCE[0]##*/}:${LINENO}${FUNCNAME[0]:+:${FUNCNAME[0]}}] '
+
+PROG=${0##*/}
+
+usage() {
+cat <<EOF
+Usage: $PROG [<option>]
+
+Prune Hax m0trace files.
+
+Options:
+    -n <nr files> Max number of m0trace files to retain.
+    -h, --help    Display help and exit.
+EOF
+}
+
+TEMP=$(getopt --options hn: \
+              --longoptions help \
+              --name "$PROG" -- "$@" || true)
+
+(($? == 0)) || { usage >&2; exit 1; }
+
+eval set -- "$TEMP"
+
+declare -A default_nr_files_to_prune=(
+    [virtual]=2
+    [physical]=5
+)
+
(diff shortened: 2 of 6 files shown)


knekrasov (approved):

chumakd (approved):

provisioning/setup.py
```diff
@@ -0,0 +1,96 @@
...
+
+def _report_unsupported_features(features_unavailable):
+    async def _process(features_unavailable):
+        loop = asyncio.get_event_loop()
+        uf_db = await loop.run_in_executor(None,
+                    unsupported_features.UnsupportedFeaturesDB)
+        loop = asyncio.get_event_loop()
+        await loop.run_in_executor(None, uf_db.store_unsupported_features,
+                                       str('hare'), features_unavailable)
+    _process
```
knekrasov, line 69: I believe it should be `return _process` instead

provisioning/setup.py
```diff
@@ -0,0 +1,96 @@
...
+        try:
+            features_unavailable = []
+            path = '/opt/seagate/cortx/hare/conf/setup_info.json'
+            with open(path) as hare_features_info:
+                hare_unavailable_features = json.load(hare_features_info)
+                setup_info  = get_data_from_provisioner_cli('get_setup_info')
+                for setup in hare_unavailable_features['setup_types']:
+                    if setup['name'] == setup_info['server_type']:
+                        features_unavailable.extend(setup['unsupported_features'])
+                        _report_unsupported_features(features_unavailable)
```
knekrasov, line 82: perhaps it should be 
```python
await _report_unsupported_features(features_unavailable)
```

But I would just remove all async/await stuff from this code since it doesn't give any obvious benefit.

knekrasov (commented):
Perhaps I've found the reasons why it doesn't work.

provisioning/setup.py
mssawant: I tried that actually, so I added a log in this function which does not get posted so I am not sure if the function is getting executed or not.

knekrasov: I've just added a commit to your branch. I had no ability to check it, but probably it works better now.

provisioning/setup.py
mssawant: Then I get this coroutine warning for the csm functions,
```
/usr/lib64/python3.6/argparse.py:1840: RuntimeWarning: coroutine 'UnsupportedFeatures.__call__' was never awaited
  action(self, namespace, argument_values, option_string)
```
Probably csm function is tagged as async, in order to overcome this warning I started using asyncio. Do you think this warning can be ignored?

mssawant: Was able to fix it.

vvv:
@mssawant wrote:
> I have fixed few of the codacy issues for #1356 but still codacy is [not happy](https://app.codacy.com/gh/Seagate/cortx-hare/pullRequest?prid=6295756), presently its a required check which is failing and I am out of ideas, I would highly appreciate if any of you have any inputs on the errors reported by the codacy. We have another option to temporary disable the codacy check and merge the PR. Would like your opinion on the same. Thank you.

I've opened #1365 for the Codacy problem and [solved it](https://github.com/Seagate/cortx-hare/issues/1365#issuecomment-706560959) by updating ‘main’ branch protection rules.  Now this patch can be merged.

provisioning/setup.py
```diff
@@ -0,0 +1,99 @@
...
+                                 check=True,
+                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+                                 shell=False)
+        stdout = process.stdout
+        rc = process.returncode
+    except Exception as e:
+        logging.error('Failed to fetch data from provisioner (%s)', e)
+    if rc == 0:
+        res = stdout.decode('utf-8')
+        if res != "":
```
vvv, line 47: [PEP 8](https://www.python.org/dev/peps/pep-0008/):
> For sequences, (strings, lists, tuples), use the fact that empty sequences are false

```python
assert not bool('')
```
[optional] Suggestion:
```python
    if rc != 0:
        return 'unknown'
    res = stdout.decode('utf-8')
    return json.loads(res)['ret'] if res else 'unknown'
```

provisioning/setup.py
```diff
@@ -0,0 +1,99 @@
...
+import shutil
+import subprocess
+
+from eos.utils.product_features import unsupported_features
+
+
+def get_data_from_provisioner_cli(method, output_format="json") -> str:
+    rc = 1
+    try:
+        process = subprocess.run(['provisioner', f"{method}", f"--out={output_format}"],
```
vvv, line 37: [nit] Use single-quoted strings for the sake of consistency.

provisioning/setup.py
```diff
@@ -0,0 +1,99 @@
...
+import json
+import logging
+import shutil
+import subprocess
+
+from eos.utils.product_features import unsupported_features
+
+
+def get_data_from_provisioner_cli(method, output_format="json") -> str:
+    rc = 1
```
vvv, line 35: [nit] This assignment is not needed.
```
>>> del var
NameError: name 'var' is not defined
>>> if True:
...     var = 100
... 
>>> var
100
```

provisioning/setup.py
```diff
@@ -0,0 +1,99 @@
...
+            result: str = json.loads(res)['ret']
+            return result
+    return 'unknown'
+
+
+def logrotate_config():
+    try:
+        setup_info = get_data_from_provisioner_cli('get_setup_info')
+        setup_type = setup_info['server_type']
+        shutil.copyfile(f"/opt/seagate/cortx/hare/conf/logrotate/{setup_type}",
```
vvv, line 57: [nit] `s/"/'/g`

utils/m0trace-prune
```diff
@@ -0,0 +1,84 @@
...
+    if [[ $plt = "none" ]]; then
+        echo ${default_nr_files_to_prune[physical]}
+    else
+        echo ${default_nr_files_to_prune[virtual]}
+    fi
+}
+
+max_files_to_retain=${max_files_to_retain:-$(get_max_files_by_platform)}
+
+total_m0trace_files=`find $hax_m0_dir -maxdepth 1 -type f -name "m0trace.*" | wc -l`
```
vvv, line 74: [nit] `$(...)`, single-quotes.
```bash
total_m0trace_files=$(find $hax_m0_dir -maxdepth 1 -type f -name 'm0trace.*' | wc -l)
```

utils/m0trace-prune
```diff
@@ -0,0 +1,84 @@
...
+        -h|--help)   usage; exit ;;
+        -n)          max_files_to_retain=$2; shift 2 ;;
+        --)          shift; break ;;
+        *)           break ;;
+    esac
+done
+
+hax_m0_dir=/var/motr/hax
+get_max_files_by_platform() {
+    plt=$(systemd-detect-virt)
```
vvv, line 63: https://github.com/bahamas10/bash-style-guide#variable-declaration:
> `local` should _always_ be used in functions.

https://github.com/bahamas10/bash-style-guide#quoting:
> Use double quotes for strings that require variable expansion or command substitution interpolation, and single quotes for all others.

Suggestion:
```bash
get_max_files_by_platform() {
    if [[ $(systemd-detect-virt) == none ]]; then
        echo ${default_nr_files_to_prune[physical]}
    else
        echo ${default_nr_files_to_prune[virtual]}
    fi
}
```

provisioning/setup.py
mssawant: Okay.

provisioning/setup.py
mssawant: Used doubles quotes as there is variable dereferencing. Let me try with single quotes.

provisioning/setup.py
mssawant: iIf not initialized I get a following error in case of failure,
```
ERROR:root:Error reporting hare unsupported features (local variable 'rc' referenced before assignment)
```
So changed a bit now returning from exception itself that avoids requirement to initialize rc.

provisioning/setup.py
vvv: @mssawant Python makes no semantical distinction between single-quoted and double-quoted strings.
```
>>> s = 'world'
>>> f'Hello, {s}!'
'Hello, world!'
>>> "{}ds don't come easy".format(s[:3])
"words don't come easy"
```

vvv: @mssawant `f'{method}'` should be replaced with `method`.

provisioning/setup.py
```diff
@@ -0,0 +1,99 @@
...
+import logging
+import shutil
+import subprocess
+
+from eos.utils.product_features import unsupported_features
+
+
+def get_data_from_provisioner_cli(method, output_format="json") -> str:
+    try:
+        process = subprocess.run(['provisioner', f'{method}', f'--out={output_format}'],
```
vvv, line 36: Can `f'{method}'` be replaced with `method`?

provisioning/setup.py
```diff
@@ -0,0 +1,99 @@
...
+    if rc != 0:
+        return 'unknown'
+    res = stdout.decode('utf-8')
+    return json.loads(res)['ret'] if res else 'unknown'
+
+
+def logrotate_config():
+    try:
+        setup_info = get_data_from_provisioner_cli('get_setup_info')
+        if setup_info != 'unknown':
```
vvv, line 54: `setup_info` variable is not needed.
```python
        if get_data_from_provisioner_cli('get_setup_info') != 'unknown':
            ...
```

provisioning/setup.py
```diff
@@ -0,0 +1,99 @@
...
+            setup_type = setup_info['server_type']
```
vvv, line 55: [nit] Shouldn't the variable be named `server_type` as well?

provisioning/setup.py
mssawant: ah you are right, no need to convert it to fstring, thanks.

vvv (approved):
@mssawant There are minor style issues.  Feel free to merge the patch as is.

provisioning/setup.py
mssawant: Yes.

provisioning/setup.py
mssawant: Yes will do that.

provisioning/setup.py
```diff
@@ -0,0 +1,99 @@
...
+import asyncio
+import json
+import logging
+import shutil
+import subprocess
+
+from eos.utils.product_features import unsupported_features
+
+
+def get_data_from_provisioner_cli(method, output_format="json") -> str:
```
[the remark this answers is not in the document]
mssawant: Okay.

provisioning/setup.py
mssawant: But it is referenced on the next line,
> server_type = setup_info['server_type']

vvv: It doesn't have to. 🤓
```python
        if get_data_from_provisioner_cli('get_setup_info') != 'unknown':
            ...
```
See? Equivalent code, no variable.

mssawant: Seems like something terribly wrong with my vision but I am not getting then from where do we get the `server_type` , we we do not save the output of `get_data_from_provisioner_cli`, we invoke `get_data_from_provisioner_cli` again?

vvv: @mssawant No, it's not your vision, it's mine being unfocused.  Sorry for the noise.

PR #1362, opened 2020-10-08
EOS-14115: fix ha-notifications race during bootstrap

andriytk:
During cluster bootstrap with --mkfs option, it is possible
that ha-online notification from Consul about the Motr service
when its m0mkfs process is started will not manage to change
to ha-offline after m0mkfs process has finished. Instead, it
may come after the correspondent m0d process is started already.
This will cause hax to close the ha_link to the newly started
m0d process, so the process would panic:

```
Jan 03 13:02:01.356617 hare64648-build-ssu1 mero-server[5124]: Mero panic: !rep->hae_disconnected_previously at ha_entrypoint_state_cb() ha/ha.c:308 (errno: 0) (last failed: none) [git: 8d44b0a8a] pid: 5124  /var/mero/m0d-0x7200000000000001:0xc/m0trace.5124
Jan 03 13:02:01.356949 hare64648-build-ssu1 mero-server[5124]: Mero panic reason: HA has already decided that this process has failed. There is no good reason to continue doing something, and there is no code yet to handle graceful shutdown in this case. Let's just terminate the process and let HA do it's job.
```

The problem is with how we determine when to start m0d process
after m0mkfs: instead of checking that m0mkfs process is stopped
locally on the node, we should check the service state in Consul
catalogue. Because the ha-state update (which affects ha_link)
happens (via hax) only when the service state is changed in Consul
catalogue. And such a change might be delayed until Consul
synchronises it with all its servers in the cluster.

Solution: check the service state in Consul catalogue and make
sure it became offline before starting m0d process.

### DIFF
--- a/utils/bootstrap-node
+++ b/utils/bootstrap-node
@@ -187,19 +187,33 @@ case $phase in
        ;;
 esac
 
+# It's critical to check for the service state in Consul catalogue.
+# Otherwise, if synchronisation of service status between Consul
+# servers delays (for some reason), hax may get 'offline' state from
+# Consul for already started m0d process (spoiling its ha_link).
+wait_m0mkfs_done()
+{
+    local id=$1
+    local fid=$(id2fid $id)
+
+    touch /tmp/motr-mkfs-pass-$fid
+    # Give time for service check to reset m0mkfs' HA state
+    # (see utils/check-service also).
+    while [[ -f /tmp/motr-mkfs-pass-$fid ||
+             -f /tmp/motr-mkfs-fail-$fid ||
+             $(get-service-health $id) != 'warning' ]]; do
+        sleep 1
+    done
+    # Give hax a chance to get the warning/offline state from Consul
+    # and finish ha_link before m0d process is started.
+    sleep 2
+}
+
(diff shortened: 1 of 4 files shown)


utils/bootstrap-node
```diff
@@ -187,19 +187,33 @@ case $phase in
...
+{
+    local id=$1
+    local fid=$(id2fid $id)
+
+    touch /tmp/motr-mkfs-pass-$fid
+    # Give time for service check to reset m0mkfs' HA state
+    # (see utils/check-service also).
+    while [[ -f /tmp/motr-mkfs-pass-$fid ||
+             -f /tmp/motr-mkfs-fail-$fid ||
+             $(get-service-health $id) != 'warning' ]]; do
```
mssawant, line 204: What if the services status is returned as 'failing'? do we want to change the check to `$(get-service-health $id)  == 'passing'` instead?

andriytk:
Sorry, I accidentally deleted the branch. Please review at #1363.

utils/bootstrap-node
andriytk: Good idea.

PR #1370, opened 2020-10-13
Problem: 'hctl status' fails if cluster is not running

vvv:
"Cluster is not running" situation is normal.  Non-zero exit code of `hctl status` makes impression that something went wrong.

Solution: let `hctl status` exit successfully.

### DIFF
--- a/utils/hare-status
+++ b/utils/hare-status
@@ -205,7 +205,7 @@ def setup_logging():
 
 
 @repeat_if_fails(max_retries=24)
-def show_text_status(cns: Consul):
+def show_text_status(cns: Consul) -> None:
     # In-memory buffer required because an intermittent Consul exception can
     # happen right in the middle of printing something. It is good to postpone
     # the printing to stdout until the moment when those exceptions can't
@@ -238,13 +238,13 @@ def main(argv=None):
     opts = parse_opts(argv)
     cns = Consul()
     if not cluster_online():
-        print('Cluster is not running.', file=sys.stderr)
-        return 1
+        print('Cluster is not running', file=sys.stderr)
+        return
 
     if opts.json:
         status = get_cluster_status(cns)
         print(j.dumps(status, indent=2, cls=FidEncoder))
-        return 0
+        return
     show_text_status(cns)
 
 


rajanikantchirmade:
@vvv  But if someone want to check cluster status  (running/not running) using return code of `hctl status`, it will fail.

`hctl status` return 0 -- Cluster running  
                    return 1 -- Cluster not running.

vvv:
> `hctl status` return 0 -- Cluster running
> return 1 -- Cluster not running.

I don't think we should encourage users to rely on undocumented behaviour.  If we keep it, users will [inevitably](https://www.hyrumslaw.com/) start depending on it.

So the question is, how useful is non-zero exit code of `hctl status` in “cluster not running” situation?

cc @chumakd @andriytk @azheregelya

azheregelya:
If i understand correctly, other components rely on JSON output generated by hctl. Therefore  make sense to stick on this method. 
Non-zero return code from hctl is better to use to indicate that hctl itself failed to retrieve cluster status.

max-seagate:
@vvv: a special switch (off by default) could be added to exit unsuccessfully if the cluster is not running.

utils/hare-status
```diff
@@ -238,13 +238,13 @@ def main(argv=None):
...
-        print('Cluster is not running.', file=sys.stderr)
-        return 1
+        print('Cluster is not running', file=sys.stderr)
+        return
 
     if opts.json:
         status = get_cluster_status(cns)
         print(j.dumps(status, indent=2, cls=FidEncoder))
-        return 0
+        return
```
knekrasov, line 247: Suggestion:

```python
if __name__ == '__main__':
    try:
        sys.exit(main())
    except HAConsistencyException as err:
        print(f'Consul Error: {err}')
        code = 1 if fail_on_consul_issue else 0
        sys.exit(code)
    except Exception as e:
        sys.exit('Unexpected problem appeared: {e}')
```

knekrasov: Unfortunately, Github doesn't allow commenting lines that are outside of the diff (although they may be related to the fix being discussed, huh)

knekrasov (approved):

utils/hare-status
max-seagate: @knekrasov: you can express your opinion about the GitHub issue at https://jts.seagate.com/browse/EOS-11596.

knekrasov: @max-seagate It is closed with "Can't fix this issue." 

Why don't we close our cortx issues with the same explanation? 😄

PR #1371, opened 2020-10-13
Problem: rpm build for hare main branch fails

mssawant:
Hare logrotate configuration is supposed to be installed at
/etc/logrotate.d/hare as part of rpm build but the installation
code is missing from the Makefile. This was added as part of
the commit `0cd9cb4` but was removed by `3483beb` as the logrotate
configuration is required to be installed as per the installation type
(i.e. virtual or physical) using hare setup.py utility as part of
the provisioning. But this is breaking the rpm build presently.

Solution:
Restore /etc/logrotate.d/hare installation logic in Makefile and
install a default logrotate configuration ($HARE_SRC/provisioning/
logrotate/hare) for hare before setup.py utility is invoked to
update it as per the installation type.


knekrasov (approved):

azheregelya (approved):
@chumakd FYI

PR #1376, opened 2020-10-15
Problem: setup.py uses wrong utils package

mssawant:
eos.utils python utils package is renamed to cortx-py-utils. Also
logrotate configuration file for `physical` setup is missing.

Solution:
- Use cortx-utils package instead of eos.utils package.
- Add logrotate configuration file for `physical` setups.
- Update setup_info.json file to use `server_type` instead of name.

### DIFF
--- a/provisioning/logrotate/physical
+++ b/provisioning/logrotate/physical
@@ -1,0 +1,9 @@
+/var/log/hare/*.log
+{
+    rotate 10
+    maxsize 50M
+    weekly
+    compress
+    missingok
+    copytruncate
+}
--- a/provisioning/setup.py
+++ b/provisioning/setup.py
@@ -28,8 +28,7 @@ import logging
 import shutil
 import subprocess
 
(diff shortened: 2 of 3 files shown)


rajanikantchirmade (approved):
Why some one needs change sources for configuring different setup.

I may be some configuration file which describe cluster type and can be somewhere @ /var/lib/hare/cluster_type.json   just like we use CDF.

{
    "cluster_type : physical|virtual"
    "storage_type: PODS|JBOD
}

Is this information not sufficient ?

mssawant:
@rajanikantchirmade, yes we have such file, setup_info.json which is installed as part of installation and referred to based on the setup type, i.e. virtual or physical.  Accordingly we report the functions that are not supported and update the logrotate configuration.

knekrasov (approved):
As long as hax doesn't depend on cortx.utils, I'm ok 😄

PR #1377, opened 2020-10-15
Problem: Hax is not broadcasting the events to other node services

mssawant:
Hax detects motr process failures through service monitor but broadcasts
the events only to local node's motr processes. Due to this motr processes
from peer nodes do not have knowledge about the failure and continue to
use stale connections.
Motr requires for the processes to reconnect on restarts otherwise
subsequent IO fails. Reconnection happens on the respective process stop/
failed/start events.

Solution:
- Use Hare EQ and BQ infrastructure to broadcast process events to all the
  motr processes in the cluster.
- Distinguish between motr-mkfs and m0d process events using process event
  type.

### DIFF
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -19,7 +19,7 @@
 import logging
 import time
 from queue import Empty, Queue
-from typing import List
+from typing import List, Union
 
 from hax.message import (BroadcastHAStates, EntrypointRequest, HaNvecGetEvent,
                          ProcessEvent, SnsRebalancePause, SnsRebalanceResume,
@@ -27,11 +27,45 @@ from hax.message import (BroadcastHAStates, EntrypointRequest, HaNvecGetEvent,
                          SnsRebalanceStop, SnsRepairPause, SnsRepairResume,
                          SnsRepairStart, SnsRepairStatus, SnsRepairStop)
 from hax.motr import Motr
-from hax.queue.publish import EQPublisher
-from hax.types import MessageId, StobIoqError, StoppableThread
+from hax.queue.publish import EQPublisher, BQPublisher
+from hax.types import (ConfHaProcess, HAState, m0HaProcessEvent,
+                       m0HaProcessType, MessageId, ServiceHealth,
+                       StobIoqError, StoppableThread)
 from hax.util import ConsulUtil, dump_json, repeat_if_fails
 
 LOG = logging.getLogger('hax')
+
+
+def broadcast_process_event(q: Queue, event: ConfHaProcess):
+    ha_event = m0HaProcessEvent(event.chp_event)
+    proc_type = m0HaProcessType(event.chp_type)
+    if proc_type == m0HaProcessType.M0_CONF_HA_PROCESS_M0D:
+        status_map = {m0HaProcessEvent.M0_CONF_HA_PROCESS_STOPPED:
+                      ServiceHealth.FAILED,
+                      m0HaProcessEvent.M0_CONF_HA_PROCESS_STARTED:
+                      ServiceHealth.OK}
+        svc_status = status_map.get(ha_event, ServiceHealth.UNKNOWN)
+        if svc_status != ServiceHealth.UNKNOWN:
+            q.put(BroadcastHAStates(states=[HAState(fid=event.fid,
+                                                    status=svc_status)],
+                                    is_broadcast_local=False,
+                                    reply_to=None))
+
+
+def broadcast_local(ha_states: BroadcastHAStates, motr: Motr):
+    result: List[MessageId] = motr.broadcast_ha_states(ha_states.states)
+    if ha_states.reply_to:
+        ha_states.reply_to.put(result)
+
+
+def broadcast_cluster(publisher_q: Union[EQPublisher, BQPublisher],
+                      ha_states: BroadcastHAStates):
+    states_json = [{'fid': f'{state.fid}', 'status': repr(state.status)}
+                   for state in ha_states.states]
+    payload = dump_json(states_json)
+    LOG.debug('Broadcast HA states JSON: %s', payload)
+    offset = publisher_q.publish('ha-notify', payload)
+    LOG.debug('Written to epoch: %s', offset)
 
 
 class ConsumerThread(StoppableThread):
@@ -50,6 +84,7 @@ class ConsumerThread(StoppableThread):
         self.is_stopped = False
         self.consul = ConsulUtil()
         self.eq_publisher = EQPublisher()
+        self.bq_publisher = BQPublisher()
 
     def stop(self) -> None:
         self.is_stopped = True
@@ -92,6 +127,8 @@ class ConsumerThread(StoppableThread):
                         # intermittent error gets resolved.
                         decorated = (repeat_if_fails(wait_seconds=5))(fn)
                         decorated(item.evt)
+                        event: ConfHaProcess = item.evt
+                        broadcast_process_event(q, event)
                     elif isinstance(item, HaNvecGetEvent):
                         fn = motr.ha_nvec_get_reply
                         # If a consul-related exception appears, it will
@@ -103,10 +140,10 @@ class ConsumerThread(StoppableThread):
                         decorated(item)
                     elif isinstance(item, BroadcastHAStates):
                         LOG.info('HA states: %s', item.states)
-                        result: List[MessageId] = motr.broadcast_ha_states(
-                            item.states)
-                        if item.reply_to:
-                            item.reply_to.put(result)
+                        if item.is_broadcast_local:
+                            broadcast_local(item, motr)
+                        else:
+                            broadcast_cluster(self.eq_publisher, item)
                     elif isinstance(item, StobIoqError):
                         LOG.info('Stob IOQ: %s', item.fid)
                         payload = dump_json(item)
--- a/hax/hax/message.py
+++ b/hax/hax/message.py
@@ -47,6 +47,9 @@ class ProcessEvent(BaseMessage):
 @dataclass
 class BroadcastHAStates(BaseMessage):
     states: List[HAState]
+    # Differentiates between local and cluster broadcast.
+    # See hax/hax/handler.py
+    is_broadcast_local: bool
     reply_to: Optional[Queue]
 
 
--- a/hax/hax/queue/__init__.py
+++ b/hax/hax/queue/__init__.py
@@ -51,11 +51,14 @@ class BQProcessor:
             'M0_HA_MSG_NVEC': self.handle_device_state_set,
             'STOB_IOQ_ERROR': self.handle_ioq_stob_error,
         }
-        if msg_type not in handlers:
+        if msg_type in handlers:
+            handlers[msg_type](payload)
+        elif msg_type == 'HA_NOTIFY':
+            self.handle_ha_notify(payload)
+        else:
             LOG.warn('Unsupported message type given: %s. Message skipped.',
                      msg_type)
             return
-        handlers[msg_type](payload)
 
     def handle_device_state_set(self, payload: Dict[str, Any]) -> None:
         # To add check for multiple object entries in a payload.
@@ -68,7 +71,9 @@ class BQProcessor:
         q: Queue = Queue(1)
         LOG.debug('HA broadcast, node: %s device: %s state: %s',
                   payload['node'], payload['device'], payload['state'])
-        self.queue.put(BroadcastHAStates(states=[hastate], reply_to=q))
+        self.queue.put(BroadcastHAStates(states=[hastate],
+                                         is_broadcast_local=True,
+                                         reply_to=q))
         ids: List[MessageId] = q.get()
         self.herald.wait_for_any(HaLinkMessagePromise(ids))
 
@@ -82,7 +87,8 @@ class BQProcessor:
         self.queue.put(
             BroadcastHAStates(
                 states=[HAState(fid,
-                                status=ServiceHealth.FAILED)], reply_to=q))
+                                status=ServiceHealth.FAILED)],
+                is_broadcast_local=True, reply_to=q))
         ids: List[MessageId] = q.get()
         self.herald.wait_for_any(HaLinkMessagePromise(ids))
 
@@ -96,3 +102,21 @@ class BQProcessor:
             LOG.error('Invalid json payload, no key (%s) present', error)
             return None
         return HAState(sdev_fid, status=state)
+
+    def handle_ha_notify(self, payload: List[Dict[str, Any]]) -> None:
+        hastates: List[HAState] = []
+        for state in payload:
+            hastatus = ServiceHealth.UNKNOWN
+            if state['status'] == 'FAILED':
+                hastatus = ServiceHealth.FAILED
+            elif state['status'] == 'OK':
+                hastatus = ServiceHealth.OK
+            hastates.append(HAState(Fid.parse(state['fid']), hastatus))
+
+        if hastates:
+            q: Queue = Queue(1)
+            self.queue.put(BroadcastHAStates(states=hastates,
+                                             is_broadcast_local=True,
+                                             reply_to=q))
+            ids: List[MessageId] = q.get()
+            self.herald.wait_for_any(HaLinkMessagePromise(ids))
--- a/hax/hax/types.py
+++ b/hax/hax/types.py
@@ -159,6 +159,44 @@ class SnsCmStatus(Enum):
     CM_STATUS_PAUSED = 4
 
 
+class m0HaProcessEvent(Enum):
+    M0_CONF_HA_PROCESS_STARTING = 0
+    M0_CONF_HA_PROCESS_STARTED = 1
+    M0_CONF_HA_PROCESS_STOPPING = 2
+    M0_CONF_HA_PROCESS_STOPPED = 3
+
+    def str_to_Enum(self):
+        events = {'M0_CONF_HA_PROCESS_STARTING':
+                  self.M0_CONF_HA_PROCESS_STARTING,
+                  'M0_CONF_HA_PROCESS_STARTED':
+                  self.M0_CONF_HA_PROCESS_STARTED,
+                  'M0_CONF_HA_PROCESS_STOPPING':
+                  self.M0_CONF_HA_PROCESS_STOPPING,
+                  'M0_CONF_HA_PROCESS_STOPPED':
+                  self.M0_CONF_HA_PROCESS_STOPPED}
+        return events[self.name]
+
+    def __repr__(self):
+        return self.name
+
+
+class m0HaProcessType(Enum):
+    M0_CONF_HA_PROCESS_OTHER = 0
+    M0_CONF_HA_PROCESS_KERNEL = 1
+    M0_CONF_HA_PROCESS_M0MKFS = 2
+    M0_CONF_HA_PROCESS_M0D = 3
+
+    def str_to_Enum(self):
+        types = {'M0_CONF_HA_PROCESS_OTHER': self.M0_CONF_HA_PROCESS_OTHER,
+                 'M0_CONF_HA_PROCESS_KERNEL': self.M0_CONF_HA_PROCESS_KERNEL,
+                 'M0_CONF_HA_PROCESS_M0MKFS': self.M0_CONF_HA_PROCESS_M0MKFS,
+                 'M0_CONF_HA_PROCESS_M0D': self.M0_CONF_HA_PROCESS_M0D}
+        return types[self.name]
+
+    def __repr__(self):
+        return self.name
+
+
 # struct m0_spiel_repreb_status
 ReprebStatus = NamedTuple('ReprebStatus', [('fid', Fid),
                                            ('state', SnsCmStatus),
--- a/hax/hax/util.py
+++ b/hax/hax/util.py
@@ -85,6 +85,11 @@ ha_process_events = ('M0_CONF_HA_PROCESS_STARTING',
                      'M0_CONF_HA_PROCESS_STOPPING',
                      'M0_CONF_HA_PROCESS_STOPPED')
 
+ha_process_types = ('M0_CONF_HA_PROCESS_OTHER',
+                    'M0_CONF_HA_PROCESS_KERNEL',
+                    'M0_CONF_HA_PROCESS_M0MKFS',
+                    'M0_CONF_HA_PROCESS_M0D')
+
 ha_conf_obj_states = ('M0_NC_UNKNOWN',
                       'M0_NC_ONLINE',
                       'M0_NC_FAILED',
@@ -317,29 +322,32 @@ class ConsulUtil:
             raise HAConsistencyException('Failed to communicate to'
                                          ' Consul Agent') from e
 
+    def get_m0d_status(self, srv: ServiceData) -> Tuple[str, str]:
+        try:
+            key = f'processes/{srv.fid}'
+            raw_data = self.kv.kv_get(key)
+            LOG.debug('Raw value from KV: %s', raw_data)
+            data = raw_data['Value']
+            data_json = json.loads(data)
+            value: str = data_json['state']
+            stype: str = data_json['type']
+            LOG.info('key (%s) state (%s) type (%s)', key, value, stype)
+            return value, stype
+        except Exception:
+            return 'Unknown', 'Unknown'
+
     def get_m0d_statuses(self) -> List[Tuple[ServiceData, str]]:
         """
         Return the list of all Motr service statuses according to Consul
         watchers. The following services are considered: ios, confd.
         """
-        def get_status(srv: ServiceData) -> str:
-            try:
-                key = f'processes/{srv.fid}'
-                raw_data = self.kv.kv_get(key)
-                LOG.debug('Raw value from KV: %s', raw_data)
-                data = raw_data['Value']
-                value: str = json.loads(data)['state']
-                return value
-            except Exception:
-                return 'Unknown'
-
         m0d_services = set(['ios', 'confd'])
         result = []
         for service_name in self.catalog_service_names():
             if service_name not in m0d_services:
                 continue
             data = self.get_service_data_by_name(service_name)
-            result += [(item, get_status(item)) for item in data]
+            result += [(item, self.get_m0d_status(item)[0]) for item in data]
         return result
 
     def get_service_data_by_name(self, name: str) -> List[ServiceData]:
@@ -441,7 +449,11 @@ class ConsulUtil:
         assert 0 <= event.chp_event < len(ha_process_events), \
             f'Invalid event type: {event.chp_event}'
 
-        data = json.dumps({'state': ha_process_events[event.chp_event]})
+        LOG.info('Updating process status type=%s event=%s',
+                 ha_process_types[event.chp_type],
+                 ha_process_events[event.chp_event])
+        data = dump_json({'type': ha_process_types[event.chp_type],
+                          'state': ha_process_events[event.chp_event]})
         key = f'processes/{event.fid}'
         LOG.debug('Setting process status in KV: %s:%s', key, data)
         self.kv.kv_put(key, data)
@@ -527,11 +539,43 @@ class ConsulUtil:
             raise RuntimeError(
                 f'Node {local_nodename} has no service {service_name}')
         service = local_services[0]
+        svc_data = mkServiceData(self._service_by_name(local_nodename,
+                                                       service_name))
+        status, stype = self.get_m0d_status(svc_data)
+        # Motr processes are of 2 types, mkfs and m0d.
+        # The type of motr process also needs to be considered to determine
+        # the correct health of the process.
+        # Motr mkfs and m0d processes for confd and ioservices share the same
+        # motr fids. Motr mkfs processes are short lived, thus it is possible
+        # that when the status check is done the status is reported as
+        # 'warning' which is acceptable for motr mkfs process. This should not
+        # be confused with motr m0d status, thus type needs to be checked as
+        # well.
+        if stype != 'M0_CONF_HA_PROCESS_M0D':
+            ok = all(x.get('Status') == 'passing' or
+                     (x.get('Status') == 'warning' and
+                      status in ['M0_CONF_HA_PROCESS_STARTING',
+                                 'M0_CONF_HA_PROCESS_STOPPING',
+                                 'M0_CONF_HA_PROCESS_STOPPED',
+                                 'M0_CONF_HA_PROCESS_STARTED'])
+                     for x in service['Checks'])
+        else:
+            ok = all(x.get('Status') == 'passing' or
+                     (x.get('Status') == 'warning' and
+                     status in ['M0_CONF_HA_PROCESS_STARTING',
+                                'M0_CONF_HA_PROCESS_STARTED'])
+                     for x in service['Checks'])
 
-        ok = all(x.get('Status') == 'passing' for x in service['Checks'])
-        status = ServiceHealth.OK if ok else ServiceHealth.FAILED
+        LOG.debug('Service=%s status=%s type=%s',
+                  service_name, status, stype)
+        if ok:
+            svc_status = ServiceHealth.OK
+        elif status == 'Unknown':
+            svc_status = ServiceHealth.UNKNOWN
+        else:
+            svc_status = ServiceHealth.FAILED
         fid = create_process_fid(int(service['Service']['ID']))
-        return HAState(fid=fid, status=status)
+        return HAState(fid=fid, status=svc_status)
 
 
 def dump_json(obj) -> str:
--- a/hax/hax/motr/__init__.py
+++ b/hax/hax/motr/__init__.py
@@ -22,7 +22,7 @@ from errno import EAGAIN
 from typing import Any, List
 
 from hax.exception import ConfdQuorumException, RepairRebalanceException
-from hax.message import (BroadcastHAStates, EntrypointRequest, HaNvecGetEvent,
+from hax.message import (EntrypointRequest, HaNvecGetEvent,
                          ProcessEvent)
 from hax.motr.delivery import DeliveryHerald
 from hax.motr.ffi import HaxFFI, make_array, make_c_str
@@ -188,18 +188,13 @@ class Motr:
         return message_ids
 
     def _process_event_cb(self, fid, chp_event, chp_type, chp_pid):
-        LOG.info('fid=%s, chp_event=%s', fid, chp_event)
+        LOG.info('fid=%s, chp_event=%s chp_type=%s', fid, chp_event, chp_type)
         self.queue.put(
             ProcessEvent(
                 ConfHaProcess(chp_event=chp_event,
                               chp_type=chp_type,
                               chp_pid=chp_pid,
                               fid=fid)))
-        if chp_event == 3:
-            self.queue.put(
-                BroadcastHAStates(
-                    states=[HAState(fid=fid, status=ServiceHealth.FAILED)],
-                    reply_to=None))
 
     def _stob_ioq_event_cb(self, fid, sie_conf_sdev, sie_stob_id, sie_fd,
                            sie_opcode, sie_rc, sie_offset, sie_size,
--- a/hax/hax/servicemon.py
+++ b/hax/hax/servicemon.py
@@ -75,7 +75,9 @@ class ServiceMonitor(StoppableThread):
         if not state_list:
             return
         LOG.debug('Changes in statuses: %s', state_list)
-        self.q.put(BroadcastHAStates(states=state_list, reply_to=None))
+        self.q.put(BroadcastHAStates(states=state_list,
+                                     is_broadcast_local=False,
+                                     reply_to=None))
 
     def _execute(self):
         service_names: List[str] = self._get_services()
@@ -91,7 +93,7 @@ class ServiceMonitor(StoppableThread):
 
                     for name in service_names:
                         health: HAState = self.consul.get_local_service_health(
-                            name)
+                                              name)
                         if (health.status != known_statuses[name]):
                             delta.append(health)
                             known_statuses[name] = health.status
--- a/rules/ha-notify
+++ b/rules/ha-notify
@@ -1,0 +1,20 @@
+#!/bin/bash
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+h0q bq HA_NOTIFY "$HARE_RC_EVENT_PAYLOAD"


hax/hax/handler.py
```diff
@@ -92,6 +95,21 @@ def pull_msg():
                         # intermittent error gets resolved.
                         decorated = (repeat_if_fails(wait_seconds=5))(fn)
                         decorated(item.evt)
+                        event: ConfHaProcess = item.evt
+                        ha_event = pevent(event.chp_event)
+                        proc_type = ptype(event.chp_type)
+                        if proc_type == ptype.M0_CONF_HA_PROCESS_M0D:
+                            svc_status = ServiceHealth.UNKNOWN
```
knekrasov, line 102: 1. Perhaps it makes sense to extract the whole block under this `if` into a separate function.
2. The if/elif can be simplified like this:
```python
status_map = {pevent.M0_CONF_HA_PROCESS_STOPPED: ServiceHealth.FAILED, pevent.M0_CONF_HA_PROCESS_STARTED: ServiceHealth.OK}
svc_status = status_map.get(ha_event, ServiceHealth.UNKNOWN)
```

hax/hax/handler.py
```diff
@@ -103,10 +121,22 @@ def pull_msg():
...
+                            result: List[MessageId] = motr.broadcast_ha_states(
+                                item.states)
+                            if item.reply_to:
+                                item.reply_to.put(result)
+                        else:
+                            states_json = []
+                            for state in item.states:
+                                states_json.append(
+                                    {'fid': f'{state.fid}',
+                                     'status': repr(state.status)})
```
knekrasov, line 134: 5 lines above can be written shorter:
```python
states_json = [{'fid': f'{state.fid}', 'status': repr(state.status)} for state in item.states]
```

hax/hax/handler.py
```diff
@@ -103,10 +121,22 @@ def pull_msg():
...
-                            item.states)
-                        if item.reply_to:
-                            item.reply_to.put(result)
+                        if item.is_broadcast_local:
+                            result: List[MessageId] = motr.broadcast_ha_states(
+                                item.states)
+                            if item.reply_to:
+                                item.reply_to.put(result)
+                        else:
+                            states_json = []
```
knekrasov, line 130: Please consider refactoring if and else branches into separate functions.

hax/hax/message.py
```diff
@@ -47,6 +47,7 @@ class ProcessEvent(BaseMessage):
 @dataclass
 class BroadcastHAStates(BaseMessage):
     states: List[HAState]
+    is_broadcast_local: bool
```
knekrasov, line 50: Can you add a one-line comment telling what this field is about?
BTW doesn't it make sense just create two different message types (instead of a single one) if the semantics differs much?

hax/hax/queue/__init__.py
```diff
@@ -96,3 +100,22 @@ def to_ha_state(self, objinfo: Dict[str, str]) -> Optional[HAState]:
             LOG.error('Invalid json payload, no key (%s) present', error)
             return None
         return HAState(sdev_fid, status=state)
+
+    def handle_ha_notify(self, payload: Dict[str, Any]) -> None:
+        hastates: List[HAState] = []
+        for state in payload:
+            data = json.loads(json.dumps(state))
```
knekrasov, line 107: Why do you need to re-encode it? Is it a workaround for some problem?
In other words, why can't we just say
```python
data = state
```

hax/hax/types.py
```diff
@@ -159,6 +159,44 @@ class SnsCmStatus(Enum):
     CM_STATUS_PAUSED = 4
 
 
+class m0HaProcessEvent(Enum):
+    M0_CONF_HA_PROCESS_STARTING = 0
+    M0_CONF_HA_PROCESS_STARTED = 1
+    M0_CONF_HA_PROCESS_STOPPING = 2
+    M0_CONF_HA_PROCESS_STOPPED = 3
+
+    def strToEnum(self):
```
knekrasov, line 168: [Optional]
In python underscore style is more preferred: strToEnum -> str_to_enum

hax/hax/util.py
```diff
@@ -317,29 +322,31 @@ def get_session_node(self, session_id: str) -> str:
             raise HAConsistencyException('Failed to communicate to'
                                          ' Consul Agent') from e
 
+    def get_m0d_status(self, srv: ServiceData) -> Tuple[str, str]:
+        try:
+            key = f'processes/{srv.fid}'
+            raw_data = self.kv.kv_get(key)
+            LOG.debug('Raw value from KV: %s', raw_data)
+            data = raw_data['Value']
+            value: str = json.loads(data)['state']
```
knekrasov, line 331: `json.loads(data)` is evaluated twice. No need to repeat it.

hax/hax/util.py
```diff
@@ -441,7 +448,11 @@ def update_process_status(self, event: ConfHaProcess) -> None:
         assert 0 <= event.chp_event < len(ha_process_events), \
             f'Invalid event type: {event.chp_event}'
 
-        data = json.dumps({'state': ha_process_events[event.chp_event]})
+        LOG.info('Updating process status type=%s event=%s',
+                 ha_process_types[event.chp_type],
+                 ha_process_events[event.chp_event])
+        data = json.dumps({'type': ha_process_types[event.chp_type],
```
knekrasov, line 454: In many cases hax code uses simplejson.dumps/loads in favor of json.dumps/loads
As far as I remember this was caused by the way how poorly json module encodes/decodes Enums and our Fid. Please consider using `[dump_json](https://github.com/Seagate/cortx-hare/blob/main/hax/hax/util.py#L537)()` function from our util.py.

As a rule of a thumb, one probably should prefer simplejson instead of json module (until otherwise is proven).

knekrasov (changes requested):

hax/hax/message.py
mssawant: So basically we want to have 2 types of broadcasts, local and to entire cluster. So Idea is that when a user module needs to broadcast an event `is_local_broadcast` is set to false so the event is notified to all the nodes in the cluster. And when the message is processed as part of EQ (`is_broadcast_local=False`) the respective nodes broadcast the event locally using `is_broadcast_local=True`.
I thought of having another message type as well but amount of changes did not justify the need. We anyway need to broadcast, just need to differentiate between local vs global. But if this makes the code unreadable then we can have a separate message type.

hax/hax/queue/__init__.py
mssawant: The payload in this case is a list of dictionary strings, i.e. [{fid:0x000.., status: FAILED}], while trying to iterate over the list and indexing in a dictionary, was getting an error as `Invalid index type "str" for "str" for Union[int, slice]` . After loading the individual dictionary string I was able to index into the dictionary.

knekrasov: Then probably it is safe to change the type hint of payload?

BTW mypy is very dumb, so there is always a possibility to force the type like this:
myvalue: str = mydict['key']
or myvalue: str = str(mydict['key'])

anyway this is more performant than code/reencode cycle just to fool the type checker 😄

mssawant: Let me try..

hax/hax/message.py
knekrasov: Understood. Let's keep this boolean flag `is_broadcast_local` for now; we can revisit the need of separate class later.

Just please add an explanatory comment to the code to give an idea what this flag is about.
Thanks!

mssawant: Sure.

hax/hax/handler.py
mssawant: Agree. Thanks.

hax/hax/handler.py
mssawant: Okay.

hax/hax/handler.py
mssawant: Okay.

hax/hax/queue/__init__.py
mssawant: Working, had to take out HA_NOTIFY from the function vector though due to difference in the payload hint from rest of the functions.

hax/hax/types.py
mssawant: Accepted.

hax/hax/util.py
mssawant: Right, fixed.

hax/hax/util.py
mssawant: Okay, fixed.

hax/hax/queue/__init__.py
```diff
@@ -82,7 +87,8 @@ def handle_ioq_stob_error(self, payload: Dict[str, Any]) -> None:
         self.queue.put(
             BroadcastHAStates(
                 states=[HAState(fid,
-                                status=ServiceHealth.FAILED)], reply_to=q))
+                                status=ServiceHealth.FAILED)],
+                is_broadcast_local=True, reply_to=q))
         ids: List[MessageId] = q.get()
         self.herald.wait_for_any(HaLinkMessagePromise(ids))
```
knekrasov, line 93: This is probably one reason why it makes sense to have two separate classes for BroadcastHAStates (one local, another one - cluster-wise).
This "wait_for_any" semantics makes sense if and only if it is local.

But anyway, let's address this later.

knekrasov (approved):

andriytk:
> Hax detects motr process failures through service monitor but broadcasts
the events only to local node's motr processes. Due to this motr processes
from peer nodes do not have knowledge about the failure and continue to
use stale connections.

Hax broadcasts ha-status events to local m0-processes only by design. Each node has its own `hax` agent. When some ha-status change happens - the cluster-wide update happens via Consul's catalogue: `hax` agent on each node will get this update from Consul (via watch mechanism or polling - doesn't matter) and update all the local m0-processes.

Now, what's the problem we are trying to fix here?

mssawant:
@andriytk, No, presently hax service monitor monitor's only local node's services using `Consul's catalogue` by filtering the local node-name thus only if any of the local node's m0d service fails service monitor detects the same and broadcasts. Broadcast logic invokes `motr.broadcast_ha_states(ha_states.states)` which invokes C land ha_notify that sends notifications over the available halinks which are again only for the local m0d processes which establish connection with the local hax.

So overall the broadcasts happen only on the local node.

mssawant:
@andriytk,
>Now, what's the problem we are trying to fix here?

As mentioned in the commit message, if a m0d service restarts in a running cluster then subsequent IOs fail targeted to the restarted m0d. This happens is a peer node's s3servers (or any clovis aaps running remotely) send ios to the restarted ios and because they still have the stale rpc connection as they did not receive any notification about the respective ios restart.
Now we are notifying all the motr services in the cluster including the clients about the ios restart (stop and start) which helps them to re-connect the restarted ioservice and IO succeeds.

hax/hax/queue/__init__.py
mssawant: Okay, so if its a cluster wide broadcast then the ids list will be empty right? do you see any problem here presently?

knekrasov: At this moment - no, I don't see issues, that's why I approved it. Just the code can look a bit misleading from the first glance which is not good in longer term.

mssawant: Makes sense. Thanks.

mssawant:
@andriytk , this is working in LDR1 because s3servers are also restarted if any of the ioservices restart s3servers from both the nodes are restarted due to pacemaker constraints of s3auth-clone dependency of both the ioservices and s3servers dependency on s3auth-clone.

andriytk:
@mssawant , sorry - I don't understand. If any service (hax, confd, ios, s3server) fails - its status will be changed in Consul's catalogue, right? If so - it will be picked up by all hax agents on all nodes and broadcasted to all m0-processes. What am I missing? Isn't it how it works now? If it isn't - then when was it changed and why?

mssawant:
@andriytk , check-service check monitors local node's services only and reports the same to local consul agent. Hax service monitor also monitors only the local node services as mentioned above.
```
                        health: HAState = self.consul.get_local_service_health(
                                              name)
                        if (health.status != known_statuses[name]):
                            delta.append(health)
                            known_statuses[name] = health.status
```
We don't have watchers that monitor all the services of the cluster. There no kv as such which is updated here that is being watched so that all the consul nodes get notified about it. Service monitor on discovering the local node service failure presently only broadcasts to the local node's motr and hax services.
You can try on experiment, bootstrap a 2 node cluster, not a pacemaker cluster but just Hare + motr cluster. Start a motr m0client,
```
m0client -l '192.168.9.108:12345:44:4' -H '192.168.9.108:12345:1:1' -p '<0x7000000000000001:39>' -P '0x7200000000000000:0'
```
(update the ip addresses).

once started, do some IO,
```
write 1048681 /tmp/srcfile 4096 200 50
```
now restart a m0d service on peer node, not where m0client is started.
```
systemctl restart m0d@0x7200000000000001:0x27
```
make sure m0d is started.
Now read the data above written data,
```
read 1048681 /tmp/srcfile1 4096 200 50
```
This will fail. 
Now stop and restart m0client and try again, read should succeed.

andriytk:
> check-service check monitors local node's services only and reports the same to local consul agent.

Yes. And the local Consul agent updates the services states in its catalogue - which it distributes among all the server agents in the cluster. So all hax agents can pick up the updated statuses from the catalogue on all the nodes. Right?

rajanikantchirmade:
> Was m0client subscribed in hax to receive ha-updates? But it should not be because hax did not see any change in Consul's catalogue about the crashed m0d.
@andriytk  
I think this  may be reason for `Hax` not checking service state since `m0client` is not registered as services (as we have S3server registered)

knekrasov:
@mssawant It just came to my head that probably you didn't need to involve BQ queue here at all.
ServiceMonitor thread polls the service states. It filters out explicitly the processes that are not running at the current node (because we needed to repeat the Consul's watch handler behaviour).

Why don't we just get rid of that filtering out? 

Every hax at every node will get the status of all processes running at all nodes. But they will be broadcast to local motr land. Will that suffice?

mssawant:
@andriytk,
>Now, as for your example with m0client - the question is why it didn't receive the NC_FAILED about the crashed m0d, right? Well, maybe it's because local hax agent didn't send it to it for some reason? Maybe due to some bug in hax? Was m0client subscribed in hax to receive ha-updates? But it should not be because hax did not see any change in Consul's catalogue about the crashed m0d. (Unless there is some issue with the local check of this m0d service which did not update Consul about its crash.) Do you have some logs from hax to analyse?

Hax did not see any change because it does not look for peer nodes services. it just looks for local node's services. Which seems fine to me. but broadcast should happen to all. You can try that small exercise actually its pretty simple.

andriytk:
> Hax did not see any change because it does not look for peer nodes services. it just looks for local node's services. Which seems fine to me.

@mssawant , why hax should look for peer nodes? `hax` should work *only* with local processes - that's the way it was designed.

Sorry, I completely lost in what the problem you are trying to fix here...

mssawant:
@knekrasov , I did think of it but getting statuses of all the nodes all the services which can be large enough and parsing it on every node every second which can too much work. I thought it would be better to just keep the monitoring as it is and the node that has a failed service just does a broadcast to everyone.
EQ/BQ seemed to be the right way where a node needs to just let everyone know about its local problem.

knekrasov:
@mssawant What @andriytk also tries to say is that when we were using service watch mechanism (i.e. before we introduced ServiceMonitor thread in hax), Consul did notify about ALL registered services at all nodes. So local hax will also know about a failure about another machine.

So probably we just can revert my PR with commit https://github.com/Seagate/cortx-hare/commit/385bc6cd (since we don't seem to need the service monitor manually - @andriytk  is that correct?)

andriytk:
Yes, looks like I got the problem now - I thought that `hax` is polling Consul catalogue for the change of *any* service in the cluster, not only the local ones.

I agree - we better roll back that polling change in `hax`.

Currently, we completely rely on Consul for the services ha-updates within the cluster. Consul already maintains the states of all the services in its catalogue anyway, so we get it for free. It is just simpler and easier than implement the same via the BQ queue in the KV Store. I don't see any need to change this.

andriytk:
> Hax detects motr process failures through service monitor but broadcasts
the events only to local node's motr processes. Due to this motr processes
from peer nodes do not have knowledge about the failure and continue to
use stale connections.

@mssawant you probably meant "but broadcasts the events only *about* local node's Motr processes", right? Because `hax` should broadcast *to* local node's processes only - that's Ok and that's how it should work.

knekrasov:
@andriytk 
> you probably meant "but broadcasts the events only about local node's Motr processes", right? That would make sense then. Because hax should broadcast to local node's processes only - that's Ok and that's how it should work.

Are you sure that that is correct? I mean that it should work like this.

As @max-seagate said, currently ha_note events must be cluster-wise, not node-wise.

knekrasov:
@andriytk just consider an example.

`Node A` has m0d1 running, `Node B` has m0d2 running.
They communicate to each other via ha_link.
Then m0d2 dies. Hax at Node B reacts (because it reacts on local services only, as you've just described).

It will mean that m0d1 (that is running at another node) will not receive ha_note with `M0_NC_FAILED`. And thus ha_link will not be disconnected. m0d1 will not know that another participant of their communication is dead. Just because it is at another node.

andriytk:
m0d2 dies -> Consul will notice it via check mechanism (it runs checks on all the services once a second calling `systemctl status` command) -> update its catalogue -> this update will be noticed by both `hax` agents, on `node A` and `node B` -> each of them will notify all its local m0-processes.

knekrasov:
@mssawant  please ignore our with @andriytk last few comments, they are misleading.

He meant that your PR description is correct and describes how the system is supposed to work :-) (sorry if I added extra noise, I truly misunderstood the message).

Hopefully everybody is at the same page now.

mssawant:
@knekrasov , @andriytk , I reverted servicemon related changes,
- Cluster bootstrapped fine
- Initial IO was successful
- Restarted ioservice on node-2
- post that I just see the FAILED notification for node 2 services only and nothing on node-1.
   ```
RN  [ha/halon/interface.c:1136:m0_halon_interface_delivered]  hl=0x7f15f405e470 ep=192.168.9.107@tcp:12345:2:2 epoch=0 tag=90 type=7
 [DEBUG] {qconsumer} Got ProcessEvent(evt=<hax.types.ConfHaProcess object at 0x7f161405c240>) message from queue
 [DEBUG] {qconsumer} Attempting to invoke the repeatable call: update_process_status
 [DEBUG] {qconsumer} Setting process status in KV: processes/0x7200000000000001:0x27:{"state": "M0_CONF_HA_PROCESS_STOPPED"}
 [DEBUG] {qconsumer} The repeatable call succeeded: update_process_status
 [DEBUG] {qconsumer} Waiting for the next message
 [DEBUG] {qconsumer} Got BroadcastHAStates(states=[HAState(fid=0x7200000000000001:0x27, status=FAILED)], reply_to=None) message from queue
 [INFO] {qconsumer} HA states: [HAState(fid=0x7200000000000001:0x27, status=FAILED)]
 [DEBUG] {qconsumer} Broadcasting HA states [HAState(fid=0x7200000000000001:0x27, status=FAILED)] over ha_link
 [DEBUG] {qconsumer} Process fid=0x7200000000000001:0x27 encloses 7 services as follows: [FidWithType(fid=0x7300000000000001:0x30, service_type='addb2'), FidWithType(fid=0x7RN  [ha/halon/interface.c:1112:m0_halon_interface_send]  hl=0x7f15f4002ac0 ep=192.168.9.107@tcp:12345:1:1 epoch=0 tag=23 type=2
RN  [ha/halon/interface.c:1112:m0_halon_interface_send]  hl=0x7f15f400d710 ep=192.168.9.107@tcp:12345:2:1 epoch=0 tag=5 type=2
RN  [ha/halon/interface.c:1112:m0_halon_interface_send]  hl=0x7f15f4029be0 ep=192.168.9.107@tcp:12345:2:1 epoch=0 tag=11 type=2
RN  [ha/halon/interface.c:1112:m0_halon_interface_send]  hl=0x7f15f4037d10 ep=192.168.9.107@tcp:12345:2:2 epoch=0 tag=11 type=2
RN  [ha/halon/interface.c:1112:m0_halon_interface_send]  hl=0x7f15f405e470 ep=192.168.9.107@tcp:12345:2:2 epoch=0 tag=11 type=2
 [DEBUG] {qconsumer} Broadcast HA state complete with the following message_ids = [MessageId(0x7f15f4002ac0, 23), MessageId(0x7f15f400d710, 5), MessageId(0x7f15f4029be0, 11) [DEBUG] {qconsumer} Waiting for the next message
``` 
IO fails post restart although all the services are running,
```
[root@ssc-vm-c-0553 cortx-hare-1]# hctl status
Profile: 0x7000000000000001:0x39
Data pools:
    0x6f00000000000001:0x3a
Services:
    ssc-vm-c-0553.colo.seagate.com
    [started]  hax        0x7200000000000001:0x21  192.168.9.107@tcp:12345:1:1
    [started]  confd      0x7200000000000001:0x24  192.168.9.107@tcp:12345:2:1
    [started]  ioservice  0x7200000000000001:0x27  192.168.9.107@tcp:12345:2:2
    [unknown]  m0_client  0x7200000000000001:0x33  192.168.9.107@tcp:12345:4:1
    [unknown]  m0_client  0x7200000000000001:0x36  192.168.9.107@tcp:12345:4:2
    ssc-vm-c-0552.colo.seagate.com  (RC)
    [started]  hax        0x7200000000000001:0x6   192.168.9.108@tcp:12345:1:1
    [started]  confd      0x7200000000000001:0x9   192.168.9.108@tcp:12345:2:1
    [started]  ioservice  0x7200000000000001:0xc   192.168.9.108@tcp:12345:2:2
    [unknown]  m0_client  0x7200000000000001:0x18  192.168.9.108@tcp:12345:4:1
    [unknown]  m0_client  0x7200000000000001:0x1b  192.168.9.108@tcp:12345:4:2
```

```
m0client >>write 1048680 /tmp/srcfile 4096 200 50
m0client >>read 1048680 /tmp/srcfile1 4096 200 50
motr[12412]:  ca80  ERROR  [io_req_fop.c:159:io_bottom_half]  [0x1d828a0] rpc item 0x7f68bc0aa010 rc=-110
motr[12412]:  ca80  ERROR  [io_req_fop.c:208:io_bottom_half]  ioo=0x1d828a0 from=192.168.9.107@tcp:12345:2:2 rc=-110 ti_rc=-110 @<4300000300000000:100068>
motr[12412]:  ca80  ERROR  [io_req_fop.c:159:io_bottom_half]  [0x1d828a0] rpc item 0x7f68bc0ab460 rc=-110
motr[12412]:  ca80  ERROR  [io_req_fop.c:159:io_bottom_half]  [0x1d828a0] rpc item 0x7f68bc0ab460 rc=-110
motr[12412]:  ca80  ERROR  [io_req_fop.c:208:io_bottom_half]  ioo=0x1d828a0 from=192.168.9.107@tcp:12345:2:2 rc=-110 ti_rc=-110 @<4300000200000000:100068>
motr[12412]:  ca80  ERROR  [io_req_fop.c:159:io_bottom_half]  [0x1d828a0] rpc item 0x7f68bc0aa010 rc=-110
motr[12412]:  9a80  ERROR  [io_req_fop.c:159:io_bottom_half]  [0x1d828a0] rpc item 0x7f68b80487d0 rc=-110
motr[12412]:  9a80  ERROR  [io_req_fop.c:208:io_bottom_half]  ioo=0x1d828a0 from=192.168.9.107@tcp:12345:2:2 rc=-110 ti_rc=-110 @<4300000200000000:100068>
```
This means that presently service restart notification is not received by both the nodes.

**Comment by mssawant** [comment]
Experiment 2: Killing peer node's io service

After restarting m0client post earlier experiment, IO succeeded
```
m0client >>m0client >>quit
Help:
touch   OBJ_ID
write   OBJ_ID SRC_FILE BLOCK_SIZE BLOCK_COUNT BLOCKS_PER_IO UPDATE_FLAG OFFSET
read    OBJ_ID DEST_FILE BLOCK_SIZE BLOCK_COUNT BLOCKS_PER_IO
delete  OBJ_ID
help
quit

q[root@ssc-vm-c-0552 cortx-hare-1]#
[root@ssc-vm-c-0552 cortx-hare-1]# m0client -l '192.168.9.108:12345:44:4' -H '192.168.9.108:12345:1:1' -p '<0x7000000000000001:39>' -P '0x7200000000000000:0'
m0client >>read 1048680 /tmp/srcfile1 4096 200 50
```


```
[root@ssc-vm-c-0553 cortx-hare-1]# hctl status
Profile: 0x7000000000000001:0x39
Data pools:
    0x6f00000000000001:0x3a
Services:
    ssc-vm-c-0553.colo.seagate.com
    [started]  hax        0x7200000000000001:0x21  192.168.9.107@tcp:12345:1:1
    [started]  confd      0x7200000000000001:0x24  192.168.9.107@tcp:12345:2:1
    [started]  ioservice  0x7200000000000001:0x27  192.168.9.107@tcp:12345:2:2
    [unknown]  m0_client  0x7200000000000001:0x33  192.168.9.107@tcp:12345:4:1
    [unknown]  m0_client  0x7200000000000001:0x36  192.168.9.107@tcp:12345:4:2
    ssc-vm-c-0552.colo.seagate.com  (RC)
    [started]  hax        0x7200000000000001:0x6   192.168.9.108@tcp:12345:1:1
    [started]  confd      0x7200000000000001:0x9   192.168.9.108@tcp:12345:2:1
    [started]  ioservice  0x7200000000000001:0xc   192.168.9.108@tcp:12345:2:2
    [unknown]  m0_client  0x7200000000000001:0x18  192.168.9.108@tcp:12345:4:1
    [unknown]  m0_client  0x7200000000000001:0x1b  192.168.9.108@tcp:12345:4:2

[root@ssc-vm-c-0553 cortx-hare-1]# ps -aux | grep m0d
root      8976  0.3  3.4 10080292 273768 ?     SLsl 11:17   0:05 /home/520478/cortx-motr-new/cortx-motr/motr/.libs/lt-m0d -e lnet:192.168.9.107@tcp:12345:2:1 -A linuxstob:addb-stobs -f <0x7200000000000001:0x24> -T linux -S stobs -D db -m 65536 -q 16 -w 8 -c /etc/motr/confd.xc -H 192.168.9.107@tcp:12345:1:1 -U
root     10719  0.1  3.8 10179188 304884 ?     SLsl 11:19   0:01 /home/520478/cortx-motr-new/cortx-motr/motr/.libs/lt-m0d -e lnet:192.168.9.107@tcp:12345:2:2 -A linuxstob:addb-stobs -f <0x7200000000000001:0x27> -T ad -S stobs -D db -m 65536 -q 16 -w 8 -H 192.168.9.107@tcp:12345:1:1 -U
root     28965  0.0  0.0 112812   964 pts/0    S+   11:43   0:00 grep --color=auto m0d
[root@ssc-vm-c-0553 cortx-hare-1]# kill -9 10719
[root@ssc-vm-c-0553 cortx-hare-1]# ps -aux | grep m0d
root      8976  0.3  3.4 10080292 273768 ?     SLsl 11:17   0:05 /home/520478/cortx-motr-new/cortx-motr/motr/.libs/lt-m0d -e lnet:192.168.9.107@tcp:12345:2:1 -A linuxstob:addb-stobs -f <0x7200000000000001:0x24> -T linux -S stobs -D db -m 65536 -q 16 -w 8 -c /etc/motr/confd.xc -H 192.168.9.107@tcp:12345:1:1 -U
root     29242  0.0  0.0 112812   964 pts/0    S+   11:44   0:00 grep --color=auto m0d
[root@ssc-vm-c-0553 cortx-hare-1]# hctl status
Profile: 0x7000000000000001:0x39
Data pools:
    0x6f00000000000001:0x3a
Services:
    ssc-vm-c-0553.colo.seagate.com
    [started]  hax        0x7200000000000001:0x21  192.168.9.107@tcp:12345:1:1
    [started]  confd      0x7200000000000001:0x24  192.168.9.107@tcp:12345:2:1
    [offline]  ioservice  0x7200000000000001:0x27  192.168.9.107@tcp:12345:2:2
    [unknown]  m0_client  0x7200000000000001:0x33  192.168.9.107@tcp:12345:4:1
    [unknown]  m0_client  0x7200000000000001:0x36  192.168.9.107@tcp:12345:4:2
    ssc-vm-c-0552.colo.seagate.com  (RC)
    [started]  hax        0x7200000000000001:0x6   192.168.9.108@tcp:12345:1:1
    [started]  confd      0x7200000000000001:0x9   192.168.9.108@tcp:12345:2:1
    [started]  ioservice  0x7200000000000001:0xc   192.168.9.108@tcp:12345:2:2
    [unknown]  m0_client  0x7200000000000001:0x18  192.168.9.108@tcp:12345:4:1
    [unknown]  m0_client  0x7200000000000001:0x1b  192.168.9.108@tcp:12345:4:2

[root@ssc-vm-c-0553 cortx-hare-1]# systemctl status m0d@0x7200000000000001:0x27
● m0d@0x7200000000000001:0x27.service - m0d helper for 0x7200000000000001:0x27 service-FID
   Loaded: loaded (/usr/lib/systemd/system/m0d@.service; static; vendor preset: disabled)
   Active: failed (Result: signal) since Tue 2020-10-20 11:44:14 MDT; 1min 0s ago
  Process: 10719 ExecStart=/usr/libexec/cortx-motr/motr-server m0d-%i (code=killed, signal=KILL)
 Main PID: 10719 (code=killed, signal=KILL)

[root@ssc-vm-c-0553 cortx-hare-1]# systemctl status m0d@0x7200000000000001:0x27
● m0d@0x7200000000000001:0x27.service - m0d helper for 0x7200000000000001:0x27 service-FID
   Loaded: loaded (/usr/lib/systemd/system/m0d@.service; static; vendor preset: disabled)
   Active: active (running) since Tue 2020-10-20 11:45:50 MDT; 8s ago
 Main PID: 30368 (lt-m0d)
   CGroup: /system.slice/system-m0d.slice/m0d@0x7200000000000001:0x27.service
           └─30368 /home/520478/cortx-motr-new/cortx-motr/motr/.libs/lt-m0d -e lnet:192.168.9.107@tcp:12345:2:2 -A linuxstob:addb-stobs -f <0x7200000000000001:0x27> -T ad...

Oct 20 11:45:50 ssc-vm-c-0553.colo.seagate.com motr-server[30368]: MOTR_PROCESS_FID: 0x7200000000000001:0x27
Oct 20 11:45:50 ssc-vm-c-0553.colo.seagate.com motr-server[30368]: MOTR_HA_EP: 192.168.9.107@tcp:12345:1:1
Oct 20 11:45:50 ssc-vm-c-0553.colo.seagate.com motr-server[30368]: WARN: query not existing genders var 'm0_max_rpc_msg_size'
Oct 20 11:45:50 ssc-vm-c-0553.colo.seagate.com motr-server[30368]: WARN: query not existing genders var 'm0_min_rpc_recvq_len'
Oct 20 11:45:50 ssc-vm-c-0553.colo.seagate.com motr-server[30368]: WARN: query not existing genders var 'm0_directio'
Oct 20 11:45:50 ssc-vm-c-0553.colo.seagate.com motr-server[30368]: vm.max_map_count = 30000000
Oct 20 11:45:50 ssc-vm-c-0553.colo.seagate.com motr-server[30368]: + exec /home/520478/cortx-motr-new/cortx-motr/motr/m0d -e lnet:192.168.9.107@tcp:12345:2:2 -A l...5:1:1 -UOct 20 11:45:51 ssc-vm-c-0553.colo.seagate.com motr-server[30368]: lt-m0d: systemd notifications not allowed
Oct 20 11:45:51 ssc-vm-c-0553.colo.seagate.com motr-server[30368]: Started
Oct 20 11:45:51 ssc-vm-c-0553.colo.seagate.com motr-server[30368]: Press CTRL+C to quit.

[root@ssc-vm-c-0553 cortx-hare-1]# hctl status
Profile: 0x7000000000000001:0x39
Data pools:
    0x6f00000000000001:0x3a
Services:
    ssc-vm-c-0553.colo.seagate.com
    [started]  hax        0x7200000000000001:0x21  192.168.9.107@tcp:12345:1:1
    [started]  confd      0x7200000000000001:0x24  192.168.9.107@tcp:12345:2:1
    [started]  ioservice  0x7200000000000001:0x27  192.168.9.107@tcp:12345:2:2
    [unknown]  m0_client  0x7200000000000001:0x33  192.168.9.107@tcp:12345:4:1
    [unknown]  m0_client  0x7200000000000001:0x36  192.168.9.107@tcp:12345:4:2
    ssc-vm-c-0552.colo.seagate.com  (RC)
    [started]  hax        0x7200000000000001:0x6   192.168.9.108@tcp:12345:1:1
    [started]  confd      0x7200000000000001:0x9   192.168.9.108@tcp:12345:2:1
    [started]  ioservice  0x7200000000000001:0xc   192.168.9.108@tcp:12345:2:2
    [unknown]  m0_client  0x7200000000000001:0x18  192.168.9.108@tcp:12345:4:1
    [unknown]  m0_client  0x7200000000000001:0x1b  192.168.9.108@tcp:12345:4:2
```

IO post restart fails in this case too,
```
m0client >>read 1048680 /tmp/srcfile2 4096 200 50
motr[32603]:  ea80  ERROR  [io_req_fop.c:159:io_bottom_half]  [0x1e90950] rpc item 0x7f7f98049350 rc=-110
motr[32603]:  ea80  ERROR  [io_req_fop.c:208:io_bottom_half]  ioo=0x1e90950 from=192.168.9.107@tcp:12345:2:2 rc=-110 ti_rc=-110 @<4300000200000000:100068>
motr[32603]:  ea80  ERROR  [io_req_fop.c:159:io_bottom_half]  [0x1e90950] rpc item 0x7f7f98044f10 rc=-110
motr[32603]:  ea80  ERROR  [io_req_fop.c:159:io_bottom_half]  [0x1e90950] rpc item 0x7f7f98044f10 rc=-110
motr[32603]:  ea80  ERROR  [io_req_fop.c:208:io_bottom_half]  ioo=0x1e90950 from=192.168.9.107@tcp:12345:2:2 rc=-110 ti_rc=-110 @<4300000300000000:100068>
motr[32603]:  ea80  ERROR  [io_req_fop.c:159:io_bottom_half]  [0x1e90950] rpc item 0x7f7f98049350 rc=-110
motr[32603]:  6a80  ERROR  [io_req_fop.c:159:io_bottom_half]  [0x1e90950] rpc item 0x7f7fa40aa1c0 rc=-110
```
This means that ioservice failure notification are not received by both the nodes.

**Comment by andriytk** [comment]
@mssawant the watch for the services ha-status changes should be called on all the nodes in the cluster.

We have `consul-watch-handler` which logs all the services statuses whenever they got changed (for debug purposes) to `/var/log/hare/consul-watch-handler.log` file also - can you check it on both the nodes, please?

**Comment by mssawant** [comment]
@andriytk , On killing ioservice on both the nodes I see the notifications in consul-watch-handler.log on both the nodes but there is no corresponding broadcast event related to it on either of them somehow. Something seems to be broken,  don't see any related change in https://github.com/Seagate/cortx-hare/pull/1348.
```
node-1

2020-10-20 14:46:02 [
2020-10-20 14:46:02   {
2020-10-20 14:46:02     "Node": "ssc-vm-c-0552.colo.seagate.com",
2020-10-20 14:46:02     "ServiceID": "12",
2020-10-20 14:46:02     "ServiceName": "ios",
2020-10-20 14:46:02     "status": "online"
2020-10-20 14:46:02   },
2020-10-20 14:46:02   {
2020-10-20 14:46:02     "Node": "ssc-vm-c-0553.colo.seagate.com",
2020-10-20 14:46:02     "ServiceID": "39",
2020-10-20 14:46:02     "ServiceName": "ios",
2020-10-20 14:46:02     "status": "offline"
2020-10-20 14:46:02   }
2020-10-20 14:46:02 ]



2020-10-20 14:47:45 [
2020-10-20 14:47:45   {
2020-10-20 14:47:45     "Node": "ssc-vm-c-0552.colo.seagate.com",
2020-10-20 14:47:45     "ServiceID": "12",
2020-10-20 14:47:45     "ServiceName": "ios",
2020-10-20 14:47:45     "status": "online"
2020-10-20 14:47:45   },
2020-10-20 14:47:45   {
2020-10-20 14:47:45     "Node": "ssc-vm-c-0553.colo.seagate.com",
2020-10-20 14:47:45     "ServiceID": "39",
2020-10-20 14:47:45     "ServiceName": "ios",
2020-10-20 14:47:45     "status": "online"
2020-10-20 14:47:45   }
2020-10-20 14:47:45 ]

node-2

2020-10-20 14:46:02     "Node": "ssc-vm-c-0552.colo.seagate.com",
2020-10-20 14:46:02     "ServiceID": "12",
2020-10-20 14:46:02     "ServiceName": "ios",
2020-10-20 14:46:02     "status": "online"
2020-10-20 14:46:02   },
2020-10-20 14:46:02   {
2020-10-20 14:46:02     "Node": "ssc-vm-c-0553.colo.seagate.com",
2020-10-20 14:46:02     "ServiceID": "39",
2020-10-20 14:46:02     "ServiceName": "ios",
2020-10-20 14:46:02     "status": "offline"
2020-10-20 14:46:02   }
2020-10-20 14:46:02 ]
2020-10-20 14:47:45 [
2020-10-20 14:47:45   {
2020-10-20 14:47:45     "Node": "ssc-vm-c-0552.colo.seagate.com",
2020-10-20 14:47:45     "ServiceID": "12",
2020-10-20 14:47:45     "ServiceName": "ios",
2020-10-20 14:47:45     "status": "online"
2020-10-20 14:47:45   },
2020-10-20 14:47:45   {
2020-10-20 14:47:45     "Node": "ssc-vm-c-0553.colo.seagate.com",
2020-10-20 14:47:45     "ServiceID": "39",
2020-10-20 14:47:45     "ServiceName": "ios",
2020-10-20 14:47:45     "status": "online"
2020-10-20 14:47:45   }
2020-10-20 14:47:45 ]
```

**Comment by mssawant** [comment]
I wonder if commit `7acd9458941aa19ce97a04c62b9ef9d856f3c4df` could have broken this mechanism.
Earlier there was this do_post routine that used to broadcast, now I don't see how broadcast happens on such Consul events.
```
-class KVHandler(BaseHTTPRequestHandler):
-    def __init__(self, req, client_addr, server):
-        super().__init__(req, client_addr, server)
-        self.server = server
-
-    def _set_headers(self):
-        self.send_response(200)
-        self.send_header('Content-type', 'application/json')
-        self.end_headers()
-
-    def do_GET(self):
-        self._set_headers()
-        result = json.dumps({'message': 'I am alive'})
-        self.wfile.write(result.encode())
-
-    def do_HEAD(self):
-        self._set_headers()
-
-    def do_POST(self):
-        self._set_headers()
-        content_length = int(self.headers['Content-Length'])
-        post_data = self.rfile.read(content_length)
-        logging.debug('POST request received: %s', post_data)
-
-        ha_states = self.to_ha_states(KVHandler.parse_json(post_data))
-        logging.info('HA states: %s', ha_states)
-        self.server.halink.broadcast_ha_states(ha_states)
-        logging.debug('POST request processed')
```
Now we have this,
```
    LOG.info(f'Starting HTTP server at {web_address}:{port} ...')
    try:
        web.run_app(app, host=web_address, port=port)
        LOG.debug('Server stopped normally')
    finally:
        LOG.debug('Stopping the threads')
        for thread in threads_to_wait:
            thread.stop()
        for thread in threads_to_wait:
            thread.join()
```
@knekrasov can you please confirm regarding this? cc @andriytk.

knekrasov:
@mssawant 
>  wonder if commit 7acd9458941aa19ce97a04c62b9ef9d856f3c4df could have broken this mechanism.

Please have a look at https://github.com/Seagate/cortx-hare/blob/7ff5ae233e84faada869436bbb23169a309ba924/hax/hax/server.py (it is the commit right before servicemon.py was introduced).

We setup aiohttp server so that it listens to POST / requests and handles them as ha updates (this is where Consul's POST requests are coming to):
image

The server just puts that command to the queue for ConsumerThread (the only one that interacts with Motr land):
image

And [there](https://github.com/Seagate/cortx-hare/blob/7ff5ae233e84faada869436bbb23169a309ba924/hax/hax/handler.py#L102-L107) it does quite the same thing as it was done in KVHandler:
image


So I don't see any semantic difference to KVHandler version. As you see, the logic how Consul jsons get translated into ha_notes is the same. We definitely don't filter anything out in Python code.

What was changed since the times of KVHandler:
1. HTTP server is now aiohttp
2. ha_notes are not sent out to Motr from the server's thread.

andriytk:
@mssawant , could it be some setup issue? Did you rebuild `hax`?..

knekrasov:
@mssawant, @andriytk I've got two news.

1. I can confirm that reverting ServiceMonitor did help. Both nodes are notified by consul. So ha_notes are cluster-wide.
2. At the same time, I bumped into the initial issue why we had introduced ServiceMonitor. Confd just got stuck while starting. I'm investigating the logs now.

knekrasov:
@mssawant, I prepared patch that reverts the changes selectively https://github.com/Seagate/cortx-hare/pull/1380

So if we are not afraid of that issue I reproduced at 2-node cluster above, we probably can merge it.

mssawant:
@knekrasov, let me run some test with your patch.

PR #1379, opened 2020-10-21
EOS-14447 motr: update motr config post update

madhavemuri:
Currently after motr rpm is updated, motr config is not getting updated,
and new motr rpm installation will over-write the changes done
during deployment through "m0provision config" for initial deployment.

The solution is to add the motr post update steps in build-ees-ha-update

Signed-off-by: Madhavrao Vemuri <madhav.vemuri@seagate.com>

### DIFF
--- a/utils/build-ees-ha-update
+++ b/utils/build-ees-ha-update
@@ -65,11 +65,20 @@ reset_all() {
     $hare_exec/prov-ha-reset
 }
 
+motr_update() {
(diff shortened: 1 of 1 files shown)


madhavemuri:
Seems like rpm upgrade checks md5sum of file from already installed package against package to be 
upgrade with, and if md5sum not matched then only it will look for md5sum of physical file. 

And as there is no plan to change /etc/sysconfig/motr for Beta, closing this PR.

PR #1380, opened 2020-10-21
EOS-13678: servicemon is not needed

knekrasov:
Problem:
1. ServiceMonitor is no longer needed
2. ServiceMonitor filters out non-local services while ha_notes must be
cluster-wide

So it turns out that it makes sense to revert https://github.com/Seagate/cortx-hare/pull/1348 (selectively; logging improvements and ServiceHealth type should be kept).

Closes https://github.com/Seagate/cortx-hare/pull/1377

cc @mssawant


andriytk:
> ServiceMonitor is no longer needed

@knekrasov , could you please add a short explanation - why it's no longer needed (just for the record). Thanks!

andriytk (approved):

mssawant:
@knekrasov, @andriytk, bootstrapping is working fine but I am seeing problems as follows,
motr ioservice restart:
- Here the service restart is quick, sometimes I see the event being missed by the consul service watcher, did not find the event logged to the /var/log/hare/consul-watch-handler.log as well.
   For this hax still receives M0_CONF_HA_PROCESS_[STOPPING, STOPPED, STARTING, STARTED] events. Hax handles this through ConfHaProcess event and we can broadcast the event to local processes.
But I am still trying to confirm if Consul misses this event or somehow gets it but broadcasts it with a delay.

motr serivice killed and restarted:

In this case most of the time Consul watch handler event was seen and broadcast did happen to all the motr processes including client but I see that server.py:`process_ha_states()` broadcasts the event but does not wait for replies. 
```
def process_ha_states(queue: Queue):
    async def _process(request):
        data = await request.json()

        loop = asyncio.get_event_loop()
        # Note that queue.put is potentially a blocking call
        await loop.run_in_executor(
            None, lambda: queue.put(
                BroadcastHAStates(states=to_ha_states(data), reply_to=None)))
        return web.Response()
```
Here I saw that first hax broadcasts FAILED to active motr processes and immediately if `online` event is received ha broadcasts `OK` . Now even after this client IO fails, so I suspect that the events were not received in order somehow by the motr processes and client rpc connection still remains as cancelled. in my patch https://github.com/Seagate/cortx-hare/pull/1377, this case works fine probably because we are using EQ/BQ to broadcast which waits for replies after broadcast.
I think this (out of order events received by motr processes) situation is also affecting the case 1 (motr restart).

So overall @knekrasov, I think we can merge this patch as this patch is anyway needed but I think we still need some additional fixes.

mssawant (approved):

andriytk:
> But I am still trying to confirm if Consul misses this event...

@mssawant Consul polls the status of the service once a sec. So if the change happened quicker (i.e. some service stopped and re-started within a seconds) - it may be missed by Consul. This is expected.

Looks like we don't have any legal scenario (at least I'm not aware of) when we would need to restart the services manually outside of Pacemaker or to restart them quicker than 1 second. So I think we are fine about this so far.

When such a need will arise - we could think about it then.

andriytk:
> Here I saw that first hax broadcasts FAILED to active motr processes and immediately if online event is received ha broadcasts OK .

This is expected in the quick restarts when Consul still managed to notice them. Consul would send FAILED for already restarted process and then immediately Ok after it notices that the process is already started.

Again: it's because the quick restarts are not supported now.

mssawant:
Update:
Following is the situation:
- Started cluster
- Started client on node-1
- Data written
- Killed ioservice on node-2
- Consul event received (good)
- Node 2 ioservice started 
- Consul event received immediately via systemd, broadcast happened to all motr processes 
(including client) but m0d is not actually started yet.
- ioservice M0_HA_CONF_PROCESS_STARTED event received later but this is presently not a broadcast.
   (Client IO is still failing here)
```
m0client >>read 1048680 /tmp/srcfile1 4096 200 50
motr[32439]:  f650  ERROR  [rpc/session.c:576:m0_rpc_session_validate]  <! rc=-125 Cancelled session
IO failed! rc = -125
```
   - So added a local motr broadcast event on receiving motr process event.
      But this is not a Consul based event, hence the broadcast happens only locally and client does not receive the updated event.
      Client connection is still not established and IO still fails.
      This process event also needs to be a broadcast to all the motr processes in the cluster as in 
       https://github.com/Seagate/cortx-hare/pull/1377

cc @andriytk, @knekrasov

knekrasov:
@mssawant 
It seems like there are 2 sources of information: 
1. Consul notifications about the process status
2. Notifications that come from the m0d process itself (such as "M0_CONF_HA_PROCESS_STARTING").

And it seems like hax needs to have a more sophisticated logic to broadcast the proper ha_notes rather than simply forward the facts received via \[1].

What I can suggest:
1. Each hax process must store a map like [Process -> Its last known state] in its RAM.
2. Last known state can be one of (OFFLINE, STARTING, STARTED)
3. This state changes according to the following diagram:
image
4. ha_notes must be sent only when the state **changes** to OFFLINE or STARTED.
   * In this case hax must put a message to BQ. Then every hax will process that message and send out ha_notes as a result.


Notes:
1. This algorithm assumes there will be no timing issues with Consul notifications. In other words, we assume that Consul will never send "offline" about the service which is currently running.
2. The described state diagram can cause repeated "M0_NC_FAILED" notifications for the same process. Probably this is not an issue at all.
3. The state diagram doesn't tell how should hax set its internal state after failover. Perhaps we can rely on process statuses in our Consul KV. And that means that state diagram should affect Consul KV accordingly.

@andriytk, @rajanikantchirmade please share your thoughts on this proposal. Thank you!

andriytk:
> And it seems like hax needs to have a more sophisticated logic to broadcast the proper ha_notes rather than simply forward the facts received via [1].

@knekrasov it does not seem so to me (yet). It may be needed if we are to support the "quick services restarts" - yes. But do we really need to support it?..

andriytk:
> Consul event received immediately via systemd, broadcast happened to all motr processes
(including client) but m0d is not actually started yet.

@mssawant few questions:
1) did processes (including the client) receive NC_FAILED and then NC_ONLINE about the process? If yes
2) do you mean that NC_ONLINE was received before m0d was ready to serve requests and this caused some problem? If yes
3) why did it cause the problem? Do you think the [patch](https://github.com/Seagate/cortx-motr/pull/146) from @rajanikantchirmade may solve it?

mssawant:
Hi @knekrasov , landing these changes to main, I have found the problem and the fix. Will post a PR  over these changes.

PR #1382, opened 2020-10-23
EOS-13263: Hare utility to mask features and logrotate configuration is missing

mssawant:
- CSM requires Hare to report any unavailable features based on the
  installation type.
- Logrotate settings for hare logs must be configured based on the given installation
  type. Such a utility is missing in hare.
- Cron job for pruning hax m0trace files is also missing based on the installation
  type.

Solution:
- Add a utility to accept options for configuring logrotate and to report
masked features based on the installation type fetched from the provisioner.
Sample provisioner installation information,
```
{
"ret": {
"nodes": 1,
"server_type": "physical",
"servers_per_node": 2,
"storage_type": "5u84"
}
}
```
- Add a cron job script to prune /var/motr/hax/m0trace.* files based on the
  installation type (virtual or physical).
- Use asyncio to invoke csm api store_unsupported_features()
- Use cortx-utils package to use unsupported_features module.
- Add logrotate configuration file for `virtual` and `physical` setups.
- Install default hare logrotate configuration file as part of hare
  package installation. Default logrotate configuration is replaced with
  the setup (virtual or physical) specific logrotate configuration
  by setup.py utility through provisioner.

Signed-off-by: Mandar Sawant <mandar.sawant@seagate.com>

PR #1387, opened 2020-10-27
Revert "EOS-13263: Hare utility to mask features and logrotate configuration is missing"

rajanikantchirmade:
Reverts Seagate/cortx-hare#1382

Some regression is observed due to this change. Some of the packages required 
for this changes do not have their dependency got installed Hare component
deployment failed due to this (EOS-14651).


knekrasov:
@rajanikantchirmade Is this a temporal solution? I just wonder why we need to revert a good and required functionality at our side. It looks like the problem should be addressed from a problematic component instead?

knekrasov (approved):

azheregelya (approved):

chumakd (approved):

PR #1393, opened 2020-11-03
EOS-14192: IO fails after ioservice restart

mssawant:
When motr service (confd or ioservice) fails Consul detects the same via hare
    consul service watcher. The state is updated in the Consul services catalog which
    is then replicated to all the Consul nodes. Consul agents notify the the local hare
    hax service which then broadcasts the event to all the local motr processes.
    The failure broadcast was happening which helps motr clients to disconnect
    from the failed peer motr service but service online broadcast did not
    happen due to wrong status check. Due to this the motr cliens were not able
    to re-establish the connection with the restarted motr service.

    The consul service watch mechanism is based on systemd and watches systemd unit state.
    The watch is triggered every 1s and every change in systemd unit is detected and
    notified to hax service over http.

    Solution:
    Check for ServiceHealth.OK instead of 'online' while broadcasting process event to
    motr services.

### DIFF
--- a/hax/hax/motr/__init__.py
+++ b/hax/hax/motr/__init__.py
@@ -22,13 +22,13 @@ from errno import EAGAIN
 from typing import Any, List
 
 from hax.exception import ConfdQuorumException, RepairRebalanceException
-from hax.message import (BroadcastHAStates, EntrypointRequest, HaNvecGetEvent,
+from hax.message import (EntrypointRequest, HaNvecGetEvent,
                          ProcessEvent)
 from hax.motr.delivery import DeliveryHerald
 from hax.motr.ffi import HaxFFI, make_array, make_c_str
 from hax.types import (ConfHaProcess, Fid, FidStruct, FsStats, HaNote,
                        HaNoteStruct, HAState, MessageId, ObjT, ReprebStatus,
-                       StobIoqError)
+                       StobIoqError, ServiceHealth)
 from hax.util import ConsulUtil
 
 LOG = logging.getLogger('hax')
@@ -171,7 +171,7 @@ class Motr:
         cns = ConsulUtil()
 
         def ha_obj_state(st):
-            return HaNoteStruct.M0_NC_ONLINE if st.status == 'online' \
+            return HaNoteStruct.M0_NC_ONLINE if st.status == ServiceHealth.OK \
                 else HaNoteStruct.M0_NC_FAILED
 
         notes = []
@@ -195,10 +195,6 @@ class Motr:
                               chp_type=chp_type,
                               chp_pid=chp_pid,
                               fid=fid)))
-        if chp_event == 3:
-            self.queue.put(
-                BroadcastHAStates(states=[HAState(fid=fid, status='offline')],
-                                  reply_to=None))
 
     def _stob_ioq_event_cb(self, fid, sie_conf_sdev, sie_stob_id, sie_fd,
                            sie_opcode, sie_rc, sie_offset, sie_size,


hax/hax/motr/__init__.py
```diff
@@ -171,7 +171,7 @@ def broadcast_ha_states(self, ha_states: List[HAState]) -> List[MessageId]:
         cns = ConsulUtil()
 
         def ha_obj_state(st):
-            return HaNoteStruct.M0_NC_ONLINE if st.status == 'online' \
+            return HaNoteStruct.M0_NC_ONLINE if st.status == ServiceHealth.OK \
```
knekrasov, line 174: Nice finding, @mssawant  👍 
I wonder why mypy didn't notice this.

hax/hax/queue/__init__.py
```diff
@@ -51,11 +51,14 @@ def payload_process(self, msg: str) -> None:
             'M0_HA_MSG_NVEC': self.handle_device_state_set,
             'STOB_IOQ_ERROR': self.handle_ioq_stob_error,
         }
-        if msg_type not in handlers:
+        if msg_type in handlers:
+            handlers[msg_type](payload)
+        elif msg_type == 'HA_NOTIFY':
+            self.handle_ha_notify(payload)
```
knekrasov, line 57: Why can't we add 'HA_NOTIFY' to handlers dictionary? I believe `self.handle_ha_notify` has the same signature to other handlers.. And thus no extra 'elif' would be needed.

hax/hax/queue/__init__.py
```diff
@@ -96,3 +99,17 @@ def to_ha_state(self, objinfo: Dict[str, str]) -> Optional[HAState]:
...
+            hastatus = ServiceHealth.UNKNOWN
+            if state['status'] == 'FAILED':
+                hastatus = ServiceHealth.FAILED
+            elif state['status'] == 'OK':
+                hastatus = ServiceHealth.OK
+            hastates.append(HAState(Fid.parse(state['fid']), hastatus))
+
+        if hastates:
+            self.queue.put(BroadcastHAStates(states=hastates,
+                                             reply_to=None))
```
knekrasov, line 115: This call will not wait for delivery. Can't it cause problems in the future?

I guess we had to introduce DeliveryHerald specially for queue processing semantics: we don't want to mark a message in the queue as read until we can't be sure that the message is delivered to Motr world for sure. Isn't it the same case?


If so, this piece of code could look like this (draft, didn't check for correctness):
```python
        q: Queue = Queue(1)
        self.queue.put(
            BroadcastHAStates(BroadcastHAStates(states=hastates,
                                                reply_to=q))
```

hax/hax/types.py
```diff
@@ -216,3 +216,41 @@ def __repr__(self):
...
+        return self.name
+
+
+class m0HaProcessType(Enum):
+    M0_CONF_HA_PROCESS_OTHER = 0
+    M0_CONF_HA_PROCESS_KERNEL = 1
+    M0_CONF_HA_PROCESS_M0MKFS = 2
+    M0_CONF_HA_PROCESS_M0D = 3
+
+    def str_to_Enum(self):
```
knekrasov, line 248: Do you need this method? I didn't find in the sources you use it.

And anyway, it seems like you don't need it (at least, it could have been way simpler: `m0HaProcessType['<your string here>']`:
```python
>>> class testme(Enum):
...     VAL1 = 0
...     VAL2 = 1
...
>>> t =testme(1)
>>> t
<testme.VAL2: 1>
>>> t =testme('VAL1')
  File "/usr/lib64/python3.6/enum.py", line 293, in __call__
    return cls.__new__(cls, value)
  File "/usr/lib64/python3.6/enum.py", line 535, in __new__
    return cls._missing_(value)
  File "/usr/lib64/python3.6/enum.py", line 548, in _missing_
    raise ValueError("%r is not a valid %s" % (value, cls.__name__))
ValueError: 'VAL1' is not a valid testme
>>> t = testme['VAL1']
>>> t
<testme.VAL1: 0>
```

knekrasov (commented):
In general looks good, but there is a couple of questions to clarify.

andriytk (changes requested):
Looks like there are two changes in this PR: 1) bugfix for missing 'online' notification and 2) redesign of HA-notifications mechanism.

`+1` for the 1st change.

But I'm not sure about the 2nd one. Is it really needed? It overcomplicates the matter a lot from one side and, from another side - the use case seems kind of artificial one. I don't know any scenario when `systemctl restart m0d*` command would be called (except manually by a developer for some unknown reason). All the real-life scenarios involve: 1) the crash of m0d process 1st; 2) detection of this situation by Pacemaker (or some other processes management agent) and 3) restart. All this takes time which is very probably more than 1 second (the current check interval configured in Consul to monitor the services).

I'd advise parking the change 2 for now. Maybe it would appear useful later, but I hardly can see its usefulness at the moment.

hax/hax/queue/__init__.py
mssawant: Yeah, I did had waiting part earlier but removed to see how it affects. It worked but let's add waiting here too to maintain the semantics.

mssawant:
@andriytk, I think if there's a case where the functionality is broken the that needs to be fixed. There's one more weird behavior that I am seeing. So in following situation,
where node-1 is a consul leader and is I restart ioservice on node-2 then there's no notification by consul-watcher but If I restart an ioservice on node-1 (the leader node) the I see notification by consul-watcher, any idea about such a behavior?

andriytk:
@mssawant it's probably just intermittent result. Consul's check runs once a second the `systemctl status` command on the service. So if restart happened within a second - it might be missed.

But again - why do we bother about this scenario? Who is going to run `systemctl restart` command on the services and why?

mssawant:
@andriytk, Its not Consul its systemd, when I restart node-2 ioservice, systemd state is not changing and remains active where as the pid changes,
```
● m0d@0x7200000000000001:0x27.service - m0d helper for 0x7200000000000001:0x27 service-FID
   Loaded: loaded (/usr/lib/systemd/system/m0d@.service; static; vendor preset: disabled)
   Active: active (running) since Tue 2020-11-03 10:11:51 MST; 33s ago
 Main PID: 3183 (lt-m0d)
   CGroup: /system.slice/system-m0d.slice/m0d@0x7200000000000001:0x27.service
           └─3183 /home/520478/cortx-motr-new/cortx-motr/motr/.libs/lt-m0d -e lnet:192.168.9.107@tcp:12345:2:2 -A linuxstob:addb-stobs -f <0x7200000000000001:0x27> -T ad -S stobs -D db -m 65536 -q 16 -w 8 -H 192.168.9.107@tcp:12345:1:1 -U

Nov 03 10:11:51 ssc-vm-c-0553.colo.seagate.com motr-server[3183]: MOTR_PROCESS_FID: 0x7200000000000001:0x27
Nov 03 10:11:51 ssc-vm-c-0553.colo.seagate.com motr-server[3183]: MOTR_HA_EP: 192.168.9.107@tcp:12345:1:1
Nov 03 10:11:51 ssc-vm-c-0553.colo.seagate.com motr-server[3183]: WARN: query not existing genders var 'm0_max_rpc_msg_size'
Nov 03 10:11:51 ssc-vm-c-0553.colo.seagate.com motr-server[3183]: WARN: query not existing genders var 'm0_min_rpc_recvq_len'
Nov 03 10:11:51 ssc-vm-c-0553.colo.seagate.com motr-server[3183]: WARN: query not existing genders var 'm0_directio'
Nov 03 10:11:51 ssc-vm-c-0553.colo.seagate.com motr-server[3183]: vm.max_map_count = 30000000
Nov 03 10:11:51 ssc-vm-c-0553.colo.seagate.com motr-server[3183]: + exec /home/520478/cortx-motr-new/cortx-motr/motr/m0d -e lnet:192.168.9.107@tcp:12345:2:2 -A linuxstob:addb-stobs -f '<0x7200000000000001:0x27>' -T ad -S stobs -D db -m 65536 -q 16 -w 8 -H 192.168.9.107@tcp:12345:1:1 -U
Nov 03 10:11:52 ssc-vm-c-0553.colo.seagate.com motr-server[3183]: Started
Nov 03 10:11:52 ssc-vm-c-0553.colo.seagate.com motr-server[3183]: lt-m0d: systemd notifications not allowed
Nov 03 10:11:52 ssc-vm-c-0553.colo.seagate.com motr-server[3183]: Press CTRL+C to quit.
```

```
● m0d@0x7200000000000001:0x27.service - m0d helper for 0x7200000000000001:0x27 service-FID
   Loaded: loaded (/usr/lib/systemd/system/m0d@.service; static; vendor preset: disabled)
   Active: active (running) since Tue 2020-11-03 10:12:57      5s ago
 Main PID: 4161 (lt-m0d)
   CGroup: /system.slice/system-m0d.slice/m0d@0x7200000000000001:0x27.service
           └─4161 /home/520478/cortx-motr-new/cortx-motr/motr/.libs/lt-m0d -e lnet:192.168.9.107@tcp:12345:2:2 -A linuxstob:addb-stobs -f <0x7200000000000001:0x27> -T ad -S stobs -D db -m 65536 -q 16 -w 8 -H 192.168.9.107@tcp:12345:1:1 -U

Nov 03 10:12:57 ssc-vm-c-0553.colo.seagate.com motr-server[4161]: MOTR_PROCESS_FID: 0x7200000000000001:0x27
Nov 03 10:12:57 ssc-vm-c-0553.colo.seagate.com motr-server[4161]: MOTR_HA_EP: 192.168.9.107@tcp:12345:1:1
Nov 03 10:12:57 ssc-vm-c-0553.colo.seagate.com motr-server[4161]: WARN: query not existing genders var 'm0_max_rpc_msg_size'
Nov 03 10:12:57 ssc-vm-c-0553.colo.seagate.com motr-server[4161]: WARN: query not existing genders var 'm0_min_rpc_recvq_len'
Nov 03 10:12:57 ssc-vm-c-0553.colo.seagate.com motr-server[4161]: WARN: query not existing genders var 'm0_directio'
Nov 03 10:12:57 ssc-vm-c-0553.colo.seagate.com motr-server[4161]: vm.max_map_count = 30000000
Nov 03 10:12:57 ssc-vm-c-0553.colo.seagate.com motr-server[4161]: + exec /home/520478/cortx-motr-new/cortx-motr/motr/m0d -e lnet:192.168.9.107@tcp:12345:2:2 -A linuxstob:addb-stobs -f '<0x7200000000000001:0x27>' -T ad -S stobs -D db -m 65536 -q 16 -w 8 -H 192.168.9.107@tcp:12345:1:1 -U
Nov 03 10:12:58 ssc-vm-c-0553.colo.seagate.com motr-server[4161]: Started
Nov 03 10:12:58 ssc-vm-c-0553.colo.seagate.com motr-server[4161]: lt-m0d: systemd notifications not allowed
Nov 03 10:12:58 ssc-vm-c-0553.colo.seagate.com motr-server[4161]: Press CTRL+C to quit.
```

mssawant:
probably this is the reason,
```
restart PATTERN…
Stop and then start one or more units specified on the command line. If the units are not running yet, they will be started.

Note that restarting a unit with this command does not necessarily flush out all of the unit's resources before it is started again. For example, the per-service file descriptor storage facility (see FileDescriptorStoreMax= in systemd.service(5)) will remain intact as long as the unit has a job pending, and is only cleared when the unit is fully stopped and no jobs are pending anymore. If it is intended that the file descriptor store is flushed out, too, during a restart operation an explicit systemctl stop command followed by systemctl start should be issued.
```

mssawant:
I think I agree that probably there's no need to handle this case explicitly and m0d service restarting implicitly without changing the systemd state is very rare. May happen while testing but failing the service is an ideal scenario. Let me revert the other explicit handling changes.

hax/hax/queue/__init__.py
mssawant: Reverted these changes.

hax/hax/types.py
mssawant: Reverted changes.

hax/hax/queue/__init__.py
mssawant: Reverted changes.

andriytk (approved):

rajanikantchirmade (approved):

PR #1395, opened 2020-11-05
EOS-14192: IO fails after ioservice restart

mssawant:
When motr service (confd or ioservice) fails Consul detects the same via hare
    consul service watcher. The state is updated in the Consul services catalog which
    is then replicated to all the Consul nodes. Consul agents notify the the local hare
    hax service which then broadcasts the event to all the local motr processes.
    The failure broadcast was happening which helps motr clients to disconnect
    from the failed peer motr service but service online broadcast did not
    happen due to wrong status check. Due to this the motr cliens were not able
    to re-establish the connection with the restarted motr service.

    The consul service watch mechanism is based on systemd and watches systemd unit state.
    The watch is triggered every 1s and every change in systemd unit is detected and
    notified to hax service over http.

    Solution:
    Check for ServiceHealth.OK instead of 'online' while broadcasting process event to
    motr services.


hax/hax/motr/__init__.py
```diff
@@ -195,10 +195,6 @@ def _process_event_cb(self, fid, chp_event, chp_type, chp_pid):
                               chp_type=chp_type,
                               chp_pid=chp_pid,
                               fid=fid)))
-        if chp_event == 3:
```
knekrasov, line 198: I guess this check was added by @rajanikantchirmade.
Are you sure you don't need it?

andriytk (approved):

hax/hax/motr/__init__.py
mssawant: Yeah we don't need this now because this is happening automatically, hax notifies M0_NC_FAILED to motr process as hax receives notification from Consul regarding motr service state change in consul service catalog (failure/stop).

knekrasov (approved):

PR #1398, opened 2020-11-10
EOS-14739:fs_free_seg, fs_free_disk is not reducing after a huge s3 IOs

suvratjoshi:
The changes in this PR are for two issues :
1. The original one as reported in JIRA ticket :
In this, hax service had got stopped upon getting HAConsistencyException while trying to access Consul KV in function get_conf_obj_status. This had happened when the consul had lost the leader and new leader election again had happened. So, in this PR the decorator has been put for get_conf_obj_status to make retries forever.

2. The issue which I had seen during attempts to repro, and mentioned in the JIRA ticket comments (failure which had occurred during access to consul catalog) :
Similar to above, the decorator has been put around function "_ensure_motr_all_started".

The testing for this issue has been done by killing the consul process and checking that the retries are being made and recovery is successful and hax / fs-stats-updater doesn't get stopped.

The Jenkins build is passed for this PR branch changes :
http://eos-jenkins.colo.seagate.com/job/GitHub-custom-ci-builds/job/centos-7.8/job/cortx-custom-ci/142/
RPM used for testing on hardware : 
http://cortx-storage.colo.seagate.com/releases/cortx/github/integration-custom-ci/release/centos-7.8.2003/custom-build-142

### DIFF
--- a/hax/hax/filestats.py
+++ b/hax/hax/filestats.py
@@ -24,7 +24,7 @@ from typing import List
 from hax.exception import HAConsistencyException, InterruptedException
 from hax.motr import Motr, log_exception
 from hax.types import FsStatsWithTime, StoppableThread
-from hax.util import ConsulUtil
+from hax.util import ConsulUtil, repeat_if_fails
 
 LOG = logging.getLogger('hax')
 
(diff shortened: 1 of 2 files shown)


mssawant:
@suvratjoshi, fstatsupdater in hax updates Consul every 5 secs and HAConsistencyException is absorbed. So even if hax is not able to update the Consul the update will happen in net iteration.
Question is why the Consul failed here and had to re-elect the leader? I think confd failed. Why the confd failed any idea?

Also, It would be good if you re-format the commit message as follows,
if there's no jira ticket,
```
Problem: <summary>
```
else
```
EOS-<ticket#>: Summary of the problem
```
```
<Brief  description of the problem.>
Solution:
<description> of the solution.
```

knekrasov (approved):

mssawant:
@knekrasov, I think I understand what you are talking about, the HAConsistency exception is handled inside if loop in while, only for `self.consul.update_fs_stats(data)` call and not outside it. So if the call to `_ensure_motr_all_started()` fails which is outside the while and if loop then the fsStatsUpdater may bail out.
But then the same problem can be seen for `_ioservices_running()` too, if so then that also needs to be handled, no?

knekrasov:
@mssawant Hm, it seems like yes, you're right. Good catch!

mssawant (approved):

PR #1399, opened 2020-11-11
Problem: Hax consumer thread logs long debug messages

rajanikantchirmade:
Problem: Hax consumer thread logs long debug messages

Hax consymer thread writes long debug log messages. It creates
IEM processing in SSPL slow. This needs to be fixed by writing
proper optimised messages.

Solution:
Skip long hax debgug messages.

### DIFF
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -77,7 +77,6 @@ class ConsumerThread(StoppableThread):
                             raise StopIteration()
                         item = pull_msg()
 
-                    LOG.debug('Got %s message from queue', item)
                     if isinstance(item, EntrypointRequest):
                         # While replying any Exception is catched. In such a
                         # case, the motr process will receive EAGAIN and


hax/hax/handler.py
```diff
@@ -77,7 +77,17 @@ def pull_msg():
...
+                    # @todo XXX Fixme.
+                    # Some of items in event queue are long
+                    # e.g HaNvecGetEvent, BroadcastHAStates these may contains
+                    # long list of conf objects fids and their stats, and as
+                    # size of cluster increses, size of these events increses.
+                    # Lets skip these messages to unblock IEM issue and need
+                    # proper optimised logs
+                    if not (isinstance(item, HaNvecGetEvent) or
+                            isinstance(item, BroadcastHAStates)):
+                        LOG.debug('Got %s message from queue', item)
```
knekrasov, line 89: @rajanikantchirmade please remove this LOG.debug line at all. I don't think it makes much sense to log it here. Sadly, in Python logging subsystem there is no "trace" verbosity level: such a message should be go there (and seen only when we're debugging something).

knekrasov (changes requested):

hax/hax/handler.py
mssawant: @knekrasov, @rajanikantchirmade, I think need a wrapper over our present logging system and provide a hctl command to set the log level in hax. So we can have a global log level specified which can be default to ERROR for now so that only ERROR and exceptions are logged while all the other log level messages are ignored.
One should be able to get and set the log level using hctl command at the runtime. 
This would take sometime but I think it would be better than to removing the debug messages and re-adding them during debugging and rebuilding the code. I think this would help testing too.

rajanikantchirmade: @knekrasov   Yes, it make sense to remove this log completely.

@mssawant  Yes, we need some improvement there. But now lets unblock SSPL guys and later  Handler can have proper optimized logs.

mssawant (approved):

NeeravChoudhari (approved):

chumakd (approved):

PR #1402, opened 2020-11-17
bootstrap-node: fix hax startup of client-only nodes

andriytk:
Currently, hax starts on the nodes which run confd or ios
which is wrong. Client-only nodes should have hax running also.

Solution: fix bootstrap-node to start hax always on the node,
regardless of its configuration.

Signed-off-by: Andriy Tkachuk <andriy.tkachuk@seagate.com>

### DIFF
--- a/utils/bootstrap-node
+++ b/utils/bootstrap-node
@@ -156,11 +156,10 @@ wait_m0d_started() {
     done
 }
 
-if [[ -n $CONFD_IDs || -n $IOS_IDs ]]; then
-    sudo mkdir -p /var/motr/hax
-    sudo systemctl start hare-hax
-    wait_active hare-hax 5
-fi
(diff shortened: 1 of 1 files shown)


mssawant:
@andriytk, what about consul agent? does consul agent start on the client node ?

andriytk:
Yes, it does.

mssawant (approved):

mssawant:
oh yes, I forgot, sorry for the noise.
```
while read node bind_ip; do
    ssh $node "PATH=$PATH $(which mk-consul-env) \
                          --mode client --bind $bind_ip --join $join_ip &&
               sudo systemctl start hare-consul-agent" &
    pids+=($!)
done < <(get_client_nodes)
```

chumakd (approved):

PR #1405, opened 2020-11-24
Problem: hax creates multiple ConsulUtil instances

andriytk:
Various sub-modules of hax create multiple ConsulUtil instances which in turn create multiple Consul objects. These objects are usually destroyed by Python's GC, but sometimes there might be lagging and the number of such connections would grow beyond 200 - the default limit in Consul. After this, nobody could connect
to Consul.

Solution: re-use the ConsulUtil object instantiated by hax.py in various other hax submodules.

### DIFF
--- a/hax/hax/hax.py
+++ b/hax/hax/hax.py
@@ -59,8 +59,9 @@ def _run_qconsumer_thread(queue: Queue, motr: Motr) -> ConsumerThread:
     return thread
 
 
-def _run_stats_updater_thread(motr: Motr) -> FsStatsUpdater:
-    thread = FsStatsUpdater(motr, interval_sec=30)
+def _run_stats_updater_thread(motr: Motr,
+                              consul_util: ConsulUtil) -> FsStatsUpdater:
+    thread = FsStatsUpdater(motr, consul_util, interval_sec=30)
     thread.start()
     return thread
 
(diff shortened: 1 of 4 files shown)


utils/consul-metrics-logger
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
mssawant, line 1: This is a strayed file that got into the latter update to the patch. Sorry. I have removed it from https://github.com/Seagate/cortx-hare/pull/1404, kindly update.
Thank you.

mssawant (approved):

utils/consul-metrics-logger
andriytk: Oh, Ok. Don't you think it should be kept?

andriytk: Ah, sorry - it's an empty file! :) Will remove it also.

azheregelya (approved):

andriytk:
Tested on the SAGE cluster, seems working.

hax/hax/hax.py
```diff
@@ -112,10 +114,11 @@ def main():
                    ha_service=cfg.ha_fid,
                    rm_service=cfg.rm_fid)
         LOG.info('Motr API has been started')
-        stats_updater = _run_stats_updater_thread(motr)
+        stats_updater = _run_stats_updater_thread(motr, consul_util=util)
```
mssawant, line 117: **optional** : Named argument is not required here.

PR #1409, opened 2020-11-29
EOS-15336: Delay in entrypoint processing leads to motr process failure

mssawant:
Solution:
Do not queue the FAILED notification instead broadcast it directly.

### DIFF
--- a/hax/hax/motr/__init__.py
+++ b/hax/hax/motr/__init__.py
@@ -20,7 +20,7 @@ import ctypes as c
 import logging
 from errno import EAGAIN
 from typing import Any, List
-from queue import Queue
+# from queue import Queue
 
 from hax.exception import ConfdQuorumException, RepairRebalanceException
 from hax.message import (BroadcastHAStates, EntrypointRequest, HaNvecGetEvent,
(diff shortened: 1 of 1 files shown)


hax/hax/motr/__init__.py
```diff
@@ -20,7 +20,7 @@
 import logging
 from errno import EAGAIN
 from typing import Any, List
-from queue import Queue
+# from queue import Queue
```
knekrasov, line 23: No need to comment the code out (unless this is a testing of a WIP branch, of course). Please remove the code you don't need.

hax/hax/motr/__init__.py
```diff
@@ -126,14 +126,12 @@ def _entrypoint_request_cb(self, reply_context: Any, req_id: Any,
...
-                q: Queue = Queue(1)
+                # q: Queue = Queue(1)
                 LOG.debug('first entrypoint request, broadcasting FAILED')
-                self.queue.put(
-                    BroadcastHAStates(
-                        states=[HAState(fid=process_fid,
-                                        status=ServiceHealth.FAILED)],
-                        reply_to=q))
-                ids: List[MessageId] = q.get()
+                states = [HAState(fid=process_fid,
```
knekrasov, line 131: Do you have an idea why you experience a delay? Can't it be a symptom that the Queue capacity is too low and we need to enlarge it?

Seriously, initially the queue size was chosen quite small - 8 objects or so. I can imagine that under a high load (when there are many Motr processes to communicate to and thus many events are happening, the size can be a bottleneck.

The idea is that when the Queue is full and a thread is trying to add a message there, the thread is blocked until some free slot in the queue becomes available.

knekrasov: Another idea: if we do depend on the speed of reaction from hax, we can add more "consumer" threads. But before doing that we need to make sure that this doesn't bring other side effects.

For instance, currently all Motr-related jobs are processed one-by-one in the order they were required. If we start processing them in parallel, can this bring issues to Motr world? If there is no issue for sure, I think we should enlarge the number of threads.

@andriytk, @max-seagate, @chumakd Thoughts?

andriytk:
Why is the delay happens and why it leads to the motr process failure?

vaibhavparatwar:
@mssawant do we need this PR now?

Issue #1410, opened 2020-11-30
Problem: Python is not suitable for Hax


andriytk:
The main problem is that Python does not do the static type checking. On another day I was trying to make some minor change and missed to provide an argument to the function - I could detect the issue only after building the new rpm, deploying it into the system and running the code. It's very inconvenient and time-consuming in the long run.

Suggested solution: re-write `hax` in Go language. Go is an easy-to-learn system programming language with a strong static type-checking. It has very good bindings with C and with Consul (which is written in Go) - something that is essential for `hax`.

andriytk:
cc @NeeravChoudhari , @knekrasov , @chumakd , @azheregelya .

knekrasov:
@andriytk Can you describe more what issue you faced and failed to find with mypy?

The thing is that Python has type checks and the static analysis is already involved into the build procedure. So strictly speaking, we can't say that there is no static type checking.

Another question: can anybody recall why Python was chosen for Hare? What objectives/requirements were there at the moment? I think we need to consider those objectives and thoughts while choosing a new language anyway.

andriytk:
Python was chosen because it was thought to be simple to program in and that many people are acquainted with it. That was the main reason, as far as I remember. We just needed something to quickly prototype in and that most people could read and understand.

andriytk:
Here is the change with the missing argument to the function which builds without any issues:

```diff
--- a/hax/hax/hax.py
+++ b/hax/hax/hax.py
@@ -117,7 +117,7 @@ def main():
         stats_updater = _run_stats_updater_thread(motr, consul_util=util)
         # [KN] This is a blocking call. It will work until the program is
         # terminated by signal
-        run_server(q, herald, consul_util=util,
+        run_server(q, herald,
                    threads_to_wait=[consumer, stats_updater])
     except Exception:
         LOG.exception('Exiting due to an exception')
```

vvv:
> can anybody recall why Python was chosen for Hare? What objectives/requirements were there at the moment?

Team members could program in C, Bash, Haskell, and Python.  Our priority was to deliver fast; learning new programming language was a luxury we didn't have.  Seagate upper management, having their fingers burnt with Haskell, became very conservative about the choice of technologies.  I guess Go lang would be a leap too far in those circumstances, but, to be honest, I don't remember it even to be considered by team members.  We didn't think [that far ahead].

[that far ahead]: https://youtu.be/-X5oCV7hWgs?t=10m54s

---

[...]

PR #1411, opened 2020-12-01
 Problem: mypy doesn't detect missing args of function

knekrasov:
Solution:
1. make mypy behavior stricter. Use `--check-untyped-defs` flags
everywhere in the build pipeline. More details on this solution: https://mypy.readthedocs.io/en/stable/common_issues.html#no-errors-reported-for-obviously-wrong-code)
2. fix the type errors uncovered by enabling this mode.

The issue was revealed in #1410

### DIFF
--- a/cfgen/Makefile
+++ b/cfgen/Makefile
@@ -33,8 +33,9 @@ flake8: $(PYTHON_SCRIPTS)
 	flake8 $(FLAKE8_OPTS) $^
 
 .PHONY: typecheck
+override MYPY_OPTS := --config-file ../hax/mypy.ini $(MYPY_OPTS)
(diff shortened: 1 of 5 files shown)


andriytk (approved):

chumakd:
Suggestion: add this configuration option into `hax/mypy.ini` config file, like this:

```ini
[mypy]
check_untyped_defs = True
```
That way it doesn't have to be repeated for every `mypy` command in the `Makefile`.

chumakd (approved):

knekrasov:
Thanks @chumakd! Adding this parameter to mypy.ini has also fixed my vim environment 😁

PR #1419, opened 2020-12-02
EOS-15360: motr status is not updated in Consul in case of failure

mssawant:
Solution:
Update Consul status in case of motr failure.

### DIFF
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -28,7 +28,8 @@ from hax.message import (BroadcastHAStates, EntrypointRequest, HaNvecGetEvent,
                          SnsRepairStart, SnsRepairStatus, SnsRepairStop)
 from hax.motr import Motr
 from hax.queue.publish import EQPublisher
-from hax.types import MessageId, StobIoqError, StoppableThread
+from hax.types import (ConfHaProcess, HAState, MessageId, StobIoqError,
+                       m0HaProcessEvent, ServiceHealth, StoppableThread)
 from hax.util import ConsulUtil, dump_json, repeat_if_fails
 
 LOG = logging.getLogger('hax')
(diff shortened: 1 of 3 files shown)


knekrasov:
@mssawant Is this PR still relevant? Maybe it is time to merge it?

hax/hax/handler.py
```diff
@@ -50,14 +51,38 @@ def __init__(self, q: Queue, motr: Motr):
...
+    def check_state_and_update(self, ha_states: List[HAState]) -> None:
+        for state in ha_states:
+            if state.status == ServiceHealth.FAILED:
+                m0status = m0HaProcessEvent.M0_CONF_HA_PROCESS_STOPPED
+                pevent = ConfHaProcess(chp_event=m0status,
+                                       chp_type=3,
+                                       chp_pid=0,
+                                       fid=state.fid)
+                self._update_process_status(pevent)
+                LOG.info('MANDAR fid: %s state: %s',
```
knekrasov, line 78: I guess this is a debug logging

hax/hax/handler.py
```diff
@@ -50,14 +51,38 @@ def __init__(self, q: Queue, motr: Motr):
...
+                         state.fid, m0status)
+
     def _do_work(self, q: Queue, motr: Motr):
         ffi = motr._ffi
         LOG.info('Handler thread has started')
         ffi.adopt_motr_thread()
+        self.statuses = self.consul.get_m0d_statuses()
```
knekrasov, line 85: How `self.statuses` are used? What are they for?

hax/hax/types.py
```diff
@@ -218,21 +218,22 @@ def __repr__(self):
 HAState = NamedTuple('HAState', [('fid', Fid), ('status', ServiceHealth)])
 
 
-class m0HaProcessEvent(Enum):
+class m0HaProcessEvent(IntEnum):
```
knekrasov, line 221: Why IntEnum is more suitable than Enum?

mssawant:
@knekrasov, This problem is fixed by fd1b6fad, will abandon this patch as it is obsolete and no more required.
Thanks for bringing this up.

PR #1422, opened 2020-12-04
Problem: Consul state is not updated in case of process failure

mssawant:
Motr process states are persisted in Consul by hax as and when the motr
    process updates the same to hax. But the process failure is not updated
    in the Consul when a motr process crashes abrubtly, in which case no
    process event is sent.
    fsStatsUpdater thread in the hax checks the motr processes (confd, ios)
    state in Consul and if all are STARTED then it starts rconfc and tries to
    fetch statistics from motr. So if the failure state is not updated in Consul
    fsStatsUpdater always see the motr process states as STARTED and keeps trying
    in which case rconfc fails too.

    Solution:
    Update process failuer in Consul by updating the motr process state to
    M0_CONF_HA_PROCESS_STOPPED.

### DIFF
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -21,14 +21,18 @@ import time
 from queue import Empty, Queue
 from typing import List
 
-from hax.message import (BroadcastHAStates, EntrypointRequest, HaNvecGetEvent,
+from hax.message import (BroadcastHAStates, FirstEntrypointRequest,
+                         EntrypointRequest, HaNvecGetEvent,
                          ProcessEvent, SnsRebalancePause, SnsRebalanceResume,
                          SnsRebalanceStart, SnsRebalanceStatus,
                          SnsRebalanceStop, SnsRepairPause, SnsRepairResume,
                          SnsRepairStart, SnsRepairStatus, SnsRepairStop)
 from hax.motr import Motr
+from hax.motr.delivery import DeliveryHerald
 from hax.queue.publish import EQPublisher
-from hax.types import MessageId, StobIoqError, StoppableThread
+from hax.types import (ConfHaProcess, HAState, MessageId, StobIoqError,
+                       m0HaProcessEvent, ServiceHealth, StoppableThread,
+                       HaLinkMessagePromise)
 from hax.util import ConsulUtil, dump_json, repeat_if_fails
 
 LOG = logging.getLogger('hax')
@@ -43,16 +47,36 @@ class ConsumerThread(StoppableThread):
     The thread exits gracefully when it receives message of type Die (i.e.
     it is a 'poison pill').
     """
-    def __init__(self, q: Queue, motr: Motr):
+    def __init__(self, q: Queue, motr: Motr, herald: DeliveryHerald):
         super().__init__(target=self._do_work,
                          name='qconsumer',
                          args=(q, motr))
         self.is_stopped = False
         self.consul = ConsulUtil()
         self.eq_publisher = EQPublisher()
+        self.herald = herald
 
     def stop(self) -> None:
         self.is_stopped = True
+
+    @repeat_if_fails(wait_seconds=1)
+    def _update_process_status(self, event: ConfHaProcess) -> None:
+        # If a consul-related exception appears, it will
+        # be processed by repeat_if_fails.
+        #
+        # This thread will become blocked until that
+        # intermittent error gets resolved.
+        self.consul.update_process_status(event)
+
+    def update_process_failure(self, ha_states: List[HAState]) -> None:
+        for state in ha_states:
+            if state.status == ServiceHealth.FAILED:
+                m0status = m0HaProcessEvent.M0_CONF_HA_PROCESS_STOPPED
+                pevent = ConfHaProcess(chp_event=m0status,
+                                       chp_type=3,
+                                       chp_pid=0,
+                                       fid=state.fid)
+                self._update_process_status(pevent)
 
     def _do_work(self, q: Queue, motr: Motr):
         ffi = motr._ffi
@@ -78,20 +102,30 @@ class ConsumerThread(StoppableThread):
                         item = pull_msg()
 
                     LOG.debug('Got %s message from queue', item)
-                    if isinstance(item, EntrypointRequest):
+                    if isinstance(item, FirstEntrypointRequest):
+                        LOG.debug('first entrypoint request, broadcast FAILED')
+                        ids: List[MessageId] = motr.broadcast_ha_states(
+                            [HAState(fid=item.process_fid,
+                                     status=ServiceHealth.FAILED)])
+                        LOG.debug('waiting for broadcast of %s for ep: %s',
+                                  ids, item.remote_rpc_endpoint)
+                        self.herald.wait_for_all(HaLinkMessagePromise(ids))
+                        motr.send_entrypoint_request_reply(
+                            EntrypointRequest(
+                                reply_context=item.reply_context,
+                                req_id=item.req_id,
+                                remote_rpc_endpoint=item.remote_rpc_endpoint,
+                                process_fid=item.process_fid,
+                                git_rev=item.git_rev,
+                                pid=item.pid,
+                                is_first_request=item.is_first_request))
+                    elif isinstance(item, EntrypointRequest):
                         # While replying any Exception is catched. In such a
                         # case, the motr process will receive EAGAIN and
                         # hence will need to make new attempt by itself
                         motr.send_entrypoint_request_reply(item)
                     elif isinstance(item, ProcessEvent):
-                        fn = self.consul.update_process_status
-                        # If a consul-related exception appears, it will
-                        # be processed by repeat_if_fails.
-                        #
-                        # This thread will become blocked until that
-                        # intermittent error gets resolved.
-                        decorated = (repeat_if_fails(wait_seconds=5))(fn)
-                        decorated(item.evt)
+                        self._update_process_status(item.evt)
                     elif isinstance(item, HaNvecGetEvent):
                         fn = motr.ha_nvec_get_reply
                         # If a consul-related exception appears, it will
@@ -105,6 +139,7 @@ class ConsumerThread(StoppableThread):
                         LOG.info('HA states: %s', item.states)
                         result: List[MessageId] = motr.broadcast_ha_states(
                             item.states)
+                        self.update_process_failure(item.states)
                         if item.reply_to:
                             item.reply_to.put(result)
                     elif isinstance(item, StobIoqError):
--- a/hax/hax/motr/__init__.py
+++ b/hax/hax/motr/__init__.py
@@ -20,17 +20,17 @@ import ctypes as c
 import logging
 from errno import EAGAIN
 from typing import Any, List
-from queue import Queue
 
 from hax.exception import ConfdQuorumException, RepairRebalanceException
-from hax.message import (BroadcastHAStates, EntrypointRequest, HaNvecGetEvent,
+from hax.message import (BroadcastHAStates, FirstEntrypointRequest,
+                         EntrypointRequest, HaNvecGetEvent,
                          ProcessEvent)
 from hax.motr.delivery import DeliveryHerald
 from hax.motr.ffi import HaxFFI, make_array, make_c_str
 from hax.types import (ConfHaProcess, Fid, FidStruct, FsStats, HaNote,
                        HaNoteStruct, HAState, MessageId, ObjT, ReprebStatus,
                        StobIoqError, ServiceHealth, m0HaProcessEvent,
-                       m0HaProcessType, HaLinkMessagePromise)
+                       m0HaProcessType)
 from hax.util import ConsulUtil
 
 LOG = logging.getLogger('hax')
@@ -113,8 +113,8 @@ class Motr:
                                                    str(process_fid)) +
                   ' The request will be processed in another thread.')
         try:
-            cns = ConsulUtil()
-            if is_first_request and (not cns.is_proc_client(process_fid)):
+            if (is_first_request and
+                    (not self.consul_util.is_proc_client(process_fid))):
                 # This is the first start of this process or the process has
                 # restarted.
                 # Let everyone know that the process has restarted so that
@@ -126,19 +126,20 @@ class Motr:
                 # anyway detect the failure and report the same so we exclude
                 # reporting the same during their first entrypoint request.
                 # But we need to do it for motr server processes.
-                q: Queue = Queue(1)
-                LOG.debug('first entrypoint request, broadcasting FAILED')
                 self.queue.put(
-                    BroadcastHAStates(
-                        states=[HAState(fid=process_fid,
-                                        status=ServiceHealth.FAILED)],
-                        reply_to=q))
-                ids: List[MessageId] = q.get()
-                LOG.debug('waiting for broadcast of %s ep: %s',
-                          ids, remote_rpc_endpoint)
-                self.herald.wait_for_all(HaLinkMessagePromise(ids))
+                    FirstEntrypointRequest(
+                        reply_context=reply_context,
+                        req_id=req_id,
+                        remote_rpc_endpoint=remote_rpc_endpoint,
+                        process_fid=process_fid,
+                        git_rev=git_rev,
+                        pid=pid,
+                        is_first_request=is_first_request
+                    ))
+                return
         except Exception:
-            pass
+            LOG.exception('Failed to notify failure for %s',
+                          process_fid)
 
         LOG.debug('enqueue entrypoint request for %s',
                   remote_rpc_endpoint)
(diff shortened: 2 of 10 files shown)


hax/hax/handler.py
```diff
@@ -50,10 +51,31 @@ def __init__(self, q: Queue, motr: Motr):
         self.is_stopped = False
         self.consul = ConsulUtil()
         self.eq_publisher = EQPublisher()
+        self.statuses: List = []
```
knekrasov, line 54: It seems like this `statuses` is obsolete and not actually used. Is it correct?

hax/hax/handler.py
```diff
@@ -50,10 +51,31 @@ def __init__(self, q: Queue, motr: Motr):
...
 
     def stop(self) -> None:
         self.is_stopped = True
 
+    def _update_process_status(self, event: ConfHaProcess) -> None:
```
knekrasov, line 59: [Optional]
This version looks shorter and more readable:
```python
    @repeat_if_fails(wait_seconds=1)
    def _update_process_status(self, event: ConfHaProcess) -> None:
        # If a consul-related exception appears, it will
        # be processed by repeat_if_fails.
        #
        # This thread will become blocked until that
        # intermittent error gets resolved.
        self.consul.update_process_status(event)
```

hax/hax/handler.py
```diff
@@ -102,11 +117,13 @@ def pull_msg():
                         decorated = (repeat_if_fails(wait_seconds=5))(fn)
                         decorated(item)
                     elif isinstance(item, BroadcastHAStates):
-                        LOG.info('HA states: %s', item.states)
+                        LOG.info('HA states: %s Current statuses: %s',
```
knekrasov, line 120: It looks like a debug logging to me. If it doesn't help in common maintenance activities, please remove it. Otherwise please use `LOG.debug` instead.

hax/hax/motr/__init__.py
```diff
@@ -110,6 +112,36 @@ def _entrypoint_request_cb(self, reply_context: Any, req_id: Any,
                   " '{}', process fid = {}".format(remote_rpc_endpoint,
                                                    str(process_fid)) +
                   ' The request will be processed in another thread.')
+        try:
+            cns = ConsulUtil()
```
knekrasov, line 116: `ConsulUtil` will be created every time a new entrypoint request comes which is probably not the best idea.

Please also note that `self.consul_util` is accessible here. BTW I guess it was you who added it 😉

hax/hax/motr/__init__.py
```diff
@@ -110,6 +112,36 @@ def _entrypoint_request_cb(self, reply_context: Any, req_id: Any,
...
+                    BroadcastHAStates(
+                        states=[HAState(fid=process_fid,
+                                        status=ServiceHealth.FAILED)],
+                        reply_to=q))
+                ids: List[MessageId] = q.get()
+                LOG.debug('waiting for broadcast of %s for ep: %s',
+                          ids, remote_rpc_endpoint)
+                self.herald.wait_for_all(HaLinkMessagePromise(ids))
+        except Exception:
+            pass
```
knekrasov, line 141: Please log an exception. Silent swallowing makes a problem during support.

hax/hax/motr/__init__.py
```diff
@@ -110,6 +112,36 @@ def _entrypoint_request_cb(self, reply_context: Any, req_id: Any,
...
+                LOG.debug('first entrypoint request, broadcasting FAILED')
+                self.queue.put(
+                    BroadcastHAStates(
+                        states=[HAState(fid=process_fid,
+                                        status=ServiceHealth.FAILED)],
+                        reply_to=q))
+                ids: List[MessageId] = q.get()
+                LOG.debug('waiting for broadcast of %s for ep: %s',
+                          ids, remote_rpc_endpoint)
+                self.herald.wait_for_all(HaLinkMessagePromise(ids))
```
knekrasov, line 139: Is it safe to invoke `wait_for_all` in Motr locality thread (note that we're still in Motr-land thread)? It will be blocked by this call for potentially a long time.

More than that, please note that this particular function is invoked from `hax.c` where GIL is locked. It is locked until `def _entrypoint_request_cb` exits. That means that NO other Python thread will be active until that moment. The whole hax application will freeze until `wait_for_all()` gets unblocked - that will happen from another Motr threads, so there is no deadlock. But freezing is still a bad thing in my opinion.

Suggestion: move this logic to ConsumerThread.

hax/hax/motr/delivery.py
```diff
@@ -69,14 +69,45 @@ def wait_for_any(self,
...
+                LOG.debug('Blocking until %s is confirmed', promise)
+                condition.wait(timeout=timeout_sec)
+            with self.lock:
+                if promise not in self.recently_delivered:
+                    raise NotDelivered('None of message tags =' +
+                                       str(promise) +
+                                       '  were delivered to Motr')
+                msgid = self.recently_delivered[promise]
+                if msgid in promise:
+                    promise._ids.remove(msgid)
```
knekrasov, line 96: Not sure that this is needed. If we don't remove the id from the promise, the algorithm will keep working, but no CPU time will be spent on removal.

hax/hax/motr/delivery.py
```diff
@@ -69,14 +69,45 @@ def wait_for_any(self,
...
+                LOG.debug('Thread unblocked - %s just received',
+                          self.recently_delivered[promise])
+                del self.recently_delivered[promise]
```
knekrasov, line 99: ...but deletion of the promise must happen only when it is the last iteration

Otherwise there is potentially a time window when this promise is removed from the dict but a delivery happens.

hax/hax/motr/hax.c
```diff
@@ -541,7 +541,21 @@ static void msg_is_delivered_cb(struct m0_halon_interface *hi,
...
+
+	PyGILState_STATE gstate;
+	gstate = PyGILState_Ensure();
+
+	/* Notify hax about delivery failure. */
+	M0_LOG(M0_DEBUG, "msg not delivered, tag=%"PRIu64,
+		hl->hln_tag_broadcast_delivery);
+
+	PyObject *py_fid = toFid(proc_fid);
+	PyObject_CallMethod(hc0->hc_handler, "_msg_delivered_cb", "(OsKK)",
```
knekrasov, line 554: I think it makes sense (at least) to add a bool parameter saying whether it is delivered (success story) or not delivered. I understand that at this point it doesn't make much sense since the only use case when delivery callback is used (at the moment) is DeliveryHerald mechanism support. But I would still prefer to keep the intentions and semantics clear. If it is not delivered, let's not hide the truth - it wasn't delivered.

knekrasov (changes requested):

hax/hax/motr/__init__.py
mssawant: I think there's some problem in merging to main here. I overwrote the earlier version. Let me fix this.

hax/hax/motr/__init__.py
mssawant: Agree but we want to wait until the FAILED notifications are delivered to make sure there's no race between entrypoint response and FAILED notification.
Actually there's a bug here (known). Ideally this FAILED broadcast needs to be to entire cluster and not just local. I have an implementation for this in works but its not working cleanly yet.
Presently we do not have a way to block asynchronously until the FAILED notifications are delivered. I am trying a callback mechanism along with using BQ. I wanted to discuss this with you actually.

mssawant: But for now I want to get the basic changes in then fix the global broadcast case separately.

hax/hax/motr/hax.c
mssawant: I think then let's have a corresponding callback function _msg_not_delivered_cb() to be more explicit.

knekrasov: Sounds good to me.

hax/hax/motr/__init__.py
knekrasov: I don't quite understand, why can't we just move all this processing to another thread. If `self.herald.wait_for_all` is invoked from a Pythonic thread (not Motr one), the problem of the application freeze will disappear.

mssawant: Yes we can, then we need to move posting entrypoint request where ever we move FAILED notification and wait_for_all. We can do this by introducing a new instance type that will also post entrypoint request after waiting_for_all deliveries.
But if we move this to Consumer thread then Consumer thread will be blocked, right?

knekrasov: Yes, that's true. But in that case ConsumerThread will be the **only** thread being blocked at that point.

In your code, since the thread is external, nobody will release GIL till Condition gets triggered. That means that all Python threads are frozen until Motr finishes its dirty deals 😁

mssawant: Sorry but I need to understand how come GIL is released when we wait in ConsumerThread and why it is not released when we wait in motr thread.
But overall I see one problem with this, motr has a thread per locality (locality is per cpu core) and every type of operation is assigned a locality. In this case lets consider 2 operations hlink broadcast and entrypoint request. Now, in motr hlink operation and entrypoint operation both run in the same locality and thus share the thread. So if one operation blocks (entrypoint) other operation will get affected (hlink broadcast), I was having this suspicion which is causing the delay in overall motr process startup. From this perspective either we need to fix this in motr or in hax. I would prefer hax. So yes we need to release the motr thread.
Let me try something. But let's talk when you have time.

mssawant: I was thinking this is how it works,
```
The python interpreter creates a new process and spawns the threads
When thread-1 starts running, it will first acquire the GIL and lock it.
If thread-2 wants to execute now, it will have to wait for the GIL to be released even if another processor is free.
Now, suppose thread-1 is waiting for an I/O operation. At this time, it will release the GIL, and thread-2 will acquire it.
After completing the I/O ops, if thread-1 wants to execute now, it will again have to wait for the GIL to be released by thread-2.
```
is it that motr thread will not release the GIL because its an external thread?

mssawant: By the way from the logs seem things are moving in hax while the entrypoint request thread is blocked,
```
Dec 03 20:43:14 ssc-vm-c-0552.colo.seagate.com hare-hax[14000]: 2020-12-03 20:43:14,005 [DEBUG] {Dummy-1} Blocking until HaLinkMessagePromise{MessageId(0x7f202401e710, 9), MDec 03 20:43:14 ssc-vm-c-0552.colo.seagate.com hare-hax[14000]: 2020-12-03 20:43:14,016 [DEBUG] {qconsumer} The repeatable call succeeded: update_process_status
Dec 03 20:43:14 ssc-vm-c-0552.colo.seagate.com hare-hax[14000]: 2020-12-03 20:43:14,016 [DEBUG] {qconsumer} Waiting for the next message
Dec 03 20:43:17 ssc-vm-c-0552.colo.seagate.com hare-hax[14000]: 2020-12-03 20:43:17,142 [DEBUG] {fs-stats-updater} Attempting to invoke the repeatable call: _ioservices_runnDec 03 20:43:17 ssc-vm-c-0552.colo.seagate.com hare-hax[14000]: 2020-12-03 20:43:17,147 [DEBUG] {fs-stats-updater} Services "confd" received: [{'ID': 'f6dc1035-d338-339c-83fDec 03 20:43:17 ssc-vm-c-0552.colo.seagate.com hare-hax[14000]: 2020-12-03 20:43:17,149 [DEBUG] {fs-stats-updater} Raw value from KV: {'LockIndex': 0, 'Key': 'processes/0x72Dec 03 20:43:17 ssc-vm-c-0552.colo.seagate.com hare-hax[14000]: 2020-12-03 20:43:17,151 [DEBUG] {fs-stats-updater} Raw value from KV: {'LockIndex': 0, 'Key': 'processes/0x72Dec 03 20:43:17 ssc-vm-c-0552.colo.seagate.com hare-hax[14000]: 2020-12-03 20:43:17,152 [DEBUG] {fs-stats-updater} Services "ios" received: [{'ID': 'f6dc1035-d338-339c-83f9-Dec 03 20:43:17 ssc-vm-c-0552.colo.seagate.com hare-hax[14000]: 2020-12-03 20:43:17,154 [DEBUG] {fs-stats-updater} Raw value from KV: None
Dec 03 20:43:17 ssc-vm-c-0552.colo.seagate.com hare-hax[14000]: 2020-12-03 20:43:17,155 [DEBUG] {fs-stats-updater} Raw value from KV: None
```
No?

knekrasov: You won 👍 

It seems like I was too pessimistic on external non-python threads. Please feel free to ignore the request in this conversation.

hax/hax/handler.py
mssawant: Agree.

hax/hax/motr/__init__.py
mssawant: Added.

hax/hax/motr/hax.c
mssawant: Done.

hax/hax/handler.py
mssawant: Removed.

hax/hax/handler.py
mssawant: Removed.

hax/hax/motr/__init__.py
mssawant: Fixed.

hax/hax/motr/delivery.py
mssawant: Removed for now.

hax/hax/motr/delivery.py
knekrasov: @mssawant I took a liberty to correct this code a bit. Please have a look at https://github.com/mssawant/cortx-hare-1/commit/34f2594b5e37e8cbf1a1143dfc377aca161181a5

I think without that dicts `self.recently_delivered` and `self.waiting_clients` will grow till infinity.

andriytk:
That's why I didn't want to maintain the processes ha-status in two places - in Consul's catalogue and in the KV Store...

Could you please somebody remind me - why are we doing this?

knekrasov:
@mssawant I wrote some unit tests for `wait_for_all()` functionality and uncovered some algorithmic problems. 

Unit tests and fixes to DeliveryHerald are here: https://github.com/mssawant/cortx-hare-1/commit/24141cdd89e768e60b399bb5c315e7b4372e5655

hax/hax/motr/delivery.py
mssawant: Thanks, I see one more improvement while testing. Will update the patch.

mssawant:
@andriytk, Consul catalog statuses are based on systemd. So even if a systemd unit is started may not mean that the process is actually started. Now we have moved sd_notify() to a later position so the motr state and systemd state of the process may snchronise. But that delays the process startup and I doubt if we actually need it.

mssawant:
Tested on sm6/7-r23 setup with custom build http://cortx-storage.colo.seagate.com/releases/cortx/github/integration-custom-ci/release/centos-7.8.2003/custom-build-231/, working fine. Failover/failback, Motr process kill-restarts working fine.
Motr process startup time is now reduced to 8secs,
```
Dec 07 23:33:07 sm6-r23.pun.seagate.com systemd[1]: Starting Cluster Controlled m0d@0x7200000000000001:0xc...
Dec 07 23:33:07 sm6-r23.pun.seagate.com motr-server[90736]: /sys/kernel/mm/transparent_hugepage/defrag: always madvise [never]
Dec 07 23:33:07 sm6-r23.pun.seagate.com motr-server[90736]: MOTR_M0D_EP: 192.168.0.3@o2ib:12345:2:2
Dec 07 23:33:07 sm6-r23.pun.seagate.com motr-server[90736]: MOTR_PROCESS_FID: 0x7200000000000001:0xc
Dec 07 23:33:07 sm6-r23.pun.seagate.com motr-server[90736]: MOTR_HA_EP: 192.168.0.3@o2ib:12345:1:1
Dec 07 23:33:07 sm6-r23.pun.seagate.com motr-server[90736]: vm.max_map_count = 30000000
Dec 07 23:33:07 sm6-r23.pun.seagate.com motr-server[90736]: Service FID: m0d-0x7200000000000001:0xc
Dec 07 23:33:07 sm6-r23.pun.seagate.com motr-server[90736]: BE log size is not configured
Dec 07 23:33:07 sm6-r23.pun.seagate.com motr-server[90736]: + exec /usr/bin/m0d -e lnet:192.168.0.3@o2ib:12345:2:2 -A linuxstob:/var/log/seagate/motr/addb/m0d-0x7200000000000001:0xc/addb-stobDec 07 23:33:15 sm6-r23.pun.seagate.com motr-server[90736]: Started
Dec 07 23:33:15 sm6-r23.pun.seagate.com systemd[1]: Started Cluster Controlled m0d@0x7200000000000001:0xc.
Dec 07 23:33:15 sm6-r23.pun.seagate.com motr-server[90736]: m0d: systemd READY notification successful
Dec 07 23:33:15 sm6-r23.pun.seagate.com motr-server[90736]: Press CTRL+C to quit.
```

knekrasov (approved):
Note: some of too verbose logging will be improved in scope of #1421

andriytk:
Discussed on the DSU meeting with @mssawant , sounds Ok.

andriytk (approved):
Discussed the overall approach on DSU meeting with @mssawant - sounds Ok.

PR #1426, opened 2020-12-08
hax/setup.py: fix make devinstall

### LINKED ISSUE #1425: make devinstall does not work
Following https://github.com/Seagate/cortx-hare#installation:

```
[vagrant@cmu hare]$ sudo make devinstall
...
--> Installing hax with 'source /data/hare/.py3venv/bin/activate; python3.6 setup.py develop --prefix /opt/seagate/cortx/hare'
  File "setup.py", line 141, in <module>
    library_dirs=[get_motr_libs_dir()],
  File "setup.py", line 97, in get_motr_libs_dir
    assert P.isfile(libmotr), f'{libmotr}: No such file'
AssertionError: /data/cortx-motr/motr/.libs/libmotr.so: No such file
make: *** [/opt/seagate/cortx/hare/lib/python3.6/site-packages/hax.egg-link] Error 1
```

Commit 571ae39

andriytk:
Currently, if one follows https://github.com/Seagate/cortx-hare#installation
document, the installation fails on make devinstall:

```
[vagrant@cmu hare]$ sudo make devinstall
...
--> Installing hax with 'source /data/hare/.py3venv/bin/activate; python3.6 setup.py develop --prefix /opt/seagate/cortx/hare'
  File "setup.py", line 141, in <module>
    library_dirs=[get_motr_libs_dir()],
  File "setup.py", line 97, in get_motr_libs_dir
    assert P.isfile(libmotr), f'{libmotr}: No such file'
AssertionError: /data/cortx-motr/motr/.libs/libmotr.so: No such file
make: *** [/opt/seagate/cortx/hare/lib/python3.6/site-packages/hax.egg-link] Error 1
```

The problem is that it's assumed that Motr sources are located
at `../cortx-motr` path while the clone command in the document
checks out Motr at `../motr` path.

Solution: fix setup.py to align with the documentation and assume
the Motr sources are installed at `../motr`.

Closes #1425.

Signed-off-by: Andriy Tkachuk <andriy.tkachuk@seagate.com>

### DIFF
--- a/hax/setup.py
+++ b/hax/setup.py
@@ -79,7 +79,7 @@ def get_motr_dir():
     d = os.environ.get('M0_SRC_DIR')
     if d:
         return d
-    return P.normpath(P.dirname(P.abspath(__file__)) + '/../../cortx-motr')
+    return P.normpath(P.dirname(P.abspath(__file__)) + '/../../motr')
 
 
 def get_motr_libs_dir():


knekrasov:
Why wouldn't we update hare readme file instead?

I would expect that the stack will work with the default folder names created by git. It is a bit unusual when the user is required to clone the repo to a non-default folder.

andriytk:
> Why wouldn't we update hare readme file instead?

I would rather do that indeed if the name was something short and neat. Instead, we have `cortx-hare` which is rather a bit long and awkward.

chumakd:
> It is a bit unusual when the user is required to clone the repo to a non-default folder.
I second @knekrasov's point.

> that's what our README doc mentions
docs aren't carved in a stone these days, are they?

chumakd (commented):
I'd rather have the README updated instead of the code.

andriytk:
I've made a short survey among motr / hare developers and, as appeared, looks like most people use motr / hare directories instead of cortx-motr / cortx-hare (or just motr / hare symlinks to their cortx-prefixed counterparts).

knekrasov:
Ideally we need to rename repos then.

As @chumakd pointed out in that survey, many GUI tools that less experienced programmers can use will simply not offer an option to select "target" directory when checking out. Long story short, in normal case a developer will checkout to a default folder. It is just the matter of "the least astonishment" principle.

 By default the folder name is the same to the repo name. So the folder name will be `cortx-hare` in case of `https://github.com/Seagate/cortx-hare.git` repository.

I don't think that the user experience must suffer just because the repo names are not perfect. In other words, the simpler (and shorter) the Quick Start instruction is, the more people will want to dig deeper and become familiar with cortx products.

PR #1431, opened 2020-12-15
Problem: Instructions to build hare rpm from source are missing

mssawant:
Solution:
Update hare README.md to include instructions to build hare rpm from source.


johnbent (approved):
Looks really good!

knekrasov (approved):

PR #1437, opened 2020-12-23
Problem: every hax instance updates FS statistics

### LINKED ISSUE #1406: Problem: hax runs fs-stats updater on all cluster nodes
There are 16 client nodes on the SAGE cluster and 6 server nodes - and on all of them `hax` runs fs-stats updater:

```
17:04 tkachuk1@client-22:~$ sudo journalctl -lf | grep 'fs-stats-updater.*rconfc'
Nov 24 17:04:37 client-22 hare-hax[18380]: 2020-11-24 17:04:37,722 [DEBUG] {fs-stats-updater} Starting rconfc
Nov 24 17:04:38 client-22 hare-hax[18380]: 2020-11-24 17:04:38,692 [DEBUG] {fs-stats-updater} rconfc started
Nov 24 17:04:40 client-22 hare-hax[18380]: 2020-11-24 17:04:40,096 [DEBUG] {fs-stats-updater} Stopping rconfc
Nov 24 17:05:10 client-22 hare-hax[18380]: 2020-11-24 17:05:10,192 [DEBUG] {fs-stats-updater} Starting rconfc
Nov 24 17:05:11 client-22 hare-hax[18380]: 2020-11-24 17:05:11,125 [DEBUG] {fs-stats-updater} rconfc started
Nov 24 17:05:12 client-22 hare-hax[18380]: 2020-11-24 17:05:12,569 [DEBUG] {fs-stats-updater} Stopping rconfc
...
```

It does not only spam the system log, but also generates a redundant load on Consul (by updating the stats from each node in KV Store) and Motr.

Suggestion: modify `hax` code so that it runs fs-stats updater only on the RC-leader server node. On each iteration, `hax` can check whether it's running on RC-leader node by looking at the `leader` key in the KV Store and compare its value with `node-name`.

knekrasov:
That's an overkill, leads to extra load of Consul KV (given we store the
statistics there) and also causes extra fops to happen (Spiel API is
used to fetch that information).

Solution: fetch FS statistics from RC node only.

Closes #1406

### DIFF
--- a/hax/hax/filestats.py
+++ b/hax/hax/filestats.py
@@ -57,6 +57,10 @@ class FsStatsUpdater(StoppableThread):
             ffi.adopt_motr_thread()
             self._ensure_motr_all_started()
             while not self.stopped:
+                if not self._am_i_rc():
+                    self._sleep(self.interval_sec)
+                    continue
+
                 started = self._ioservices_running()
                 if not all(started):
                     self._sleep(self.interval_sec)
@@ -107,3 +111,10 @@ class FsStatsUpdater(StoppableThread):
                 LOG.debug('According to Consul all confds have been started')
                 return
             self._sleep(5)
+
+    def _am_i_rc(self):
+        # The call is already marked with @repeat_if_fails
+        leader = self.consul.get_leader_node()
+        # The call doesn't communicate via Consul REST API
+        this_node = self.consul.get_local_nodename()
+        return leader == this_node
--- a/hax/hax/util.py
+++ b/hax/hax/util.py
@@ -228,6 +228,11 @@ class ConsulUtil:
             f'No {svc_name!r} Consul service found at node {hostname!r}')
 
     def get_local_nodename(self) -> str:
+        """
+        Returns the logical name of the current node. This is the name that
+        Consul is aware of. In other words, whenever Consul references a node,
+        it will use the names that this function can return.
+        """
         try:
             local_nodename = os.environ.get('HARE_HAX_NODE_NAME') or \
                 self.cns.agent.self()['Config']['NodeName']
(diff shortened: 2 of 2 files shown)


hax/hax/filestats.py
```diff
@@ -57,6 +57,10 @@ def _execute(self, motr: Motr):
             ffi.adopt_motr_thread()
             self._ensure_motr_all_started()
             while not self.stopped:
+                if not self._am_i_rc():
```
andriytk, line 60: [Style] suggestion: change "_am_i_rc()" to shorter and more readable variant: "_is_rc()".

hax/hax/util.py
```diff
@@ -228,6 +228,11 @@ def _service_by_name(self, hostname: str, svc_name: str) -> Dict[str, Any]:
             f'No {svc_name!r} Consul service found at node {hostname!r}')
 
     def get_local_nodename(self) -> str:
+        """
+        Returns the logical name of the current node. This is the name that
+        Consul is aware of. In other words, whenever Consul references a node,
+        it will use the names that this function can return.
+        """
```
andriytk, line 235: In normal case - yes. But in the failover case - not quite true. In the failover case both hax connected to the same consul agent listening on the default localhost port - that's why the environment variable was introduced (because we could not always determine the nodename from consul).

chumakd (approved):

hax/hax/filestats.py
knekrasov: Not sure it is self descriptive "Is RC?" - "Who are you talking about?"

hax/hax/util.py
knekrasov: Maybe you can suggest a better text?

andriytk: I'm not sure any text is needed here. The code seems quite self-explanatory.

knekrasov: Well, historically it was you @andriytk who explained me why this function should be used in this context (so node name is returned and that's the name that Consul is aware of). If I don't want other developers to repeat the same communication route, I think that text is needed.

And yes, a function without a docstring makes our Codacy unhappy 😄

hax/hax/filestats.py
knekrasov: @chumakd What's your opinion? I would love to get this PR merged, so I'd like to resolve all the discussions :-)

chumakd:
looks good enough, merged

PR #1441, opened 2020-12-24
EOS-15919 Make distclean does not delete all the untracked files.

gshipra:
EOS-15919 Make distclean doesn't delete all files.

Solution: Added code in Makefile to delete leftover files, which
are created at the time of make and which should be deleted at
make distclean.
Test steps:
1. executed - make
2. executed - make distclean
Relevant output :
--> Cleaning hax
    removing hax/build
    removing hax/dist
    removing hax/hax.egg-info
    removing hax/libhax.cpython-36m-x86_64-linux-gnu.so
    removing hax/hax/__pycache__
    removing hax/hax/motr/__pycache__ dir
    removing hax/hax/queue/__pycache__ dir
    removing hax/test/__pycache__ dir
3. executed - git clean -n -fdx
It returned nothing.

Signed-off-by: Shipra Gupta <shipra.gupta@seagate.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,7 +114,8 @@ clean: clean-hax clean-mypy clean-dhall-prelude
 clean-hax:
 	@$(call _info,Cleaning hax)
 	@for d in hax/build hax/dist hax/hax.egg-info hax/*.so \
-	          hax/hax/__pycache__; do \
+		  hax/hax/__pycache__ hax/hax/motr/__pycache__ \
+		  hax/hax/queue/__pycache__ hax/test/__pycache__; do \
 	     if [[ -e $$d ]]; then \
 	         $(call _log,removing $$d); \
 	         rm -rf $$d; \


Makefile
```diff
@@ -120,6 +120,18 @@ clean-hax:
 	         rm -rf $$d; \
 	     fi; \
 	 done
+	@if [[ -d hax/hax/motr/__pycache__ ]]; then \
```
knekrasov, line 123: Can't we include this directory into `for` loop above? I don't think it is worth copying and pasting the same pieces of code.

knekrasov (changes requested):

chumakd (changes requested):
I agree with @knekrasov, it's better to add the file paths into the existing `for` loop which already doing `rm -rf` for other directories. There's no need to add a separate explicit `rm` command for every `__pycache__` dir.

Makefile
gshipra: Done the changes suggested.

knekrasov:
@gshipra One last thing. Please reword the commit message.

Commit header and commit description must be separated with an empty line. Here is how your commit message looks like:
image
(note that even vim highlights the line number 2 with red background - it is not supposed to be there).

More info on git commit messages can be found here: https://haydar-ai.medium.com/learning-how-to-git-creating-a-longer-commit-message-16ca32746c3a

Thanks!

gshipra:
> @gshipra One last thing. Please reword the commit message.
> 
> Commit header and commit description must be separated with an empty line. Here is how your commit message looks like:
> image
> (note that even vim highlights the line number 2 with red background - it is not supposed to be there).
> 
> More info on git commit messages can be found here: https://haydar-ai.medium.com/learning-how-to-git-creating-a-longer-commit-message-16ca32746c3a
> 
> Thanks!

Done. Changed the commit message to less than 50 chars and added a line between header and body.

knekrasov (approved):

chumakd (approved):

Issue #1438, opened 2020-12-23
Make Spiel object global


knekrasov:
Currently, fs-stats-updater thread starts and stops rconfc by itself. Given in R2 we'll also need to support SNS operations, Spiel context must be handled differently.

Please note that SNS-related operations are processed in ConsumerThread. Effectively SNS operation goes down to this function from `hax.c`:
{code:c}
static int spiel_sns_op(enum m0_cm_op sns_copy_machine_op,
			unsigned long long ctx,
			const struct m0_fid *pool_fid,
			struct m0_spiel_repreb_status **sns_statuses)
{
	struct hax_context *hc = (struct hax_context *)ctx;
	struct m0_spiel *spiel = m0_halon_interface_spiel(hc->hc_hi);
	int rc = 0;

	M0_ENTRY();

	if (!hc->hc_alive) {
		M0_LOG(M0_WARN, "Cannot make SPIEL API call as HAX Python"
				" context has been already destructed");
		return M0_ERR(-ESHUTDOWN);
	}

	if (!hc->hc_rconfc_initialized) {
		M0_LOG(M0_WARN, "Cannot make SPIEL API call as libmotr 'rconfc'"
				" susbsystem hasn't been initialised");
		return M0_ERR(-EDESTADDRREQ);
	}
{code}

It requires rconfc to be already initialized. But according to the current implementation, rconfc is initialized and destructed continually in fs-stats-updater thread. That doesn't look sane.

cortx-admin:
> Vaibhav Prakash Paratwar commented in Jira Server:

[~720599] - is this still relevant? if yes, can we take it in upcoming sprints? cc [~520513]

cortx-admin:
> Justin Woo commented in Jira Server:

[~534380] was this scheduled in the backlog? If so when was it scheduled to be worked on?

cortx-admin:
> Vaibhav Prakash Paratwar commented in Jira Server:

Improvement / Refactor (Design) cc [~520513]

cortx-admin:
> Vaibhav Prakash Paratwar commented in Jira Server:

Moved this to Story JIRA instead of Bug JIRA. cc [~520513] [~520478]

shailesh-vaidya:
Closing as an obsolete

PR #1442, opened 2020-12-29
Problem: hax starts rconfc in context of fileStatsUpdater

mssawant:
Presently only one spiel command is being invoked as part of fileStatsUpdater thread
but spiel use cases increase, e.g. sns repair/rebalance, rconfc must be started
only once.

Solution:
Create a motr runnable that is started as part of motr start operation.
Start rconfc as part of motr thread but wait for motr services to start.

### DIFF
--- a/hax/hax/filestats.py
+++ b/hax/hax/filestats.py
@@ -19,12 +19,11 @@
 import datetime
 import logging
 from threading import Event
-from typing import List
 
 from hax.exception import HAConsistencyException, InterruptedException
 from hax.motr import Motr, log_exception
 from hax.types import FsStatsWithTime, StoppableThread
-from hax.util import ConsulUtil, repeat_if_fails
+from hax.util import ConsulUtil, wait_for_event
 
 LOG = logging.getLogger('hax')
 
@@ -44,46 +43,35 @@ class FsStatsUpdater(StoppableThread):
         self.stopped = True
         self.event.set()
 
-    def _sleep(self, interval_sec) -> None:
-        interrupted = self.event.wait(timeout=interval_sec)
-        if interrupted:
-            raise InterruptedException()
-
     @log_exception
     def _execute(self, motr: Motr):
         try:
             ffi = motr._ffi
             LOG.info('filesystem stats updater thread has started')
             ffi.adopt_motr_thread()
-            self._ensure_motr_all_started()
             while not self.stopped:
                 if not self._am_i_rc():
-                    self._sleep(self.interval_sec)
+                    wait_for_event(self.event, self.interval_sec)
                     continue
-
-                started = self._ioservices_running()
-                if not all(started):
-                    self._sleep(self.interval_sec)
+                if not motr.is_spiel_ready():
+                    wait_for_event(self.event, self.interval_sec)
                     continue
-                result: int = motr.start_rconfc()
-                if result == 0:
-                    stats = motr.get_filesystem_stats()
-                    motr.stop_rconfc()
-                    if not stats:
-                        continue
-                    LOG.debug('FS stats are as follows: %s', stats)
-                    now_time = datetime.datetime.now()
-                    data = FsStatsWithTime(stats=stats,
-                                           timestamp=now_time.timestamp(),
-                                           date=now_time.isoformat())
-                    try:
-                        self.consul.update_fs_stats(data)
-                    except HAConsistencyException:
-                        LOG.debug('Failed to update Consul KV '
-                                  'due to an intermittent error. The '
-                                  'error is swallowed since new attempts '
-                                  'will be made timely')
-                self._sleep(self.interval_sec)
+                stats = motr.get_filesystem_stats()
+                if not stats:
+                    continue
+                LOG.debug('FS stats are as follows: %s', stats)
+                now_time = datetime.datetime.now()
+                data = FsStatsWithTime(stats=stats,
+                                       timestamp=now_time.timestamp(),
+                                       date=now_time.isoformat())
+                try:
+                    self.consul.update_fs_stats(data)
+                except HAConsistencyException:
+                    LOG.debug('Failed to update Consul KV '
+                              'due to an intermittent error. The '
+                              'error is swallowed since new attempts '
+                              'will be made timely')
+                wait_for_event(self.event, self.interval_sec)
         except InterruptedException:
             # No op. _sleep() has interrupted before the timeout exceeded:
             # the application is shutting down.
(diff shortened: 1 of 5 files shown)


knekrasov:
Related issue: #1438

hax/hax/hax.py
```diff
@@ -100,18 +100,18 @@ def main():
     consumer = _run_qconsumer_thread(q, motr, herald, util)
 
     try:
-        motr.start(cfg.hax_ep,
-                   process=cfg.hax_fid,
-                   ha_service=cfg.ha_fid,
-                   rm_service=cfg.rm_fid)
+        motr_runnable = motr.start(cfg.hax_ep,
```
knekrasov, line 103: Up to this moment we had only one thread that "owns" everything related to Motr - ConsumerThread. If ConsumerThread is running, hax can communicate to Motr.

#1438 states that this picture is not balanced though: rconfc stuff happens outside of ConsumerThread but implicitly depends on ConsumerThread (because that guy owns Motr context which is a must for rconfc magic to happen).

So per my understanding it would be more logical to keep all Motr-related stuff in a single place (i.e. rconfc and motr context together). But you added one more thread in addition to ConsumerThread and StatsUpdater 🤔 

Do you think your approach with MotrRunnable is better?

mssawant: Because rconfc_start is a blocking operation (requires motr processes to be started, atleast confd) we need a non-blocking context, hence the thread. Effectively the thread doesn't do anything but just starts rconfc and sleeps until hax process terminates.
You do have a point that this active thread can give an impression of multiple motr active processing contexts, I did think of adding a separate message type to start rconfc once all the motr processes have started but rconfc start can still block.
May be the name motrrunnable is a bit misnomer and can be renamed to may be just rconfcManager or something.

knekrasov: @mssawant 
> Because rconfc_start is a blocking operation (requires motr processes to be started, atleast confd) we need a non-blocking context, hence the thread. 

I don't think it is a problem. I think it is OK if a thread blocks until rconfc_start starts. In other words, if a thread needs to work with Motr (or with Spiel API), it needs to have the corresponding context initialized. So it must wait until it is initialized just because it can't do anything meaningful without the context anyway. 

So it seems like this splitting to a number of threads will make the things more complicated (inter-thread communication, like "a separate message type to start rconfc once all the motr processes have started" in your example) without a real benefit.

knekrasov: I'm just trying to avoid adding extra [accidental complexity](https://medium.com/background-thread/accidental-and-essential-complexity-programming-word-of-the-day-b4db4d2600d4).

mssawant: @knekrasov, If not in a separate thread, do you suggest to start rconfc in consumer thread and block consumer thread then until rconfc starts? I see a deadlock there as hax will not be able to reply to entrypoint and nvec requests which will block motr confd and ioservices.

knekrasov: @mssawant 
Hm, good point about the deadlock.

What if we use a compromise solution like this:
image

In short, 
1. rconfc is initialized in a separate thread to `qconsumer`, that thread is called `rconfc_stater` at the diagram
2. `rconfc_starter` is short-lived, it exits as soon as rconfc is started. Its aim is effectively to start rconfc asynchronously to `qconsumer`
3. overall Motr context is still owned by one thread - `qconsumer`.

mssawant: Agree, have updated the patch with necessary changes. Thank you.

hax/hax/filestats.py
```diff
@@ -19,12 +19,12 @@
 import datetime
 import logging
 from threading import Event
-from typing import List
+# from typing import List
```
knekrasov, line 22: Perhaps you forgot to remove it.

hax/hax/motr/__init__.py
```diff
@@ -297,6 +306,8 @@ def _generate_sub_services(self, note: HaNoteStruct,
         ]
 
     def close(self):
+        LOG.debug('Stopping rconfc')
+        self.stop_rconfc()
```
knekrasov, line 310: Don't we need to check whether rconfc is currently running?

hax/hax/motr/__init__.py
```diff
@@ -405,3 +416,43 @@ def resume_rebalance(self, pool_fid: Fid):
                 'Failed to send SPIEL request "sns_rebalance_resume",' +
                 'please check Motr logs for more details.')
         LOG.debug('Rebalancing resumed for pool %s', pool_fid)
+
+
+class RconfcStarter(StoppableThread):
+    def __init__(self, motr: Motr, consul_util: ConsulUtil):
+        super().__init__(target=self._execute,
+                         name='rconfc-manager',
```
knekrasov, line 424: Suggestion: `s/rconfc-manager/rconfc-starter/g`

hax/hax/util.py
```diff
@@ -137,6 +138,12 @@ def wrapper(*args, **kwds):
                                  ('cas', Optional[Any])])
 
 
+def wait_for_event(event: Event, interval_sec) -> None:
```
knekrasov, line 141: I would at least added a docstring describing that this method is actually used to perform "sleep" operation. It is not that clear from its name or signature.

hax/hax/motr/__init__.py
```diff
@@ -81,6 +84,9 @@ def start(self, rpc_endpoint: str, process: Fid, ha_service: Fid,
                 ' returned non-zero code (%s)', result)
             raise RuntimeError('Cannot start m0_halon_interface.'
                                'Please check Motr logs for more details.')
+        thread = RconfcStarter(self, self.consul_util)
+        thread.start()
+        return thread
```
knekrasov, line 89: Please update the method signature to reflect the return type.

knekrasov (changes requested):
In general looks good, there are just a few quite minor things to polish.

hax/hax/filestats.py
mssawant: Removed.

hax/hax/motr/__init__.py
mssawant: don't think so, `hax/hax/motr/hax.c::stop_rconfc()` handles it,
```
int stop_rconfc(unsigned long long ctx)
{
        struct hax_context *hc = (struct hax_context *)ctx;

        if (hc->hc_rconfc_initialized) {
                struct m0_spiel *spiel = m0_halon_interface_spiel(hc->hc_hi);
                m0_spiel_rconfc_stop(spiel);
        }

        return 0;
}
```

hax/hax/motr/__init__.py
mssawant: Done.

hax/hax/util.py
mssawant: Added.

hax/hax/motr/__init__.py
mssawant: Done.

knekrasov:
@mssawant Looks good in general. I took a liberty to add a fixup commit to your branch to shorten the code review frictions.

If you're OK with my changes, please squash them into your branch and I think we can merge.
Thank you!

knekrasov (approved):

SwapnilGaonkar7 (approved):
Looks good

PR #1448, opened 2021-01-05
EOS-16201:Hare Run I/O test m0crate: command not found

suvratjoshi:
Problem:
m0crate command throwing error for the steps given in README.md
Error : m0crate: command not found
Solution:
m0crate can be located in motr directory, so the m0crate utility
needs correct path while executing this command.
This only needs a documentation fix, which is done here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,8 +187,13 @@ You will probably need to modify `host`, `data_iface`, and `io_disks` values.
   ```sh
   /opt/seagate/cortx/hare/libexec/m0crate-io-conf >/tmp/m0crate-io.yaml
   dd if=/dev/urandom of=/tmp/128M bs=1M count=128
-  sudo ./motr/motr/m0crate/m0crate -S /tmp/m0crate-io.yaml
+  sudo m0crate -S /tmp/m0crate-io.yaml
   ```
+  Please note that m0crate will run as shown above when it will be
+  available in default system PATH which will be the case when when
+  setup is created using RPM's. If its created by building Motr
+  source code, then m0crate utility can be run using full path from
+  the motr source directory (./motr/motr/m0crate/m0crate).
 
 * Stop the cluster.
   ```sh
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -187,7 +187,7 @@ You will probably need to modify `host`, `data_iface`, and `io_disks` values.
   ```sh
   /opt/seagate/cortx/hare/libexec/m0crate-io-conf >/tmp/m0crate-io.yaml
   dd if=/dev/urandom of=/tmp/128M bs=1M count=128
-  sudo m0crate -S /tmp/m0crate-io.yaml
+  sudo ./motr/motr/m0crate/m0crate -S /tmp/m0crate-io.yaml
```
chumakd, line 190: This will work only if Motr is built from sources, which is not the case when people want to install binary Motr rpm package instead of building Motr themselves.

I suggest to add a comment instead of change the example script. Where you can describe that in case of Motr built from sources `m0crate` is not available in the default system `PATH` and needs to be called explicitly using full path in Motr source dir.

chumakd (changes requested):

chumakd (approved):

README.md
```diff
@@ -189,6 +189,11 @@ You will probably need to modify `host`, `data_iface`, and `io_disks` values.
   dd if=/dev/urandom of=/tmp/128M bs=1M count=128
   sudo m0crate -S /tmp/m0crate-io.yaml
   ```
+  Please note that m0crate will run as shown above when it will be
+  available in default system PATH which will be the case when when
```
mssawant, line 193: s/when when/when

README.md
```diff
@@ -189,6 +189,11 @@ You will probably need to modify `host`, `data_iface`, and `io_disks` values.
...
+  setup is created using RPM's. If its created by building Motr
```
mssawant, line 194: s/RPM's/RPMs

README.md
```diff
@@ -189,6 +189,11 @@ You will probably need to modify `host`, `data_iface`, and `io_disks` values.
...
+  source code, then m0crate utility can be run using full path from
```
mssawant, line 195: [suggestion] s/full path/absolute path

PR #1459, opened 2021-01-15
Problem: Mini-Provisioner integration is not documented

knekrasov:
Solution: add 20/MINIPROV RFC

### DIFF
--- a/rfc/20/README.md
+++ b/rfc/20/README.md
@@ -1,0 +1,165 @@
+---
+domain: github.com
+shortname: 20/MINIPROV
+name: Mini-Provisioning API Support
+status: draft
+editor: Konstantin Nekrasov <konstantin.nekrasov@seagate.com>
+---
+
+<!-- vim-markdown-toc GFM -->
+
+* [Language](#language)
+* [Abstract](#abstract)
+* [Vocabulary](#vocabulary)
+* [General setup.yaml structure](#general-setupyaml-structure)
+  * [Notes](#notes)
+  * [Stage conditions](#stage-conditions)
+* [Communication with ConfStore](#communication-with-confstore)
+* [Proposed changes to Hare](#proposed-changes-to-hare)
+  * [Stage implementations](#stage-implementations)
+    * [post_install](#post_install)
+    * [config](#config)
+      * [CDF generation](#cdf-generation)
+    * [init](#init)
+    * [test](#test)
+
+<!-- vim-markdown-toc -->
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Abstract
+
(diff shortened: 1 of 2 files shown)


rfc/20/README.md
```diff
@@ -0,0 +1,84 @@
...
+        args: --config $URL 
+    support_bundle:   
+      - /opt/seagate/cortx/provisioner/cli/provisioner-bundler   
+```
+
+### Stage conditions
+
+| Stage        | Purpose           | Expected initial state  | Side effects performed as a result |
+| ------------- |-------------| ----- | --- |
+| post_install  | Verify that all the required 3rd party components are installed and have correct versions (Optional?) | Component RPM has just been installed to the OS. No services are started. | None (stage either fails or succeeds) |
```
mssawant, line 64: I think we need not re-iterate what is present in the `mini-provisioner` document but specifically mention what Hare will be doing in each section, e.g. config - generate CDF using information from conf-store.

rfc/20/README.md
```diff
@@ -0,0 +1,148 @@
...
+editor: Konstantin Nekrasov <konstantin.nekrasov@seagate.com>
+---
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Abstract
+
+Hare component must implement Mini-Provisioner API (see the [specification document](https://seagatetechnology.sharepoint.com/:w:/r/sites/gteamdrv1/tdrive1224/_layouts/15/guestaccess.aspx?email=nitin.nimran%40seagate.com&e=4%3ADCGSG0&at=9&CID=C40F4AE6-CD62-43A6-8B09-A76C04A292F1&wdLOR=cA34642E5-B304-413A-BED3-2FD178DACFDF&share=Ea2OBEFIPmNBucxpoE_n5nsB126Vr3sA_-KY-4_hHYkNVA) for more details). In a nutshell Hare must have scripts that setup a singlenode cluster using [ConfStore](https://seagatetechnology.sharepoint.com/:w:/r/sites/gteamdrv1/tdrive1224/_layouts/15/Doc.aspx?sourcedoc=%7BA68B2DB0-B041-4136-8F51-94F0A4685D1E%7D&file=ConfStore_Specification.docx&action=default&mobileredirect=true) as a source of configuration parameters and publish those scripts via a `setup.yaml` file of a predefined format.
```
suvratjoshi, line 15: specification document : This link took me to the page to enter credentials for nitin.nimran@seagate.com

image

rfc/20/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+
+## Communication with ConfStore
+
+TBD
+
+## Proposed changes to Hare
+
+1. All commands will be handled by `provisioning/setup.py` file.
+2. `provisioning/setup.py` must berenamed to `provisioning/setup_hare` (the file must be executable so proper shebang must be added to the first line).
```
suvratjoshi, line 93: must be renamed

rfc/20/README.md
```diff
@@ -0,0 +1,148 @@
...
+3. `setup_hare` must provide help screen.
+
+### Stage implementations
+
+#### post_install
+```
+setup_hare post_install --config <str>
+```
+Verifies that
+1. Consul is installed
```
suvratjoshi, line 103: 1. If I heard it correctly, consul will not be part of hare component. If it is correct, then this step may not be required, but I am not sure as hare has dependency on consul.
2. Is it required to verify these installations as well?
yum -y install python3 python3-devel
yum -y install yum-utils
dhall-bin
dhall-prelude

rfc/20/README.md
knekrasov: Hm, I don't think I can fix it by myself.

@nitin-seagate  can you please give permissions to CORTX.Hare team to your document? It seems like not all team members have the access..

rfc/20/README.md
knekrasov: Yes, that will be covered in later sections of this document.

rfc/20/README.md
knekrasov: Good question. I don't think I know the answer.

On one hand, our hare.spec file specifies those packages as dependencies already (except for yum-utils and python3-devel - those are required for developer environment only, dhall is installed and packages in our Makefile and thus doesn't need RPM):

```
Requires: consul = 1.7.8
Requires: facter
Requires: jq
Requires: cortx-motr = %{h_motr_version}
Requires: python36
```

So it doesn't make much sense to check that those RPMs are installed. Perhaps it is worth checking Consul version of the binary (we saw situations when there were multiple Consul versions were installed at LDR R1 cluster).

@mssawant, @NeeravChoudhari thoughts?

rfc/20/README.md
knekrasov: Fixed. Thanks!

knekrasov:
Merging this PR as is to unblock other team members so they can start filling up other parts of RFC/20 in their PRs.

PR #1461, opened 2021-01-21
README: update

andriytk:
Get rid of the Welcome section in the beginning which is confusing and not longer true, I'm afraid.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,6 @@
 
 # Hare User Guide
 
-## Welcome
-
-Hare is an experiment in
-[Social Architecture](https://www.youtube.com/watch?v=uj-li0LO_2g),
-disguised as a software project.
-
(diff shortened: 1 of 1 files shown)


andriytk:
cc @chumakd, @vvv

azheregelya (approved):

knekrasov (approved):
Farewell, old good times

andriytk:
"old good times" of social experiments? :)

knekrasov:
Yes, the times when we had such an opportunity. No jokes, it's been a marvelous time and the results we got then prove that simply.

Some real benefits of those approaches are only being evaluated and understood by our dev team (including me). 

@vvv I didn't always understand your ideas and trade-offs, but now I can compare 😄 you were often right.

PR #1463, opened 2021-01-21
Problem: Hare uses older Consul version 1.7.8

mssawant:
Solution:
Update hare requirement for Consul version to 1.9.1.

### DIFF
--- a/hare.spec
+++ b/hare.spec
@@ -45,7 +45,7 @@ BuildRequires: python36-devel
 BuildRequires: python36-pip
 BuildRequires: python36-setuptools
 
-Requires: consul = 1.7.8
+Requires: consul = 1.9.1
 Requires: facter
 Requires: jq
 Requires: cortx-motr = %{h_motr_version}
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ and health-checking mechanisms.
   ```sh
   sudo yum -y install yum-utils
   sudo yum-config-manager --add-repo https://rpm.releases.hashicorp.com/RHEL/hashicorp.repo
-  sudo yum -y install consul-1.7.8
+  sudo yum -y install consul-1.9.1
   ```
 
 * Install Motr.


hare.spec
```diff
@@ -45,7 +45,7 @@ BuildRequires: python36-devel
 BuildRequires: python36-pip
 BuildRequires: python36-setuptools
 
-Requires: consul = 1.7.8
+Requires: consul = 1.9.1
```
suvratjoshi, line 48: Also change needed in :
File README.md
Line 59 : sudo yum -y install consul-1.7.8

mssawant: Ah, excellent catch! thanks.

mssawant: Done.

knekrasov (approved):

andriytk:
Few questions:
1) Why 1.9.1 if there is already 1.9.2?
2) Why not to try the more stable 1.8.x branch? Why to risk with 1.9? Is there some new feature there that we need?

mssawant:
>Few questions:

Why 1.9.1 if there is already 1.9.2?
Why not to try the more stable 1.8.x branch? Why to risk with 1.9? Is there some new feature there that we need?

No, Just tried with the latest available Consul version, I found 1.9.1, looks like its1.9.2 now, but did some basic testing with 1.9.1,  bootstrapping which worked. So went ahead with it. I did not check 1.8.x. I think we need to run more tests to see if everything works fine. Probably cQA can do that or I can try to run some more failure test cases next week.

mssawant:
@andriytk, tested bootstrapping, node failure, node join did not see any consul specific issue.

hare.spec
knekrasov: @mssawant please have a look at #1475. Perhaps that approach is even better?

andriytk (approved):
Further improvement is continued at #1475.

PR #1462, opened 2021-01-21
EOS-16512 Implement hare setup.yaml provisioning sections

SwapnilGaonkar7:
Initial implementation of hare_setup.py to implement steps/sections required by mini-provisioner
TODO: Add rpm version check in post_install stage.

UT: 
Ignore provisioner related error for now.

-bash-4.2$ sudo provisioning/hare_setup.py --post_install
'cortx-s3server' is not installed
Failed to fetch data from provisioner ([Errno 2] No such file or directory: 'provisioner': 'provisioner')
-bash-4.2$ echo $?
0
-bash-4.2$ sudo provisioning/hare_setup.py --init
Cluster is not running
2021-01-25 23:34:12: Generating cluster configuration... OK
2021-01-25 23:34:13: Starting Consul server on this node........... OK
2021-01-25 23:34:22: Importing configuration into the KV store... OK
2021-01-25 23:34:22: Starting Consul on other nodes... OK
2021-01-25 23:34:22: Updating Consul configuraton from the KV store... OK
2021-01-25 23:34:23: Waiting for the RC Leader to get elected....... OK
2021-01-25 23:34:28: Starting Motr (phase1, mkfs)... OK
2021-01-25 23:34:35: Starting Motr (phase1, m0d)... OK
2021-01-25 23:34:48: Starting Motr (phase2, mkfs)... OK
2021-01-25 23:35:11: Starting Motr (phase2, m0d)... OK
2021-01-25 23:35:28: Checking health of services... OK
Stopping m0d@0x7200000000000001:0xc (ios) at ssc-vm-c-0645.colo.seagate.com...
Stopping m0d@0x7200000000000001:0x17 (ios) at ssc-vm-c-0645.colo.seagate.com...
Stopped m0d@0x7200000000000001:0xc (ios) at ssc-vm-c-0645.colo.seagate.com
Stopped m0d@0x7200000000000001:0x17 (ios) at ssc-vm-c-0645.colo.seagate.com
Stopping m0d@0x7200000000000001:0x9 (confd) at ssc-vm-c-0645.colo.seagate.com...
Stopped m0d@0x7200000000000001:0x9 (confd) at ssc-vm-c-0645.colo.seagate.com
Stopping hare-hax at ssc-vm-c-0645.colo.seagate.com...
Stopped hare-hax at ssc-vm-c-0645.colo.seagate.com
Stopping hare-consul-agent at ssc-vm-c-0645.colo.seagate.com...
[...]

PR #1467, opened 2021-01-22
EOS-16511: Mini-Provisioner requires CDF generation

knekrasov:
Jira link: https://jts.seagate.com/browse/EOS-16511

Solution: leverage Dhall language and type system from CFGEN

### DIFF
--- a/provisioning/miniprov/hare_mp/cdf.py
+++ b/provisioning/miniprov/hare_mp/cdf.py
@@ -9,6 +9,8 @@ from hare_mp.store import ValueProvider
 from hare_mp.types import DList, Maybe, NodeDesc, Text
 
 DHALL_PATH = '/opt/seagate/cortx/hare/share/cfgen/dhall'
+DHALL_EXE = '/opt/seagate/cortx/hare/bin/dhall'
+DHALL_TO_YAML_EXE = '/opt/seagate/cortx/hare/bin/dhall-to-yaml'
 
 
 class CdfGenerator:
@@ -47,13 +49,13 @@ class CdfGenerator:
     def generate(self) -> str:
         gencdf = self._get_cdf_dhall()
 
-        dhall = S.Popen(['dhall'],
+        dhall = S.Popen([DHALL_EXE],
                         stdin=S.PIPE,
                         stdout=S.PIPE,
                         stderr=S.PIPE,
                         encoding='utf8')
 
-        to_yaml = S.Popen(['dhall-to-yaml'],
+        to_yaml = S.Popen([DHALL_TO_YAML_EXE],
                           stdin=dhall.stdout,
                           stdout=S.PIPE,
                           stderr=S.PIPE,
--- a/provisioning/miniprov/hare_mp/cdf.py
+++ b/provisioning/miniprov/hare_mp/cdf.py
@@ -1,0 +1,93 @@
+import os.path as P
+import subprocess as S
+from string import Template
+from typing import Dict, List
+
+import pkg_resources
+
+from hare_mp.store import ValueProvider
+from hare_mp.types import DList, Maybe, NodeDesc, Text
+
+DHALL_PATH = '/opt/seagate/cortx/hare/share/cfgen/dhall'
+
+
+class CdfGenerator:
+    def __init__(self, provider: ValueProvider):
+        super().__init__()
+        self.provider = provider
+
+    def _get_dhall_path(self) -> str:
+        if P.exists(DHALL_PATH):
+            return DHALL_PATH
+        raise RuntimeError('CFGEN Dhall types not found')
+
+    def _gencdf(self) -> str:
+        resource_path = '/'.join(('dhall', 'gencdf.dhall'))
+        raw_content: bytes = pkg_resources.resource_string(
+            'hare_mp', resource_path)
+        return raw_content.decode('utf-8')
+
+    def _create_node_descriptions(self) -> List[NodeDesc]:
+        nodes: List[NodeDesc] = []
+        conf = self.provider
+        server_dict: Dict[str, str] = conf.get('cluster>server_nodes')
+        for _, node in server_dict.items():
+            nodes.append(self._create_node(node))
+        return nodes
+
+    def _get_cdf_dhall(self) -> str:
+        dhall_path = self._get_dhall_path()
+        nodes = self._create_node_descriptions()
+
+        params_text = str(nodes)
+        gencdf = Template(self._gencdf()).substitute(path=dhall_path,
+                                                     params=params_text)
+        return gencdf
+
+    def generate(self) -> str:
+        gencdf = self._get_cdf_dhall()
+
+        dhall = S.Popen(['dhall'],
+                        stdin=S.PIPE,
+                        stdout=S.PIPE,
+                        stderr=S.PIPE,
+                        encoding='utf8')
+
+        to_yaml = S.Popen(['dhall-to-yaml'],
+                          stdin=dhall.stdout,
+                          stdout=S.PIPE,
+                          stderr=S.PIPE,
+                          encoding='utf8')
+
+        _, err_d = dhall.communicate(input=gencdf)
+        to_yaml.wait()
+        out, err = to_yaml.communicate()
+        if dhall.returncode:
+            raise RuntimeError(f'dhall binary failed: {err_d}')
+
+        if to_yaml.returncode:
+            raise RuntimeError(f'dhall-to-yaml binary failed: {err}')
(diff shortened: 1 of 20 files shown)


mssawant:
@knekrasov , will you be submitting a separate PR for the script or provisioning/setup.py changes to generate the CDF using dhall types?

mssawant (approved):
Looks good.

mssawant:
Kindly resolve the DCO check.

provisioning/gencdf.dhall
```diff
@@ -0,0 +1,52 @@
...
+      , io_disks : List Text
+      }
+
+let toNodeDesc
+    : NodeInfo -> T.NodeDesc
+    =     \(n : NodeInfo)
+      ->  { hostname = n.hostname
+          , data_iface = n.data_iface
+          , data_iface_type = Some n.data_iface_type
+          , m0_clients = { other = 3, s3 = 11 }
```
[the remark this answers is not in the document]
knekrasov: @mssawant It's a stub now. 

BTW do you have an idea how many of them should be placed here? Do we need to request a new key to appear in ConfStore for `other`s?

knekrasov:
@mssawant 
> will you be submitting a separate PR for the script or provisioning/setup.py changes to generate the CDF using dhall types?

No, I think I will wait till @SwapnilGaonkar7's #1462 gets merged so I can rebase against it and update setup_hare script.

knekrasov:
@mssawant The work is still in progress but I'd like you to have a look.

As a temporary thing (aka scaffolding), I've already added a separate python package that I'll use for CDF generation. This decision allows me to start writing unit tests (and poke the code from the debugger easily).

This is how I work with it locally:
```
$ pwd
/home/720599/projects/hare/provisioning/miniprov

$ python ./setup.py build
running build
running build_py
copying hare_mp/cdf.py -> build/lib/hare_mp
copying hare_mp/dhall/gencdf.dhall -> build/lib/hare_mp/dhall

$ python ./setup.py develop # Like devinstall; will create symlinks and install the package into the current virtualenv

$ python ./setup.py test # runs tests
```

Do you think it is worth keeping this new package and adopting `setup_hare` code from @SwapnilGaonkar7's? Or it sounds too risky for now?

provisioning/gencdf.dhall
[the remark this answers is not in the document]
mssawant: Presently we reserve 2 m0_clients of other types. This will be some number until we support dynamic addition of clients. So even 3 is fine because there's no good reason to select a number.

mssawant:
>Do you think it is worth keeping this new package and adopting setup_hare code from @SwapnilGaonkar7's? Or it sounds too risky for now?
Instead of 2 separate packages I was thinking if its possible to have just one package including provisioning and hax. Just that we need to produce a separate hare_setup utility. Can we have one setup.py for hax and provisioning? basically all the hare python sources packaged together. We have a different directory hierarchy though.

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -0,0 +1,92 @@
+import os.path as P
```
mssawant, line 1: Not really a fan of single letter variable names.
how about s/P/osPath?

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -0,0 +1,92 @@
...
+import subprocess as S
```
mssawant, line 2: why not just do import `subprocess` and use as it is?

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -0,0 +1,92 @@
...
+from string import Template
+from typing import Dict, Optional, List
+
+import pkg_resources
+
+from hare_mp.store import ValueProvider, as_json, as_list
```
mssawant, line 8: How about s/hare_mp/hare?

provisioning/miniprov/hare_mp/cdf.py
knekrasov: `osPath` is way longer and doesn't differ much from the original package name `os.path`. `P` is short (shorter than the most of variable names), it starts from upper case which differentiates it from variable names)..

provisioning/miniprov/hare_mp/cdf.py
knekrasov: Just a matter of readability. instead of `S.PIPE` and `S.Popen` one would have to write and read ubiquitous `subprocess.PIPE` and `subprocess.Popen`. Although the word "subprocess" doesn't bring that much of value to the reader anyway.

provisioning/miniprov/hare_mp/cdf.py
knekrasov: We already have hare component. Hare is bigger than this package. So I decided to differentiate them.

knekrasov:
@mssawant 
> Instead of 2 separate packages I was thinking if its possible to have just one package including provisioning and hax. Just that we need to produce a separate hare_setup utility. Can we have one setup.py for hax and provisioning? basically all the hare python sources packaged together. We have a different directory hierarchy though.

Hare != hax. We have provisioning for Hare, not for hax.

Please note that Hare has a number of other scripts that do not belong to any Python package. Hax is a python package, but its aim is not connected with provisioning. Hare is not a python package :-)

So I can see two options:
1. Reorganize the sources in Hare so that whole Hare becomes a single valid python package.
   * Rationale: all python sources will be able to reuse common types, functions and classes
2. Have multiple packages within Hare (like we'll have with this patch applied: `hare_mp`, `hax`).
   * Rationale: Hare is not only Python-related. It also contains configuration files, bash scripts etc. So it is frank to keep all python sources as valid packages but we don't say that Hare is only about python.

For me option \[2] looks more reasonable and as a bonus it doesn't require massive refactoring just right away in Hare source folder.

@andriytk, @azheregelya, @vvv, @NeeravChoudhari any thoughts?

andriytk:
@knekrasov sorry, I failed to get into the context of the problem from the summary and description above (although I tried). Could you please elaborate - what's going on here?

knekrasov:
@andriytk It is hard to elaborate it in a single comment. Can you please look through https://github.com/Seagate/cortx-hare/tree/main/rfc/20 ?

andriytk:
I've read it through, several times. Not sure it helped, sorry.

knekrasov:
@andriytk 
1. There will be a new thing for deployment purposes - Mini Provisioner. Hare will need to implement the support.
2. One of the stages of that Mini Provisioner - 'config'. As an input the script in Hare will be given a key-value thing called ConfStore. As a result the CDF must be generated. This is the purpose of this particular PR.

So I generate CDF file according to some input. The logic turned out to be a bit complicated so I wrote several auto tests. Which required me to  wrap the whole logic into Python package rather than having a standalone script.

provisioning/miniprov/hare_mp/cdf.py
mssawant: Yes, but these single letter variable names are not really very intuitive. One has to go look for them what class the belong too. I don't really find them very convenient from reader's perspective. But let's not spend much time on it.

andriytk:
OK, it's clearing up now. So Hare needs to generate CDF now by itself from "a key-value thing called ConfStore" provided as input. Right? That helps, thanks.

knekrasov:
> I only hope that Hare will still be able to understand CDF provided by a human being. This is the way we work with Hare at SAGE.

Yes, CDF will still be there. _Optionally_ there will be a way to generate CDF automatically but we don't remove CDF itself from the bootstrapping procedure.

> I wonder: can we create such a package (or set of packages) in Hare that would allow us to "reuse common types, functions and classes" from one side, but still allow us to "say that Hare is not only about python" from another side?

Well, per my understanding both of the options will address that. Potentially, one package can depend on another (this is like `hax` depends on `python-consul`). @mssawant thinks that it would be simpler to have just one "über" package that contains everything in Hare. Anyway, we'll resolve the packaging question in the next sprint. For now it is decided to keep `miniprov` package as it is for now.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -0,0 +1,328 @@
+import argparse
```
mssawant, line 24: is this new hare_setup.py? the code is duplicated from provisioning/hare_setup.py.

knekrasov: Yes, it is duplicated. I just didn't remove the old hare_setup.py (was afraid to break Provisioning if they rely on that script)

PR #1468, opened 2021-01-25
EOS-16678: Hare CI: VM power on is issued before VM snapshot revert is completed

parikshit-dharmale:
Problem:
Currently sleep of 120 sec is given to finish VM snapshot revert which is not enough.
VM power on time after snapshot revert is more than the current 5 mins wait. 

Solution: Use querying snapshot revert task progress instead of sleep.
Also increase VM start stop wait time.


knekrasov (commented):
The code looks good. 

1. Please resolve DCO issue
2. Please reword the commit message. Commit subject must be quite short (50 chars) and should directly express the problem that this commit fixes. The details should go in message body.

parikshit-dharmale:
> The code looks good.
> 
> 1. Please resolve DCO issue
> 2. Please reword the commit message. Commit subject must be quite short (50 chars) and should directly express the problem that this commit fixes. The details should go in message body.

@knekrasov I have modified the commit message and DCO issue is resolved.

knekrasov (approved):

jenkins/vm-reset
```diff
@@ -128,7 +133,23 @@ vm_snapshot_revert() {
...
     >&2 jq -r '.message' <<< "$resp"
     [[ $(jq '.success' <<< "$resp") == true ]] || die 'Cannot revert snapshot'
-    sleep 120  # XXX Can we do better?
+    
+    local snapshot_task_id
+    snapshot_task_id=$(jq -r '.task_id' <<< "$resp")
+
+    local interval=30  # seconds
+    local wait_steps=8 # wait for snapshot revert
+    while true; do
```
mssawant, line 142: this while loop and wait_till() look similar. Why don't we use wait_till() here instead of duplicating the code?

parikshit-dharmale: @mssawant power and task specific actions has different time out and function call to check the status, to keep it simple I made it separate. We can use wait_till() for both power and task but then we need to add more checks, arguments inside it to monitor current operation.

mssawant: Yes, please try passing the required timeout as an argument. I don't see many of them, it would be good to re-use the function.

jenkins/vm-reset
```diff
@@ -61,18 +61,35 @@ validate_args() {
 
 wait_till() {
     local vm_id=$1
-    local state=$2
+    local action_type=$2
+    local interval
+    local nr_steps
+
+    if [ $action_type == "power" ]; then
```
mssawant, line 68: I don't see the need for `action_type`, 
- just pass the `interval` secs
- calculate `nr_steps` based on the `interval`.

jenkins/vm-reset
```diff
@@ -61,18 +61,35 @@ validate_args() {
...
-    local nr_steps=$((WAIT_TIMEOUT*60 / interval))
     while true; do
         sleep $interval
         echo -n '.'
         ((--nr_steps > 0)) || die 'Timed out'
 
-        if [[ $(vm_power_state $vm_id) == $state ]]; then
-            echo  # add a newline after all those dots
-            return
+        if [ $action_type == "power" ]; then
```
mssawant, line 83: No need of `action_type` I think, you are not doing anything different. Pass the state as a parameter and check or the same.

parikshit-dharmale: @mssawant modified the code as per comments. Please review it.

jenkins/vm-reset
```diff
@@ -61,18 +61,31 @@ validate_args() {
 
 wait_till() {
     local vm_id=$1
-    local state=$2
+    local interval=$2
+    local state=$3
+    local nr_steps
+
+    if [ $interval -eq 5 ]; then
```
mssawant, line 68: - How interval value changes?
- pass WAIT_TIMEOUT as an argument to `wait_till()` function and have just one way to calculate `nr_steps` based on `WAIT_TIMEOUT`.  See if interval can be kept constant.

jenkins/vm-reset
```diff
@@ -61,18 +60,26 @@ validate_args() {
 
 wait_till() {
```
mssawant, line 61: please add some documentation on how this function is supposed to be used and what the arguments mean.
May be we can still simplify the use of it but let's complete this patch. If the required functionality is working its okay. 
Just add some documentation for this function.

mssawant (approved):

jenkins/vm-reset
parikshit-dharmale: @mssawant  added documentation for wait_till() function.

jenkins/vm-reset
```diff
@@ -59,20 +58,34 @@ validate_args() {
         die 'Cannot communicate with CloudForms Management Engine'
 }
 
+# Wait till the certain condition is met or timout occurs
```
mssawant, line 61: - s/timout/timeout
- ',' or '.' after occurs.

jenkins/vm-reset
```diff
@@ -59,20 +58,34 @@ validate_args() {
...
+# currently power state (on/off) and snapshot task progress is checked
```
mssawant, line 62: - Start a new line with capital letter or add a ',' to previous line.

jenkins/vm-reset
```diff
@@ -59,20 +58,34 @@ validate_args() {
...
+# It takes inputs,
+#           vm_id         id of VM
```
mssawant, line 64: No need of the spaces at the start of param lines.

parikshit-dharmale:
@mssawant Addressed review comments.

knekrasov (approved):

PR #1471, opened 2021-01-27
cfgen: improve assert msg at pool drives check

andriytk:
Currently, the assert msg looks like this:

    File "/opt/seagate/cortx/hare/bin/../bin/cfgen", line 1094, in pool_drives
      assert len(targets) == 1  # guaranteed by validate_pools_desc()
    AssertionError

which gives no clue about what's wrong with the pool configuration.

Solution: print config line which fails the assertion.

### DIFF
--- a/cfgen/cfgen
+++ b/cfgen/cfgen
@@ -1092,6 +1092,7 @@ def pool_drives(m0conf: Dict[Oid, Any],
                        for drive_id, node_id in all_drives
                        if ref['path'] == m0conf[m0conf[drive_id].sdev].filename
                        and ref.get('node') in (None, m0conf[node_id].name)]
+            # XXX Improve validate_pools_desc() to catch this error.
             assert len(targets) == 1, 'Check {} config line'.format(ref)
             drives.extend(targets)
         assert all_unique(drives), \
--- a/cfgen/cfgen
+++ b/cfgen/cfgen
@@ -1092,7 +1092,7 @@ def pool_drives(m0conf: Dict[Oid, Any],
                        for drive_id, node_id in all_drives
                        if ref['path'] == m0conf[m0conf[drive_id].sdev].filename
                        and ref.get('node') in (None, m0conf[node_id].name)]
-            assert len(targets) == 1  # guaranteed by validate_pools_desc()
+            assert len(targets) == 1, 'Check {} config line'.format(ref)
             drives.extend(targets)
         assert all_unique(drives), \
             'Pool {!r}: some of disk_refs refer to the same disk'.format(


andriytk:
cc @vvv

cfgen/cfgen
```diff
@@ -1092,7 +1092,10 @@ def pool_drives(m0conf: Dict[Oid, Any],
                        for drive_id, node_id in all_drives
                        if ref['path'] == m0conf[m0conf[drive_id].sdev].filename
                        and ref.get('node') in (None, m0conf[node_id].name)]
-            assert len(targets) == 1  # guaranteed by validate_pools_desc()
+            assert len(targets) == 1, \
+                '{} {}'.format(ref['path'], 'has wrong node: {}'.format(
```
knekrasov, line 1096: For me this code looks too complicated. `ref['path']` could have been refactored into a variable to improve readability, probably there is also a way to avoid ternary operator..

cfgen/cfgen
```diff
@@ -1092,7 +1092,10 @@ def pool_drives(m0conf: Dict[Oid, Any],
...
+                                       ref['node']) if len(targets) == 0
+                                   else 'is preset at more than one node')
```
vvv, line 1098: 1. @andriytk Instead of changing this `assert`, I'd recommend enforcing `validate_pools_desc()` and catching the error there.

2. `ref['node']` may throw `KeyError` exception.
   E.g., in [cfgen/examples/multipools.yaml] only first two of the disk references have 'node' key.
   ```yaml
   pools:
     - name: tier1-nvme
       disk_refs:
         - path: /dev/foo
           node: srvnode-1
         - path: /dev/foo
           node: srvnode-2
       data_units: 1
       parity_units: 0
     - name: tier2-ssd
       disk_refs:
         - path: /dev/disk/by-id/dm-name-mpatha
         - path: /dev/disk/by-id/dm-name-mpathb
         - path: /dev/disk/by-id/dm-name-mpathg
         - path: /dev/disk/by-id/dm-name-mpathh
   ```

[cfgen/examples/multipools.yaml]: https://github.com/Seagate/cortx-hare/blob/6eb3a1c801b83b7fc846a8ac90bef5df73955bc1/cfgen/examples/multipools.yaml#L42

vvv (changes requested):

cfgen/cfgen
andriytk: > I'd recommend enforcing validate_pools_desc()

@vvv how to do this?

andriytk:
I simplified the assert msg after the above comments. It's still useful and definitely better than nothing.

cfgen/cfgen
vvv: @andriytk Could you attach a CDF that causes assertion error?

cfgen/cfgen
```diff
@@ -1092,7 +1092,7 @@ def pool_drives(m0conf: Dict[Oid, Any],
                        for drive_id, node_id in all_drives
                        if ref['path'] == m0conf[m0conf[drive_id].sdev].filename
                        and ref.get('node') in (None, m0conf[node_id].name)]
-            assert len(targets) == 1  # guaranteed by validate_pools_desc()
+            assert len(targets) == 1, 'Check {} config line'.format(ref)
```
vvv, line 1095:
```suggestion
            # XXX Improve validate_pools_desc() to catch this error.
            assert len(targets) == 1, 'Check {} config line'.format(ref)
```

vvv (approved):

cfgen/cfgen
andriytk: There was a problem with the node, so I disabled it from the list of nodes nodes, but forgot to disable its disks in the pool...

knekrasov (approved):

PR #1475, opened 2021-01-29
hare.spec: fix consul version requirement

andriytk:
Currently, the requirement allows only 1.7.8 version of Consul which is very strict. Hare is known to work fine with a much wider range of Consul versions.

Solution: fix the requirement to allow versions of Consul from 1.7.x to 1.9.x.

### DIFF
--- a/hare.spec
+++ b/hare.spec
@@ -45,7 +45,7 @@ BuildRequires: python36-devel
 BuildRequires: python36-pip
 BuildRequires: python36-setuptools
 
-Requires: consul = 1.7.8
+Requires: consul >= 1.7.0, consul < 1.10.0
 Requires: facter
 Requires: jq
 Requires: cortx-motr = %{h_motr_version}


andriytk:
cc @chumakd

hare.spec
```diff
@@ -45,7 +45,7 @@ BuildRequires: python36-devel
 BuildRequires: python36-pip
 BuildRequires: python36-setuptools
 
-Requires: consul = 1.7.8
+Requires: consul >= 1.7.0
```
knekrasov, line 48: Isn't it too risky to allow ANY version higher than 1.7.0?
For instance, if Consul 2.0.0 gets released, that will satisfy the predicate in our specfile. But we didn't test our application against 2.0.0. 

The stack will become broken but the problem will not be discovered during build stage.

andriytk: I'd suggest putting the restriction whenever such a problem is discovered.

knekrasov: Ok, let's imagine the situation as follows:
1. CORTX stack comes to release date (e.g. R2), bugs are fixed, everybody is happy.
2. After testing stage is done, Consul releases new version.
3. In the meantime, we're going to impress a new, very important beta customer and provide them a build.
4. Build succeeds (although it will contain Consul version which we have never been tested against).
5. The build is deployed at the customer's side. But it fails.

I've seen similar situation before at other projects, so it is not "fully imaginary". Such things happen and they hit quite painful.

It is very nice when we use the latest software but probably we can do that until code freeze. After it all the versions must be frozen.

## Suggestion
1. Land this quite permissive PR as it is.
2. Create another Github issue to "freeze" the latest Consul version at some date in the future - before the testing phase starts.

@mssawant, @NeeravChoudhari, @s-arya What's your opinion.

andriytk: > Build succeeds (although it will contain Consul version which we have never been tested against).

@knekrasov do you mean that this line regulates which Consul version will be downloaded? I agree then, we better limit it from the top in this case. Let me update the patch...

mssawant:
@andriytk, @knekrasov any reason we are not using https://github.com/Seagate/cortx-hare/pull/1463 but having multiple PRs for the same problem? why was this approach not commented in https://github.com/Seagate/cortx-hare/pull/1463?

andriytk:
@mssawant, sorry - I just forgot about your PR when doing mine. :)

If it's a problem - we can close this one and update your one. Or vice versa - whatever is more convenient for you guys.

mssawant:
@andriytk, Yes, let's use one of these or let's merge https://github.com/Seagate/cortx-hare/pull/1463 and then continue here as we have already started discussions in this PR. So if you can add a comment in https://github.com/Seagate/cortx-hare/pull/1463, something like "further discussions are continued in https://github.com/Seagate/cortx-hare/pull/1475", approve and merge https://github.com/Seagate/cortx-hare/pull/1463, it would be good. We will continue the further improvement discussions here.

hare.spec
mssawant: Few points,
1. By adding a range of allowed Consul versions the benefit is that we need not have to update the code frequently.
2. But having a range of allowed Consul versions brings in additional task to verify if all the Consul versions in the range to be tested with Hare. But may be we can just test hare with just the lower and the higher version available and assume that all the in-between versions work with hare.
3. We would need to update the range if there exists a Consul version that is out of bounds anyway, this is the case with having a fixed version as well.

If we are okay with the assumption in point 2 then it looks fine.

andriytk: It's OK with me.

mssawant (approved):

hare.spec
```diff
@@ -45,7 +45,7 @@ BuildRequires: python36-devel
 BuildRequires: python36-pip
 BuildRequires: python36-setuptools
 
-Requires: consul = 1.7.8
+Requires: consul >= 1.7.0, consul < 1.10.0
```
mssawant, line 48: The last version we tested Hare with was 1.9.1, I think we should restrict the higher limit to the last tested version.

andriytk: They land only fixes when the last number is increased which don't break anything usually. I think it's safe to assume that if it works with 1.9.1 - it will work with any future 1.9.x versions then.

mssawant: Fair enough.

chumakd (approved):

knekrasov:
@mssawant Do we want to merge this PR?

mssawant:
@knekrasov , yes, sorry I was waiting for your approval. Let's merge it.

knekrasov (approved):

PR #1476, opened 2021-01-30
Problem: cfgen fails to parse CDF generated by mini provisioner

mssawant:
Mini provisioner CDF generator adds few default values for some of the
configuration keys. But hare cfgen utility that parses the generated
CDF does not allow some of the defined keys to have no values.
cfgen does not allow,
- meta-data devices to be None
- diskrefs must node be empty
- pool type cannot be null
- List of profiles must not be empty
- data_iface_type cannot be null

Solution:
- Fetch meta data devices from cortx-py-utils:conf-store and
  pass the values to Hare mini provisioner.
- Allowed diskrefs to be empty, but this needs to be fixed properly.
- Set default pool type to be sns.
- Set default non-empty list of profiles.
- Passed tcp as a default network interface type to mini provisioner.
- Add --filename option along with with config in setup.yaml

### DIFF
--- a/cfgen/cfgen
+++ b/cfgen/cfgen
@@ -331,11 +331,13 @@ Pools of {t!r} type can only have 1 data unit."""
         assert t is PoolT.sns, \
             f"{name}: disk_refs are only supported for pools of 'sns' type"
         disk_refs = pool['disk_refs']
-        assert disk_refs, \
-            f'Pool {name!r}: disk_refs must not be empty' \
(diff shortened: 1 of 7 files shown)


mssawant:
This patch is in continuation of https://github.com/Seagate/cortx-hare/pull/1467 and fixes some issues with respect cfgen parsing of cdf generated by mini provisioner in https://github.com/Seagate/cortx-hare/pull/1467. So going ahead and merging the patch.

PR #1484, opened 2021-02-02
Problem: No module named 'cortx' when running tests

knekrasov:
Mini Provisioner depends on ConfStore which is imported from
cortx.utils.

The problem is that Hare build pipeline runs in a separate virtual
environment (a.k.a. virtualenv). By default virtualenv doesn't inherit
system-wide packages from `/usr/lib/python3.6/site-packages`. It means
that even though cortx-py-utils RPM is installed, its Python packages
are not visible from Hare virtualenv.

## Solution
1. Create virtualenv with `--system-site-packages` flag.
2. After that local requirements.txt must be processed with
`--ignore-installed` flag to force pip to overwrite the packages in the
local site-packages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,11 +94,11 @@ $(MP_WHL): $(PY_VENV_DIR) $(HAX_SRC)
 
 $(PY_VENV_DIR): $(patsubst %,%/requirements.txt,cfgen hax provisioning/miniprov)
 	@$(call _info,Initializing virtual env in $(PY_VENV_DIR))
-	@$(PYTHON) -m venv $@
+	@$(PYTHON) -m venv --system-site-packages $@
 	@$(call _info,Installing pip modules in virtual env)
 	@for f in $^; do \
 	     $(call _log,processing $$f); \
-	     $(PIP) install -r $$f; \
+	     $(PIP) install --ignore-installed -r $$f; \
 	 done
 
 # Clean ----------------------------------------------- {{{1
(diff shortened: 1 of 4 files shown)


mssawant (approved):

Makefile
```diff
@@ -94,11 +94,11 @@ $(MP_WHL): $(PY_VENV_DIR) $(HAX_SRC)
 
 $(PY_VENV_DIR): $(patsubst %,%/requirements.txt,cfgen hax provisioning/miniprov)
 	@$(call _info,Initializing virtual env in $(PY_VENV_DIR))
-	@$(PYTHON) -m venv $@
+	@$(PYTHON) -m venv --system-site-packages $@
 	@$(call _info,Installing pip modules in virtual env)
 	@for f in $^; do \
 	     $(call _log,processing $$f); \
-	     $(PIP) install -r $$f; \
+	     $(PIP) install --ignore-installed -r $$f; \
```
suvratjoshi, line 101: As I understand, we are creating a virtual room with access to global packages by having '--system-site-packages' flag.
Are there any additional packages in global environment than defined in requirements.txt and required by hare build?
i.e. Is the package list in requirements.txt is a subset of global list of packages?

knekrasov: @suvratjoshi 

The idea of virtual environments is that 
1. I have a folder where my Python dependencies can be installed without root permissions
2. If I install some package to the virtualenv, I don't change the global `site-packages/`. So this kind of package manipulation will not harm any other Python app.
3. When I use virtualenv, Python interpreter will look into the virtualenv folder to resolve the imports.

So the major change of this PR is point \[3]. From now on, if a package IS NOT in the virtualenv, it will be searched in the global site-packages. Before this patch, such an import would simply fail.

requirements.txt may or may not be a "subset" of global packages. I guess it is doesn't really matter. The idea is that virtualenv is a self-sufficient sandbox where my application will run for sure (so all the deps are installed and have correct versions). cortx-py-utils installs itself into the global `site-packages/` and thus it didn't work with virtualenv. So we add a workaround.

suvratjoshi (approved):

PR #1489, opened 2021-02-03
Problem: logrotate configuration is not optional

mssawant:
hare_setup utility by default tries to install logrotate configuration which
requires provisioner cli to be installed. It is possible that on some dev setups
provisioner is not installed, in which case hare_setup throws an error that may
mislead user to assume that hare_setup command itself failed.
Also it is not required logrotate to be configured in context of every command
execution.

Solution:
Provide an explicit option to hare_setup utility to configure hare logrotate
values.

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -75,24 +75,24 @@ def get_data_from_provisioner_cli(method, output_format='json') -> str:
     res = stdout.decode('utf-8')
     return json.loads(res)['ret'] if res else 'unknown'
 
-
-def logrotate_config():
-    try:
-        setup_info = get_data_from_provisioner_cli('get_setup_info')
-        if setup_info != 'unknown':
-            server_type = setup_info['server_type']
-            shutil.copyfile(
-                f'/opt/seagate/cortx/hare/conf/logrotate/{server_type}',
-                '/etc/logrotate.d/hare')
-    except Exception as error:
-        logging.error('Error setting logrotate values for hare (%s)', error)
-
-
 def _report_unsupported_features(features_unavailable):
     uf_db = unsupported_features.UnsupportedFeaturesDB()
     loop = asyncio.get_event_loop()
     loop.run_until_complete(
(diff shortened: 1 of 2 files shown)


SwapnilGaonkar7:
If I am not wrong then hare_setup will still throw an error if provisioner is not installed.
Can we check if provisioner is installed and then attempt logrotate functionality in hare_setup.

knekrasov (approved):

knekrasov:
@SwapnilGaonkar7 No, the idea is different. 

If the user does want to configure logrotate, then (s)he will add the corresponding flag when invoking `hare_setup`. If the user doesn't want it (or it is obvious that `provisioner` CLI is not available), the argument will not be passed. So it is up to user to decide.

Your approach has another drawback. Consider the user does want to configure logrotate, then `hare_setup` will check if `provisioner` CLI is available and only then configure log rotate. So if provisioner CLI is not available, the end user will even not know that something went wrong.

SwapnilGaonkar7:
> @SwapnilGaonkar7 No, the idea is different.
> 
> If the user does want to configure logrotate, then (s)he will add the corresponding flag when invoking `hare_setup`. If the user doesn't want it (or it is obvious that `provisioner` CLI is not available), the argument will not be passed. So it is up to user to decide.
> 
> Your approach has another drawback. Consider the user does want to configure logrotate, then `hare_setup` will check if `provisioner` CLI is available and only then configure log rotate. So if provisioner CLI is not available, the end user will even not know that something went wrong.

But we are still calling log rotate in post_install section. So we will get error message if provisioner is not installed

mssawant:
@SwapnilGaonkar7 ,
>But we are still calling log rotate in post_install section. So we will get error message if provisioner is not installed
We have included logrotate flag in setup.yaml because setup.yaml is supposed to used by the provisioner. Other users, mainly dev and QA will invoke hare_setup utility directly. If mini provisioner is invoked in Provisioner's context then it is expected that provisioner cli is present.
We want to configure logrotate in production environment, thus we have included `logrotate` flag in setup.yaml.

SwapnilGaonkar7 (approved):

PR #1490, opened 2021-02-04
Problem: CDF file uses metadata device instead of volume

knekrasov:
Solution: use the partition name instead while generating CDF. At the
moment the value is not provided through ConfStore so the value is
hardocded with TODO comment.


nitin-seagate (approved):
Looks good to me...

PR #1491, opened 2021-02-04
Problem: Hare miniprovisioner post_install section do not check consul version as per requirement

SwapnilGaonkar7:
Solution: Added code to get consul version and check if matched requirement mentioned by cortx-hare
Note: Currently only numeric characters are supported for version number.

UT: Tested post_install section by running below command
[root@ssc-vm-c-1822 hare]# /opt/seagate/cortx/hare/bin/hare_setup --post_install
'cortx-s3server' is not installed
Failed to fetch data from provisioner ([Errno 2] No such file or directory: 'provisioner': 'provisioner')

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -38,6 +38,88 @@ from hare_mp.cdf import CdfGenerator
 from hare_mp.store import ConfStoreProvider
 
 
+# returns 1 if v1 is bigger
+#         -1 if v2 is bigger
+#         0 if equal
+def version_compare(v1: str, v2: str) -> int:
+
+    # This will split both the versions by '.'
+    arr1 = v1.split(".")
+    arr2 = v2.split(".")
+    n = len(arr1)
+    m = len(arr2)
+
+    # converts to integer from string
+    arr1 = [int(i) for i in arr1]
+    arr2 = [int(i) for i in arr2]
+
+    # compares which list is bigger and fills
+    # smaller list with zero (for unequal delimeters)
+    if n>m:
+      for i in range(m, n):
+         arr2.append(0)
+    elif m>n:
+      for i in range(n, m):
+         arr1.append(0)
+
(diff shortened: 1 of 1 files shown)


SwapnilGaonkar7:
@knekrasov @mssawant Please review

provisioning/miniprov/hare_mp/main.py
```diff
@@ -38,6 +38,88 @@
...
+def version_compare(v1: str, v2: str) -> int:
+
+    # This will split both the versions by '.'
+    arr1 = v1.split(".")
+    arr2 = v2.split(".")
+    n = len(arr1)
+    m = len(arr2)
+
+    # converts to integer from string
+    arr1 = [int(i) for i in arr1]
```
knekrasov, line 53: Versions can have letters, e.g. 2.0.32-rc

Here are the version names from Consul history: https://releases.hashicorp.com/consul/

Can we give a try to something like this instead of re-inventing the wheel: https://pypi.org/project/cmp_version/ ?

knekrasov: @mssawant Can you remind me what is the result of this discussion? Are we good to add external library to check the versions? And do we need to check the versions given we already have that check at RPM layer?

@SwapnilGaonkar7 My rule of a thumb would be
- If we can't leverage third party library but need to keep your hand-made version check, then we need unit tests that will prove that your version comparison work.
- If we can leverage 3rd party library, let's our own implementation of the same algorithms.

knekrasov: @SwapnilGaonkar7 any updates on this PR?

SwapnilGaonkar7: @knekrasov It was decided that as consul is common 3rd party utility used by many components, all the checks related to it must be done in common location(mostly provisioner) and not in specific component like hare.
So I think we can safely close(discard) this PR.

PR #1494, opened 2021-02-05
Problem: data_iface_type value is not taken from ConfStore

knekrasov:
JIRA: https://jts.seagate.com/browse/EOS-17250

Solution: take the type value from ConfStore under key
`cluster>{nodename}>network>data>interface_type`


SwapnilGaonkar7 (approved):
LGTM

mssawant:
@knekrasov, has provisioner added data_iface_type kv to conf-store?

PR #1501, opened 2021-02-10
Problem: Hare doesn't handle node failure properly

mssawant:
If a node is rebooted Consul does report a warning for the corresponding
service to hax but hax checks for node status which is 'passing' atleast
for few iterations, this delay in node status update leads to missing of
the Consul warning, Consul status for the process is not update in KV.
Consul KV for the process status still remains M0_CONF_HA_PROCESS_STARTED
which also causes node-join to fail.
While notifying service failure to peer motr processes it is also important
to handle delay in notification mainly in case of transient failure.
The process for which Consul reported failure might just be starting and
reporting a process failure when hax has already replied to the entrypoint
request leads to other inconsistencies with halink protocol.
Hax also does not send failure notifications for the disks corresponding
to an ioservice. This is needed to select the appropriate pool version
when writing post node failure.

Solution:
- When Consul reports process failure before broadcasting the failure to
  motr processes, check the service status one more time before broadcasting.
  This takes care of delay in failure notification if the process has already
  started. If the process has already restarted hax has already notified peer motr
  processes about this motr process on receiving its first entrypoint request.
- Update Consul status when broadcasting failure, this handles node rejoining.
- Send failure notifications for the disks corressponding to a failed ioservice.
- increase max size of consumer queue to 512, noticed a hang in processing
  in case of heavy message flow.

### DIFF
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -31,8 +31,8 @@ from hax.motr import Motr
 from hax.motr.delivery import DeliveryHerald
 from hax.queue.publish import EQPublisher
 from hax.types import (ConfHaProcess, HAState, MessageId, StobIoqError,
-                       m0HaProcessEvent, ServiceHealth, StoppableThread,
-                       HaLinkMessagePromise)
+                       m0HaProcessEvent, m0HaProcessType, ServiceHealth,
+                       StoppableThread, HaLinkMessagePromise, ObjT)
 from hax.util import ConsulUtil, dump_json, repeat_if_fails
 
 LOG = logging.getLogger('hax')
@@ -61,23 +61,59 @@ class ConsumerThread(StoppableThread):
         self.is_stopped = True
 
     @repeat_if_fails(wait_seconds=1)
-    def _update_process_status(self, event: ConfHaProcess) -> None:
+    def _update_process_status(self, q: Queue, event: ConfHaProcess) -> None:
         # If a consul-related exception appears, it will
         # be processed by repeat_if_fails.
         #
         # This thread will become blocked until that
         # intermittent error gets resolved.
         self.consul.update_process_status(event)
+        svc_status = m0HaProcessEvent.event_to_svchealth(event.chp_event)
+        if event.chp_type == m0HaProcessType.M0_CONF_HA_PROCESS_M0D:
+            # Broadcast the received motr process status to other motr
+            # processes in the cluster.
+            q.put(BroadcastHAStates(states=[HAState(fid=event.fid,
+                                    status=svc_status)],
+                  reply_to=None))
 
-    def update_process_failure(self, ha_states: List[HAState]) -> None:
+    @repeat_if_fails(wait_seconds=1)
+    def update_process_failure(self, q: Queue,
+                               ha_states: List[HAState]) -> List[HAState]:
+        new_ha_states: List[HAState] = []
         for state in ha_states:
-            if state.status == ServiceHealth.FAILED:
-                m0status = m0HaProcessEvent.M0_CONF_HA_PROCESS_STOPPED
-                pevent = ConfHaProcess(chp_event=m0status,
-                                       chp_type=3,
-                                       chp_pid=0,
-                                       fid=state.fid)
-                self._update_process_status(pevent)
+            # We are only concerned with process statuses.
+            if state.fid.container == ObjT.PROCESS.value:
+                current_status = self.consul.get_process_current_status(
+                                     state.status, state.fid)
+                if current_status == ServiceHealth.FAILED:
+                    self.consul.service_health_to_m0dstatus_update(
+                        state.fid, current_status)
+                elif current_status == ServiceHealth.UNKNOWN:
+                    # We got service status as UNKNOWN, that means hax was
+                    # notified about process failure but hax couldn't
+                    # confirm if the process is in failed state or have
+                    # failed and restarted. So, we will not loose the
+                    # event and try again to confirm the real time
+                    # process status by enqueing a broadcast event
+                    # specific to this process.
+                    # It is expected that the process status gets
+                    # eventually confirmed as either failed or passing (OK).
+                    # This situation typically arises due to delay
+                    # in receiving failure notification during which the
+                    # corresponding process might be restarting or have
+                    # already restarted. Thus it is important to confirm
+                    # the real time status of the process before
+                    # broadcasting failure.
+                    current_status = ServiceHealth.OK
+                    q.put(BroadcastHAStates(
+                          states=[HAState(fid=state.fid,
+                                          status=ServiceHealth.FAILED)],
+                          reply_to=None))
+                new_ha_states.append(HAState(fid=state.fid,
+                                             status=current_status))
+            else:
+                new_ha_states.append(state)
+        return new_ha_states
 
     def _do_work(self, q: Queue, motr: Motr):
         ffi = motr._ffi
@@ -126,7 +162,7 @@ class ConsumerThread(StoppableThread):
                         # hence will need to make new attempt by itself
                         motr.send_entrypoint_request_reply(item)
                     elif isinstance(item, ProcessEvent):
-                        self._update_process_status(item.evt)
+                        self._update_process_status(q, item.evt)
                     elif isinstance(item, HaNvecGetEvent):
                         fn = motr.ha_nvec_get_reply
                         # If a consul-related exception appears, it will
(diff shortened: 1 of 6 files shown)


hax/hax/util.py
```diff
@@ -635,6 +696,10 @@ def ensure_motr_all_started(self, event: Event):
                 return
             wait_for_event(event, 5)
 
+    def fid2objT(self, fid: str) -> str:
```
knekrasov, line 699: What strings can this method return? If we know the full vocabulary of possible return values, does it make sense to return a enum or something? That would make the contract clear.

hax/hax/handler.py
```diff
@@ -61,23 +61,70 @@ def stop(self) -> None:
...
+        svc_status = m0HaProcessEvent.event_to_svchealth(event.chp_event)
+        if event.chp_type == m0HaProcessType.M0_CONF_HA_PROCESS_M0D:
+            # Broadcast the received motr process status to other motr
+            # processes in the cluster.
+            q.put(BroadcastHAStates(states=[HAState(fid=event.fid,
+                                    status=svc_status)],
+                  reply_to=None))
 
-    def update_process_failure(self, ha_states: List[HAState]) -> None:
+    def update_process_failure(self, q: Queue,
```
knekrasov, line 79: It seems like the method needs to be decorated with `@repeat_if_fails()`
Reason: the code within can raise `HAConsistencyException`. The method is invoked directly from `qconsumer` thread (handler.py) and it seems like the exception is not handled there. So `qconsumer` will pull the message from the thread Queue and once the exception raises, the message will be forgotten and will not be processed afterwards.

hax/hax/handler.py
```diff
@@ -61,23 +61,70 @@ def stop(self) -> None:
...
+                    proc_status = self.consul.get_process_status(state.fid)
+                    svc_status = self.consul.get_service_health(node,
+                                                                state.fid.key)
+                    LOG.info('node: %s proc: %s proc_state: %s health: %s',
+                             node, state.fid, proc_status, svc_status)
+                    if svc_status == ServiceHealth.FAILED:
+                        m0status = m0HaProcessEvent.M0_CONF_HA_PROCESS_STOPPED
+                        if proc_status != 'M0_CONF_HA_PROCESS_STOPPED':
+                            pevent = ConfHaProcess(chp_event=m0status,
+                                                   chp_type=3,
```
knekrasov, line 97: what `chp_type=3` stands for?

hax/hax/hax.py
```diff
@@ -89,7 +89,7 @@ def main():
     # _run_qconsumer_thread function.
     #
     # [KN] Note: The server is launched in the main thread.
-    q: Queue = Queue(maxsize=8)
+    q: Queue = Queue(maxsize=512)
```
knekrasov, line 92: 👍

hax/hax/types.py
```diff
@@ -42,6 +42,16 @@ class Uint128Struct(c.Structure):
 ObjT.__doc__ = 'Motr conf object types and their m0_fid.f_container values'
 
 
+FidT = Enum(
```
knekrasov, line 45: According to the way how FidT is used, it feels more like a static map `fid container` -> `string name`. Why wouldn't we convert it into the static dict then?

knekrasov (changes requested):

hax/hax/handler.py
```diff
@@ -61,23 +61,70 @@ def stop(self) -> None:
...
         for state in ha_states:
-            if state.status == ServiceHealth.FAILED:
-                m0status = m0HaProcessEvent.M0_CONF_HA_PROCESS_STOPPED
-                pevent = ConfHaProcess(chp_event=m0status,
-                                       chp_type=3,
-                                       chp_pid=0,
-                                       fid=state.fid)
-                self._update_process_status(pevent)
+            if self.consul.fid2objT(str(state.fid)) == 'PROCESS':
+                svc_status = state.status
```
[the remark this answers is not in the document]
mssawant: Putting everything in a single function was a bit tricky so split the code into 2 utility functions.

hax/hax/handler.py
```diff
@@ -61,23 +61,70 @@ def stop(self) -> None:
...
+                    # check one more time before broadcasting failure.
+                    node = self.consul.get_process_node(state.fid)
+                    proc_status = self.consul.get_process_status(state.fid)
+                    svc_status = self.consul.get_service_health(node,
+                                                                state.fid.key)
+                    LOG.info('node: %s proc: %s proc_state: %s health: %s',
+                             node, state.fid, proc_status, svc_status)
+                    if svc_status == ServiceHealth.FAILED:
+                        m0status = m0HaProcessEvent.M0_CONF_HA_PROCESS_STOPPED
+                        if proc_status != 'M0_CONF_HA_PROCESS_STOPPED':
```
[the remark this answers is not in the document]
mssawant: Yeah, removed `proc_status != STOPPED` as it was redundant.

hax/hax/handler.py
mssawant: Its `m0HaProcessType.M0_CONF_HA_PROCESS_M0D`, updated.

hax/hax/types.py
mssawant: Yeah, it was a leftover. fixed the usage.

hax/hax/util.py
```diff
@@ -581,31 +609,64 @@ def set_m0_disk_state(self, fid: str, objstate: int) -> None:
...
         except (ConsulException, HTTPError, RequestException) as e:
             raise HAConsistencyException('Failed to communicate '
                                          'to Consul Agent') from e
         return status
 
+    def get_process_node(self, proc_fid: Fid) -> str:
+        fidk = proc_fid.key
+        # 'node/<node_name>/process/<process_fidk>/service/type'
+        node_items = self.kv.kv_get('m0conf/nodes', recurse=True)
+        # TODO [KN] This code is too cryptic. To be refactored.
```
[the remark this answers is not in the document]
mssawant: Removed.

hax/hax/util.py
mssawant: ObjT for a given fid, added a docstring.

hax/hax/util.py
```diff
@@ -581,31 +609,64 @@ def set_m0_disk_state(self, fid: str, objstate: int) -> None:
         self.kv.kv_put(key, data)
 
     @repeat_if_fails()
-    def get_service_health(self, service_name: str,
-                           node: str, svc_id: int) -> ServiceHealth:
-
+    def get_service_health(self, node: str, svc_id: int) -> ServiceHealth:
```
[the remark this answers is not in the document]
mssawant: Added a docstring.

hax/hax/util.py
```diff
@@ -635,6 +696,10 @@ def ensure_motr_all_started(self, event: Event):
                 return
             wait_for_event(event, 5)
 
+    def fid2objT(self, fid: str) -> str:
+        cont = f'0x{Fid.parse("0x7200000000000001:0x9").container:x}'
```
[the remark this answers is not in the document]
mssawant: fixed this.

hax/hax/handler.py
```diff
@@ -4,7 +4,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#    http://www.apache.org/licenses/LICENSE-2.0
+#    http://www.apache.org/lic://github.com/Seagate/cortx-motr/blob/b22c3dd234b7885b6528e8870192f00816d75048/pool/pool.c#L659enses/LICENSE-2.0
```
knekrasov, line 7: Is this hyperlink correct?

hax/hax/util.py
knekrasov: My point was that raw strings as result type is error prone and tools like mypy will fail to help if I mistype the string when comparing the result value to something.

But OK, we can address this later in another PR. Would you please add FIXME for the future?

hax/hax/types.py
knekrasov: Sorry, I still don't understand why you need it to be a Enum. According to what I see, you need the mapping, it doesn't need to be a separate type. And you use it like a dict, actually

```python
    def fid2objT(self, fid: str) -> str:
        """
        Returns ObjT for a given fid as a string.
        """
        cont = f'0x{Fid.parse(fid).container:x}'
        return str(FidT[cont].value)
```

So effectively FidT could either be a dict or a function that performs this mapping (while dict could be an implementation detail). And I would vote for function. Do I miss anything?

knekrasov (changes requested):
A couple of minor things left.

hax/hax/handler.py
suvratjoshi: Seems like motr's pool.c link mistakenly got copied here

mssawant: Damn, must have got pasted mistakenly from the cache, restored.

hax/hax/util.py
mssawant: No, I agree. I see, this is not needed actually. Have updated the patch.

hax/hax/types.py
mssawant: Removed, not needed. Thanks.

knekrasov (approved):

suvratjoshi (approved):

PR #1503, opened 2021-02-12
EOS-17407: hare-shutdown can leave the cluster in non-stoppable state

knekrasov:
Problems:
1. SIGKILL doesn't stop hare-shutdown script properly. There is a chance that local consul will be killed and thus hare-shutdown will not work next time.
2. There is a chance that local consul will be stopped not the last. If an error occurs after that, hare-shutdown will not work next time.

### DIFF
--- a/utils/hare-shutdown
+++ b/utils/hare-shutdown
@@ -20,13 +20,14 @@
 # help stop the cluster or local node
 
 import getopt
-import sys
 import logging
-import utils
-from socket import gethostname
+import sys
+from typing import Dict, List, Tuple
+
 from consul import Consul
-from hax.util import repeat_if_fails
-from typing import Dict, List
+from hax.util import ConsulUtil, repeat_if_fails
+
+import utils
 
 
 def check_cluster():
@@ -38,7 +39,7 @@ def check_cluster():
 
 
 @repeat_if_fails()
-def _get_processes_for_cluster():
+def _get_processes_for_cluster() -> Tuple[Dict[str, List[utils.Process]], str]:
     cns = Consul()
     processes = utils.processes_by_consul_svc_name(cns)
     leader_node = utils.get_kv(cns, 'leader')
@@ -46,17 +47,30 @@ def _get_processes_for_cluster():
 
 
 @repeat_if_fails()
-def _get_processes_for_node():
+def _get_processes_for_node() -> Dict[str, List[utils.Process]]:
     cns = Consul()
-    node = gethostname()
+    cns_util = ConsulUtil(raw_client=cns)
+    node = cns_util.get_local_nodename()
     processes = utils.processes_node(cns, node)
     return processes
 
 
 def stop_process(processes: Dict[str, List[utils.Process]]):
     for svc in utils.shutdown_sequence:
-        if svc in processes:
+        if svc not in processes:
+            continue
+        if svc != 'consul':
             utils.stop_parallel([proc for proc in processes[svc]])
+        else:
+            # Consul process at the current node must stopped the last.
+            # Reason: if something goes wrong and the user will want to
+            # shutdown the cluster again, local Consul must be available.
+            remote_procs = [
+                proc for proc in processes[svc] if not proc.is_local
+            ]
+            utils.stop_parallel(remote_procs)
+            local_procs = [proc for proc in processes[svc] if proc.is_local]
+            utils.stop_parallel(local_procs)
 
 
 def shutdown_cluster():
@@ -78,9 +92,7 @@ def node_shutdown():
 
 def main() -> None:
     try:
-        opts, args = getopt.getopt(sys.argv[1:],
-                                   "v",
-                                   ["verbose", "node"])
+        opts, args = getopt.getopt(sys.argv[1:], "v", ["verbose", "node"])
     except getopt.GetoptError as err:
         sys.exit(err)
 
--- a/utils/utils.py
+++ b/utils/utils.py
@@ -30,10 +30,11 @@ from consul import Consul, ConsulException
 from requests.exceptions import RequestException
 from urllib3.exceptions import HTTPError
 from hax.exception import HAConsistencyException
+from hax.util import ConsulUtil
 
 Process = NamedTuple('Process', [('node', str), ('consul_name', str),
                                  ('systemd_name', str), ('fidk', int),
-                                 ('status', str)])
+                                 ('status', str), ('is_local', bool)])
 shutdown_sequence = ('s3service', 'ios', 'confd', 'hax', 'consul')
 
 
@@ -74,9 +75,13 @@ def processes_node(cns: Consul, node_name: str) -> Dict[str, List[Process]]:
     """Processes grouped by Consul service name."""
     try:
         processes: Dict[str, List[Process]] = {}
+        cns_util = ConsulUtil(raw_client=cns)
+        is_local = node_name == cns_util.get_local_nodename()
+
         for node in cns.catalog.nodes()[1]:
             if node_name != node['Node']:
                 continue
+
             for svc in cns.health.node(node['Node'])[1]:
                 svc_name = svc['ServiceName']
                 if svc_name:
@@ -86,6 +91,7 @@ def processes_node(cns: Consul, node_name: str) -> Dict[str, List[Process]]:
                                 consul_name=svc_name,
                                 systemd_name=get_systemd_name(fidk, svc_name),
                                 fidk=fidk,
+                                is_local=is_local,
                                 status=svc['Status']))
             consul_status = 'passing' if consul_is_active_at(node['Node']) \
                 else 'offline'
@@ -94,6 +100,7 @@ def processes_node(cns: Consul, node_name: str) -> Dict[str, List[Process]]:
                         consul_name='consul',
                         systemd_name='hare-consul-agent',
                         fidk=0,
+                        is_local=is_local,
                         status=consul_status))
         return processes
     except (ConsulException, HTTPError, RequestException) as e:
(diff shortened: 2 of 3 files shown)


utils/utils.py
```diff
@@ -194,7 +206,11 @@ class QuitMessage:
 
 class Worker(threading.Thread):
     def __init__(self, queue: Queue):
-        super().__init__(target=self._do_work, args=(queue, ))
+        # [KN] We mark the thread as daemonic to make sure it will
```
suvratjoshi, line 209: I am assuming that these are initials of name.
Sorry but I am not  aware of. Do we generally enter our initials of name in code in this project?

knekrasov: I prefer to leave my initials in the comments that I can consider as important. But I can remove them if they look like a noise.

suvratjoshi: I don't see a problem. But generally I have seen hard restrictions in many projects related to this, so was just curious.

suvratjoshi: Ok to keep initials I think

utils/hare-shutdown
```diff
@@ -38,25 +39,38 @@ def check_cluster():
...
     for svc in utils.shutdown_sequence:
-        if svc in processes:
+        if svc not in processes:
+            continue
+        if svc != 'consul':
             utils.stop_parallel([proc for proc in processes[svc]])
+        else:
+            # Consul process at the current node must stopped the last.
+            # Reason: if something goes wrong and the user will want to
+            # shutdown the clsuter again, local Consul must be available.
```
suvratjoshi, line 67: cluster (just a typo)

suvratjoshi:
Hi @knekrasov 

Q1. On my VM based 4 node cluster setup (without your code patch) I tried hctl shutdown command and tried to kill / stop this shutdown process using ctrl+c, 2 times as shown below. Yes, it didn't get killed and continued to work and finally the hctl shutdown ended with correct shutdown. Now if ctrl+c works as expected and hctl shutdown script stops immediately, then in that case, half of the cluster processes would have already  been stopped and half of them would be still alive. What will be hctl status in that case? Can we say such cluster is still alive and working OR will it say no cluster running.
In case more than half of the processes were already stopped till the time we press ctrl+c, then if cluster is not in workable condition, do we need to put something additional in this patch to identify how useful cluster is and if not then, sending some indication to user (may be in output of hctl status), to use hctl shutdown command, to make shutdown process complete?

```
[root@ssc-vm-2183 535027]# hctl shutdown
Stopping m0d@0x7200000000000001:0xc (ios) at ssc-vm-2182.colo.seagate.com...
Stopping m0d@0x7200000000000001:0x26 (ios) at ssc-vm-2183.colo.seagate.com...
Stopping m0d@0x7200000000000001:0x40 (ios) at ssc-vm-2236.colo.seagate.com...
Stopping m0d@0x7200000000000001:0x60 (ios) at ssc-vm-2237.colo.seagate.com...
Stopped m0d@0x7200000000000001:0x26 (ios) at ssc-vm-2183.colo.seagate.com
Stopped m0d@0x7200000000000001:0x40 (ios) at ssc-vm-2236.colo.seagate.com
Stopped m0d@0x7200000000000001:0xc (ios) at ssc-vm-2182.colo.seagate.com
Stopped m0d@0x7200000000000001:0x60 (ios) at ssc-vm-2237.colo.seagate.com
Stopping m0d@0x7200000000000001:0x9 (confd) at ssc-vm-2182.colo.seagate.com...
Stopping m0d@0x7200000000000001:0x23 (confd) at ssc-vm-2183.colo.seagate.com...
Stopping m0d@0x7200000000000001:0x3d (confd) at ssc-vm-2236.colo.seagate.com...
Stopping m0d@0x7200000000000001:0x5d (confd) at ssc-vm-2237.colo.seagate.com...
^CKilled by signal 2.
Killed by signal 2.
Killed by signal 2.
**ERROR** Failed to stop m0d@0x7200000000000001:0x9 (confd) at ssc-vm-2182.colo.seagate.com
**ERROR** Failed to stop m0d@0x7200000000000001:0x3d (confd) at ssc-vm-2236.colo.seagate.com
**ERROR** Failed to stop m0d@0x7200000000000001:0x5d (confd) at ssc-vm-2237.colo.seagate.com
**ERROR** Failed to stop m0d@0x7200000000000001:0x23 (confd) at ssc-vm-2183.colo.seagate.com
Stopping hare-hax at ssc-vm-2182.colo.seagate.com...
Stopping hare-hax at ssc-vm-2183.colo.seagate.com...
Stopping hare-hax at ssc-vm-2236.colo.seagate.com...
Stopping hare-hax at ssc-vm-2237.colo.seagate.com...
^CKilled by signal 2.
Killed by signal 2.
Killed by signal 2.
**ERROR** Failed to stop hare-hax at ssc-vm-2182.colo.seagate.com
**ERROR** Failed to stop hare-hax at ssc-vm-2237.colo.seagate.com
**ERROR** Failed to stop hare-hax at ssc-vm-2236.colo.seagate.com
**ERROR** Failed to stop hare-hax at ssc-vm-2183.colo.seagate.com
Stopping hare-consul-agent at ssc-vm-2182.colo.seagate.com...
Stopping hare-consul-agent at ssc-vm-2183.colo.seagate.com...
Stopping hare-consul-agent at ssc-vm-2236.colo.seagate.com...
Stopping hare-consul-agent at ssc-vm-2237.colo.seagate.com...
Stopped hare-consul-agent at ssc-vm-2183.colo.seagate.com
Stopped hare-consul-agent at ssc-vm-2236.colo.seagate.com
Stopped hare-consul-agent at ssc-vm-2237.colo.seagate.com
Stopped hare-consul-agent at ssc-vm-2182.colo.seagate.com
Shutting down RC Leader at ssc-vm-2182.colo.seagate.com...
OK
[root@ssc-vm-2183 535027]#
```

suvratjoshi:
Q2 : This solution will be VM specific as on hardware pacemaker will be taking care. Is that correct?

knekrasov:
Hello @suvratjoshi 

You wrote:

> Q1. On my VM based 4 node cluster setup (without your code patch) I tried hctl shutdown command and tried to kill / stop this shutdown process using ctrl+c, 2 times as shown below. Yes, it didn't get killed and continued to work and finally the hctl shutdown ended with correct shutdown (agree that sometimes it will be erroneous).

The problem was as follows. Because of incorrect Ctrl-C handling, it was possible that Consul agent is stopped although some 
m0d or s3servers were not. Example: hare-shutdown was stopping m0d process, the user presses Ctrl-C, as a result shutting down of m0d gets canceled BUT hare-shutdown script doesn't finish and thus starts shutting down the next processes in its "shutdown sequence" list. The critical thing is that when consul agent is down, the user will not be able to shutdown the cluster with this script. Even if there are some m0d processes left.

> Now if ctrl+c works as expected and hctl shutdown script stops immediately, then in that case, half of the cluster processes would have already been stopped and half of them would be still alive. What will be hctl status in that case? Can we say such cluster is still alive and working OR will it say no cluster running.

`hctl status` will show the cluster as running (although some of the processes will be shown as stopped or in unknown state). Do you see a problem with it?

> In case more than half of the processes were already stopped till the time we press ctrl+c, then if cluster is not in workable condition, do we need to put something additional in this patch to identify how useful cluster is and if not then, sending some indication to user (may be in output of hctl status), to use hctl shutdown command, to make shutdown process complete?

Unfortunately, there is no formal definition of "not in workable state". What is important is that:
1. This situation is not something normal. But it is the user who has canceled `hctl shutdown`. 
2. So we presume that the user knows what the script does and thus the user knows other commands like `hctl status`. 
3. The latter will tell explicitly which processes are out of order.

Do you have ideas what else can be improved?


> Q2 : This solution will be VM specific as on hardware pacemaker will be taking care. Is that correct?

It's not a question of VM/non-VM deployment. It depends on whether we have Pacemaker or not. When the deployment assumes Pacemaker, nobody will need to invoke `hctl shutdown` at all. But there are deployments (like sage) without Pacemaker. And they are not VMs. AFAIK.

suvratjoshi:
```
hctl status will show the cluster as running (although some of the processes will be shown as stopped or in unknown state). Do you see a problem with it?
```
```
This situation is not something normal. But it is the user who has canceled hctl shutdown
```
I was thinking why somebody will kill the shutdown process. Generally I kill shutdown process only when I feel that its hanged since a very long and probably not responding. In that case your patch will really help now to ensure that shutdown command will work again.
Just as an improvement thing :
Cluster will not be usable mostly after ctrl+c, if half of processes running and half of them not running. In case somebody is not aware of looking at status of all services/processes in hctl status, then "cluster is running" message is misleading when actually cluster is not useful probably. After seeing this message of "running" he/she (or somebody else who is also using the same system in sharing) can try IO's or something.
```
Do you have ideas what else can be improved?
```
Solution-1:
So to avoid above said misleading message situation, this may be simplest solution (only if it is easily possible).
When such killing of shutdown process event occurs, then along with your patch, additionally a flag can be maintained to indicate that such an attempt was done. After this, when hctl status command is given, we can put a message in hctl status to indicate clearly that, use hctl shutdown command to finish the shutdown process gracefully and bootstrap again to start using the cluster.

OR Solution-2 (just a thought):
If SIGKILL occurs we can try to kill processes (even if required in some un-graceful way), as actually stucking of shutdown, can be the most probable reason, which has encouraged the user to use ctrl+c.
This will ensure complete shutdown of cluster, but actually it is opposite to what user is trying to do.

mssawant:
@suvratjoshi, @knekrasov, Consider not an explicit `ctrl+c` but a node reboot case. Say the node from which `hctl shutdown` was triggered got rebooted before shutting down all the nodes. Now motr or hax process may still be running on other nodes.
Tried this case and it is good that we can still trigger `hctl shutdown` from other nodes.
But, if a consul process is not running on the node then that thinks that cluster is not running while the process still can be running on other nodes. Due to this user may believe that cluster is not running and try to bootstrap the cluster. Bootstrap may hang or fail in this case, I have seen this and had to explicitly go and check on the nodes if the services are active and kill them.
I think it should be possible to detect such a scenario and report or cleanup during shutdown or bootstrap.

Not sure the best way to handle this, but one possible solution could be to invoke `hctl shutdown --node` on all the nodes without checking if consul is running locally or not. But we need to fetch nodes information first from somewhere and that could be tricky, it is possible to retrieve this information without consul from the CDF or earlier generated configuration files. 
Anyway this is not very trivial, for now we can assume that cortx-ha will take care of such situation but Hare needs to be independent. Let's do this as a separate task. Present patch looks very good.

mssawant (approved):

suvratjoshi (approved):

utils/hare-shutdown
knekrasov: Done

Issue #1482, opened 2021-02-02
[Mini Provisioner] Arguments like '--config' or '--init' should be just 'config' and 'init' respectively


knekrasov:
In current implementation the hare_setup CLI looks as follows:

```yaml
hare:
  post_install:
    script: /opt/seagate/cortx/hare/bin/hare_setup
    args: --report-unavailable-features --post_install
  init:
    script: /opt/seagate/cortx/hare/bin/hare_setup
    args: --init
  config:
    script: /opt/seagate/cortx/hare/bin/hare_setup
    args: --config $URL --filename '/var/lib/hare/cluster.yaml'
  test:
    script: /opt/seagate/cortx/hare/bin/hare_setup
    args: --test
  support_bundle:
    script: /opt/seagate/cortx/hare/bin/hare_setup
    args: --support_bundle
  reset:
    script: /opt/seagate/cortx/hare/bin/hare_setup
    args: --reset
  cleanup:
    script: /opt/seagate/cortx/hare/bin/hare_setup
    args: --cleanup
```

Things that are supposed to be subcommands look like flags. More than that, it doesn't fully correspond to [Mini-Provisioner Specifications](https://seagatetechnology.sharepoint.com/:w:/r/sites/gteamdrv1/tdrive1224/_layouts/15/Doc.aspx?sourcedoc=%7B41048EAD-3E48-4163-B9CC-69A04FE7E67B%7D&file=Mini-Provisioner%20Specifications.docx&nav=eyJjIjo2NDgwNzEzNDB9&action=default&mobileredirect=true) document and thus should be changed.

Switching this CLI to a concept of sub-commands can simplify processing of some optional parameters (like `--filename` - to specify CDF file for `init` command).

Suggested approaches:
1. [Subparsers](https://docs.python.org/3/library/argparse.html#sub-commands) in argparse (this is the library we use nowadays). Maybe quite wordy, though.
2. Switch to [Click library](https://click.palletsprojects.com/en/7.x/commands/) and use their multicommands.

SwapnilGaonkar7:
This issue is addressed under task https://jts.seagate.com/browse/EOS-17172
PR : https://github.com/Seagate/cortx-hare/pull/1504

PR #1504, opened 2021-02-15
Problem: Hare miniprovisioner init, test section uses hardcoded path for CDF file, also arguments like '--config' or '--init' should be just 'config' and 'init' respectively

SwapnilGaonkar7:
I have addressed https://github.com/Seagate/cortx-hare/issues/1482 and https://jts.seagate.com/browse/EOS-17172
Solution: Support added for init and test to take cdf file path as input and modified code to use sub-commands to handle miniprovisioner related commands

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -29,6 +29,7 @@ import shutil
 import subprocess
 import sys
 from typing import Dict, List
+
 import yaml
 from cortx.utils.product_features import unsupported_features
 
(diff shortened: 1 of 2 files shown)


SwapnilGaonkar7:
@knekrasov @mssawant Please review

provisioning/miniprov/hare_mp/main.py
```diff
@@ -343,6 +335,13 @@ def main():
         filename = parsed.filename[0] or '/var/lib/hare/cluster.yaml'
         save(filename, generate_cdf(url))
 
+    if parsed.init:
```
mssawant, line 338: I think we can avoid these multiple if loops by setting `nargs` to a non-zero value, can you try https://docs.python.org/3/library/argparse.html?
This automatically adds a check too on the args.

knekrasov: I would suggest to address #1482 together with this PR.

`init` or `config` would be a command while `--filename` is an option to a command. Otherwise we have cross-dependencies between options (`--init` handler is required to look for `--filename`) which doesn't look good and leads to over-complication in the code

mssawant: Yep, I agree with @knekrasov.

SwapnilGaonkar7:
@mssawant @knekrasov I have done code changes related to #1482.
Please review.

provisioning/setup.yaml
```diff
@@ -1,22 +1,22 @@
...
   init:
-    cmd: /opt/seagate/cortx/hare/bin/hare_setup
-    args: --init
+    cmd: /opt/seagate/cortx/hare/bin/hare_setup init
+    args:
   config:
-    cmd: /opt/seagate/cortx/hare/bin/hare_setup
-    args: --config $URL --filename '/var/lib/hare/cluster.yaml'
+    cmd: /opt/seagate/cortx/hare/bin/hare_setup config
+    args: --config $URL --filepath '/var/lib/hare/cluster.yaml'
```
knekrasov, line 10: Suggestion: `--file` or `--filename` instead of `--filepath`.

Example - docker-compose CLI:
```
Usage:
  docker-compose [-f <arg>...] [options] [COMMAND] [ARGS...]
  docker-compose -h|--help

Options:
  -f, --file FILE             Specify an alternate compose file
                              (default: docker-compose.yml)
```

provisioning/setup.yaml
```diff
@@ -1,22 +1,22 @@
...
   test:
-    cmd: /opt/seagate/cortx/hare/bin/hare_setup
-    args: --test
+    cmd: /opt/seagate/cortx/hare/bin/hare_setup test
```
knekrasov, line 12: Are you sure that `test` should go to `cmd`, not to `args`?

SwapnilGaonkar7: @knekrasov  I think it should go to cmd. Please see below from Mini-Provisioner spec doc
The setup.yaml specified by any component exhibits below format: 

```
<component>:   

  post_install:  				# Example: motr_setup post_install --config $URL 

    cmd: <component>_setup post_install	# CI setup command implemented by Component 

    args: --config <URL>			# ConfStore Config URL provided by Provisioner 
```

provisioning/miniprov/hare_mp/main.py
```diff
@@ -288,60 +277,83 @@ def save(filename: str, contents: str) -> None:
...
     parsed = p.parse_args(sys.argv[1:])
-    if parsed.config_url:
-        # --config also depends on optional parameter --filename,
-        # that is why the approach with putting the whole confi
-        # logic into the ConfigAction class will not work.
-        url = parsed.config_url[0]
-        filename = parsed.filename[0] or '/var/lib/hare/cluster.yaml'
-        save(filename, generate_cdf(url))
+
+    if parsed.func:
```
mssawant, line 355: Is there a case where `parsed.func` is not set? 
If yes, then what is the expected behavior? is it okay to exit silently? I think this means that the subcommand is not supported, shouldn't we throw and exception or log and error message in this case?

If no, then we need not have an `if` loop.

provisioning/setup.yaml
SwapnilGaonkar7: Done. I changed it to `--file`.
One question though. Isn't `--filename` sounds misleading considering that it is actually path not name.

knekrasov: Perhaps it was not the best choice from my side. Up to this point I used to think that filename can contain path to the file (either relative or absolute). But it seems like it [was incorrect](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_170).

provisioning/miniprov/hare_mp/main.py

SwapnilGaonkar7:
@mssawant @knekrasov Please have a look

knekrasov (approved):
I took the liberty to add one commit to your PR to fix formatting.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -288,60 +279,83 @@ def save(filename: str, contents: str) -> None:
...
-    if parsed.config_url:
-        # --config also depends on optional parameter --filename,
-        # that is why the approach with putting the whole confi
-        # logic into the ConfigAction class will not work.
-        url = parsed.config_url[0]
-        filename = parsed.filename[0] or '/var/lib/hare/cluster.yaml'
-        save(filename, generate_cdf(url))
+
+    if not hasattr(parsed, 'func'):
+        logging.error('Error: No valid command passed. Please check "--help"')
```
mssawant, line 355: [suggestion]: `logging.error('Error: Invalid command, please check hare_setup --help')`. 
Also it would be good if we can simply print help.

mssawant:
Just a minor comment, no further review is required. Kindly rebase and resolve the conflicts then we can merge.
Thank you.

mssawant (approved):

SwapnilGaonkar7:
@mssawant I have done rebase.

PR #1505, opened 2021-02-16
Problem: hax returns invalid RM fid post RC leader re-election

mssawant:
Hax fetches and saves RM fid when hax process starts. Hare elects an RC
leader from the given Consul server nodes which also run a motr
Confd process. Every confd runs a RM service, hax returns the fid of
motr RM service cooresponding to the confd running on the Hare RC leader
node. However, it is possible that the RC leader may change due existing RC
leader failure, in this case hax still returns the saved RM fid in response
to the new entrypoint requests from the motr processes. This may lead to
inconsistencies and failure of RM requests.

Solution:
Do not save the RM fid but fetch it while responding to the entrypoint
requests.

### DIFF
--- a/hax/hax/motr/__init__.py
+++ b/hax/hax/motr/__init__.py
@@ -30,7 +30,7 @@ from hax.types import (ConfHaProcess, Fid, FidStruct, FsStats, HaNote,
                        HaNoteStruct, HAState, MessageId, ObjT, Profile,
                        ReprebStatus, ServiceHealth, StobIoqError,
                        m0HaProcessEvent, m0HaProcessType)
-from hax.util import ConsulUtil
+from hax.util import ConsulUtil, repeat_if_fails
 
 LOG = logging.getLogger('hax')
 
@@ -49,7 +49,6 @@ class Motr:
     def __init__(self,
                  ffi: HaxFFI,
                  queue,
-                 rm_fid: Fid,
                  herald: DeliveryHerald,
                  consul_util: ConsulUtil,
                  node_uuid: str = ''):
@@ -58,7 +57,6 @@ class Motr:
         # hax.c function
         self._ha_ctx = self._ffi.init_motr_api(self, make_c_str(node_uuid))
         self.queue = queue
-        self.rm_fid = rm_fid
         self.herald = herald
         self.consul_util = consul_util
         self.spiel_ready = False
@@ -69,13 +67,19 @@ class Motr:
             raise RuntimeError('Cannot initialize Motr API')
 
     def start(self, rpc_endpoint: str, process: Fid, ha_service: Fid,
-              rm_service: Fid, profile: Profile):
+              profile: Profile):
         LOG.debug('Starting m0_halon_interface')
         self._process_fid = process
         self._profile = profile
+
+        @repeat_if_fails()
+        def _get_rm_fid() -> Fid:
+            return self.consul_util.get_rm_fid()
+
+        rm_fid = _get_rm_fid()
         result = self._ffi.start(self._ha_ctx, make_c_str(rpc_endpoint),
                                  process.to_c(), ha_service.to_c(),
-                                 rm_service.to_c())
+                                 rm_fid.to_c())
         if result:
             LOG.error(
                 'Cannot start Motr API. m0_halon_interface::start'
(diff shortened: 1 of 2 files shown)


hax/hax/motr/__init__.py
```diff
@@ -164,17 +168,19 @@ def send_entrypoint_request_reply(self, message: EntrypointRequest):
...
+            confds = util.get_confd_list()
+            rm_fid = util.get_rm_fid()
         except Exception:
             LOG.exception('Failed to get the data from Consul.'
                           ' Replying with EAGAIN error code.')
             self._ffi.entrypoint_reply(reply_context, req_id.to_c(), EAGAIN, 0,
                                        make_array(FidStruct, []),
                                        make_array(c.c_char_p, []), 0,
-                                       self.rm_fid.to_c(), None)
+                                       Fid(0, 0).to_c(),
```
knekrasov, line 182: Why is it zeroed? Is it expected by Motr in case of  rc=EAGAIN ?

hax/hax/util.py
```diff
@@ -281,6 +281,7 @@ def get_ha_fid(self) -> Fid:
         svc = self._local_service_by_name('hax')
         return mk_fid(ObjT.SERVICE, int(svc['ServiceID']) + 1)
 
+    @repeat_if_fails()
```
knekrasov, line 284: Probably it is not that required now. get_rm_fid() is invoked in while preparing the entrypoint reply. If an exception (even HAConsistencyException) appears, EAGAIN will be returned to Motr.

`qconsumer` thread will not be blocked and perhaps will manage to do something meaningful instead of blocking on `@repeat_if_fails()`

hax/hax/motr/__init__.py
mssawant: Its a error case so, either we can set rm_fid to NULL in which case motr anyway sets rm_fid to (0, 0) or we return (0,0) (as in this case).
```
void m0_halon_interface_entrypoint_reply(
                struct m0_halon_interface  *hi,
                const struct m0_uint128    *req_id,
                int                         rc,
                uint32_t                    confd_nr,
                const struct m0_fid        *confd_fid_data,
                const char                **confd_eps_data,
                uint32_t                    confd_quorum,
                const struct m0_fid        *rm_fid,
                const char                 *rm_eps)
{
        struct m0_ha_entrypoint_rep rep;

        M0_ENTRY("hi=%p req_id="U128X_F" rc=%d confd_nr=%"PRIu32" "
                 "confd_quorum=%"PRIu32" rm_fid="FID_F" rm_eps=%s",
                 hi, U128_P(req_id), rc, confd_nr, confd_quorum,
                 FID_P(rm_fid == NULL ? &M0_FID0 : rm_fid), rm_eps);

        if (hi->hif_internal->hii_cfg.hic_log_entrypoint) {
                M0_LOG(M0_WARN, "req_id="U128X_F" rc=%d confd_nr=%"PRIu32" "
                 "confd_quorum=%"PRIu32" rm_fid="FID_F" rm_eps=%s",
                 U128_P(req_id), rc, confd_nr, confd_quorum,
                 FID_P(rm_fid == NULL ? &M0_FID0 : rm_fid), rm_eps);
        }
        rep = (struct m0_ha_entrypoint_rep){
                .hae_quorum        = confd_quorum,
                .hae_confd_fids    = {
                        .af_count = confd_nr,
                        .af_elems = (struct m0_fid *)confd_fid_data,
                },
                .hae_confd_eps     = confd_eps_data,
                .hae_active_rm_fid = rm_fid == NULL ? M0_FID0 : *rm_fid,
                .hae_active_rm_ep  = (char *)rm_eps,
                .hae_control       = rc == 0 ? M0_HA_ENTRYPOINT_CONSUME :
                                               M0_HA_ENTRYPOINT_QUERY,
        };
        m0_ha_entrypoint_reply(&hi->hif_internal->hii_ha, req_id, &rep, NULL);
        M0_LEAVE();
```
As we are returning -EAGAIN, motr will request entrypoint again. So we are good.

hax/hax/util.py
mssawant: hmm.. yes that is true for entrypoint reply but get_rm_fid() is also invoked from motr::start() which may fail but this was true even earlier when we were fetching rm_fid from hax.py but _get_motr_fids() was tagged with `@repeat_if_fails()` in that case.
Have defined a nested function in motr::start() and tagged it with `repeat_if_fails()` and removed `repeat_if_fails()` from utils::get_rm_fid().

knekrasov (approved):

hax/hax/util.py
knekrasov: Yeah, the more we use `@repeat_if_fails()`, the more tricky it is to understand whether repeat will happen or when it will not (especially when analyzing the long call hierarchies). At this point the idea of checked exceptions (Java) and explicit types like `IO a` (Haskell) look not that bad 😀

suvratjoshi:
Would be nice if JIRA number can be mentioned here in the title of PR.

suvratjoshi (approved):

PR #1511, opened 2021-02-18
Problem: Hare CI I/O test with m0crate is aborted

parikshit-dharmale:
hctl bootstrap and hctl shutdown are taking longer time
to execute causing jenkins timeout to trigger and aborting
the I/O test with m0crate.

Solution: Increase the jenkins timeout for I/O test with m0crate.

Signed-off-by: Parikshit Dharmale <parikshit.dharmale@seagate.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -185,7 +185,7 @@ pipeline {
 
         stage('I/O test with m0crate') {
             options {
-                timeout(time: 10, unit: 'MINUTES')
+                timeout(time: 15, unit: 'MINUTES')
             }
             steps {
                 script {


knekrasov:
@parikshit-dharmale I wonder why Hare bootstrap and shutting down takes more time than it used to. Is it because the VM is weak?

I would approve this patch just to unblock our CI but probably we need to investigate the reasons. It may happen that this is a symptom of some bigger issue. Will you create a ticket for research in JIRA @parikshit-dharmale ?

cc @mssawant, @s-arya

knekrasov (approved):

PR #1515, opened 2021-02-19
Problem: hare_setup fails for --test option

SwapnilGaonkar7:
Solution: Fixed wrong closing bracket

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -272,7 +272,7 @@ def check_cluster_status():
                     return -1
                 ios_cnt += 1
         if (m0client_s3_cnt > 0 and len(node_info_dict[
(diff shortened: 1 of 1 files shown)


mssawant (approved):

provisioning/miniprov/hare_mp/main.py
```diff
@@ -272,7 +272,7 @@ def check_cluster_status():
                     return -1
                 ios_cnt += 1
         if (m0client_s3_cnt > 0 and len(node_info_dict[
-                node['hostname']]['s3server'] != m0client_s3_cnt)):
+                node['hostname']]['s3server']) != m0client_s3_cnt):
```
knekrasov, line 275: @SwapnilGaonkar7 I think we need automated tests that would prevent the errors like this. Can you submit a GH issue for that?

knekrasov (approved):

PR #1524, opened 2021-03-02
Problem: hare doesn't bootstrap in devinstall mode

knekrasov:
JIRA: [EOS-18433](https://jts.seagate.com/browse/EOS-18433)

Problem causes:
1. In devinstall mode, the packages must not be installed to
/opt/seagate/cortx/hare (via `pip ... --prefix <folder>`). We must use
local virtualenv folder instead and create symlinks to executables only.
2. hare_setup is not installed during devinstall because of a mistype in
Makefile.

Solution: fix Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -263,6 +263,7 @@ install-dirs:
 
 .PHONY: install-cfgen
 install-cfgen: CFGEN_INSTALL_CMD = install
+install-cfgen: CFGEN_PIP_CMD = $(PIP) install --ignore-installed --prefix $(DESTDIR)/$(PREFIX) -r $<
 install-cfgen: $(CFGEN_EXE) install-cfgen-deps unpack-dhall-bin unpack-dhall-prelude
 	@$(call _info,Installing cfgen/dhall configs)
 	@install --verbose --directory $(CFGEN_SHARE)
@@ -274,7 +275,7 @@ install-cfgen: $(CFGEN_EXE) install-cfgen-deps unpack-dhall-bin unpack-dhall-pre
 .PHONY: install-cfgen-deps
 install-cfgen-deps: cfgen/requirements.txt $(PY_VENV_DIR)
 	@$(call _info,Installing cfgen dependencies)
-	@$(PIP) install --ignore-installed --prefix $(DESTDIR)/$(PREFIX) -r $<
+	@$(CFGEN_PIP_CMD)
 
 $(CFGEN_EXE): cfgen/cfgen
 	@$(call _info,Installing cfgen)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -392,20 +392,29 @@ devinstall-systemd: $(wildcard systemd/*)
...
+# Don't specify --prefix $(DESTDIR)/$(PREFIX) since we do want to use our
+# virtualenv folder instead. Generated executables will point to python binary
+# from our virtualenv via shebang. Those executables will be put to /opt/seagate
+# as symlinks. So it is critical to have virtualenv folder populated.
+devinstall-hax: HAX_INSTALL_CMD = $(SETUP_PY) develop
 devinstall-hax: export PYTHONPATH = $(DESTDIR)/$(PREFIX)/lib/python3.$(PY3_VERSION_MINOR)/site-packages
 devinstall-hax: hax/requirements.txt $(HAX_EGG_LINK)
 	@$(call _info,Installing hax development dependencies)
-	@$(PIP) install --ignore-installed --prefix $(DESTDIR)/$(PREFIX) \
+	@$(PIP) install --ignore-installed \
```
mssawant, line 403: Okay, so earlier we were installing python dependencies in `/opt/seagate/cortx/hare/lib..` but now we are installing them in virtual env and linking the binaries to it, looks okay to me.

mssawant (approved):
Did not test but looks okay.

knekrasov:
Tested locally: hctl bootstrap works fine. 

The problem is that without having activated the virtualenv hctl status and hctl shutdown will fail like this:
```
hctl status
  File "/opt/seagate/cortx/hare/libexec/hare-status", line 26, in <module>
    import simplejson
ModuleNotFoundError: No module named 'simplejson'
```

So it seems like the patch is not complete, I need to play with symlinks a bit more to make the tools work seamlessly.

Makefile
```diff
@@ -392,20 +392,29 @@ devinstall-systemd: $(wildcard systemd/*)
...
+devinstall-miniprov: MP_INSTALL_CMD = $(SETUP_PY) develop
 devinstall-miniprov: export PYTHONPATH = $(DESTDIR)/$(PREFIX)/lib/python3.$(PY3_VERSION_MINOR)/site-packages
-devinstall-miniprov: provisioning/miniprov/requirements.txt $(HAX_EGG_LINK)
+devinstall-miniprov: provisioning/miniprov/requirements.txt $(MP_EGG_LINK)
 	@$(call _info,Installing miniprov development dependencies)
-	@$(PIP) install --ignore-installed --prefix $(DESTDIR)/$(PREFIX) \
+	@$(PIP) install --ignore-installed \
 					--requirement <(sed -ne '/^#:runtime-requirements:/,$$p' $<)
+	@$(call _info,Creating symlinks for mini-provisioner executables)
+	@ln -v -sf $(PY_VENV_DIR)/bin/hare_setup $(DESTDIR)/$(PREFIX)/bin
```
mssawant, line 417: #1529 creates link for `install-miniprov` target as well, do we need to incorporate that here? or we can merge #1529.

knekrasov: I'm not sure that the fix in #1529 is correct. Please see my comment there.

SwapnilGaonkar7 (approved):
Looks good

PR #1527, opened 2021-03-03
Problem: mini provisioner doesn't work at stable

knekrasov:
It turned out that setup.yaml file, --config option is mandatory for all commands otherwise provisioner code will fail.

Solution: Add --config option for all commands in hare setup.yaml


s-arya (approved):

SwapnilGaonkar7 (approved):
Looks good

PR #1529, opened 2021-03-04
EOS-18353: Package imports issue in hare setup

supritshinde:
Description: After 'make install' we see that hare_setup cannot find hare_mp package and its modules

Solution: We need to install the hare mini provisioner dependencies in py3venv virtual environment. Instead of /opt/seagate/cortx/hare which alias PREFIX in Makefile. Make changes to Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -306,8 +306,10 @@ $(HAX_EGG_LINK) $(HAX_EXE): $(HAX_WHL)
 	@cd hax && $(HAX_INSTALL_CMD)
 
 .PHONY: install-miniprov
-install-miniprov: MP_INSTALL_CMD = $(PIP) install --ignore-installed --prefix $(DESTDIR)/$(PREFIX) $(MP_WHL)
+install-miniprov: MP_INSTALL_CMD = $(PIP) install --ignore-installed $(MP_WHL)
 install-miniprov: $(MP_EXE)
+	@$(call _info,Symlink created for mini provisioner)
+	@ln -v -sf $(PY_VENV_DIR)/bin/hare_setup $(DESTDIR)/$(PREFIX)/bin
 
 $(MP_EGG_LINK) $(MP_EXE): $(MP_WHL)
 	@$(call _info,Installing miniprov with '$(MP_INSTALL_CMD)')


knekrasov:
The patch seems to be a subset of #1524. Not sure that we need to merge this one (unless it is a hotfix to unblock other teams, of course)

Makefile
```diff
@@ -306,8 +306,10 @@ $(HAX_EGG_LINK) $(HAX_EXE): $(HAX_WHL)
 	@cd hax && $(HAX_INSTALL_CMD)
 
 .PHONY: install-miniprov
-install-miniprov: MP_INSTALL_CMD = $(PIP) install --ignore-installed --prefix $(DESTDIR)/$(PREFIX) $(MP_WHL)
+install-miniprov: MP_INSTALL_CMD = $(PIP) install --ignore-installed $(MP_WHL)
```
mssawant, line 309: Looks like #1524 fixes it only for `devinstall-miniprov` target , we can merge this PR then as there is no overlapping, cc @knekrasov.

knekrasov: No, my #1524 also has that line fixed. If we merge this one, there will be merge conflict. Instead we can merge #1524 😉 
It is ready for review BTW.

knekrasov: Yes, correct. But please note that install target is used for rpm building. The dependencies of miniprov are not put into the prefix, they will not be packaged into the rpm. In other words, I'm afraid this line will break the RPM.

@supritshinde Did you try to build the RPM and install it afterwards? Did setup_hare work fine for you after that?

mssawant: Ah, you are right. @supritshinde please try, creating an RPM, installing it and invoking mini-provisioning commands.

supritshinde: Okay got it! So, I ran the following commands on two different machines where in first one, I have my edited Makefile and second one is unaltered.
- rm -rf .py3venv/
-  make clean
- make
- make rpm
- make install
- and then invoked '/opt/seagate/cortx/hare/bin/hare_setup --help'

>> Output of both the make rpm commands turns out to be the same: 

--> Preparing rpmbuild environment
‘cortx-hare-1.0.0.tar.gz’ -> ‘/root/rpmbuild/SOURCES/cortx-hare-1.0.0.tar.gz’
‘hare.spec’ -> ‘/root/rpmbuild/SPECS/hare.spec’
make[1]: Leaving directory `/root/hare'
make[1]: Entering directory `/root/hare'
--> Building rpm packages
error: line 51: Unknown tag: package package is not installed
make[1]: *** [__rpm] Error 1
make[1]: Leaving directory `/root/hare'
make: *** [rpm] Error 2

supritshinde: And @knekrasov Yeah the hare_setup seems fine. Verified by invoking mini provisioner command.

supritshinde: Still I need to verify one thing. I guess something is wrong in my setup rather than those updated above.

knekrasov: I still have doubts that this thing with `pip ... --prefix` is correct. 
The fact is that at this moment miniprov's requirements.txt doesn't bring anything which is not specified already in hax's requirements.txt.

Probably that's the reason why this patch did not break the RPM. But as soon as we add anything new to this requirements.txt:
```
#:build-requirements:
flake8
mypy
pkgconfig
setuptools
wheel

#:runtime-requirements:
setuptools
dataclasses
```
then those changes will not be reflected in the RPM. 

Probably the better approach is as follows:
execute BOTH of the commands:
```
pip install --ignore-installed --prefix $(DESTDIR)/$(PREFIX) $(MP_WHL)
pip install --ignore-installed $(MP_WHL)
```

PR #1532, opened 2021-03-05
EOS-18485:All services not online while unboxing after max attempts

suvratjoshi:
Problem:
All the pcs resources are not online within time limit.
Hence Unboxing process not successful. Confd's entrypoint request
was not replied in time and it timed out.
Solution:
Current queue length in hax is 8, because of which if the requests
overflow the queue, they may get dropped or not serviced due to
which motr process may end up timing out. So increasing the queue
length from 8 to 8192.

### DIFF
--- a/hax/hax/hax.py
+++ b/hax/hax/hax.py
@@ -88,7 +88,7 @@ def main():
     # _run_qconsumer_thread function.
     #
     # [KN] Note: The server is launched in the main thread.
-    q = Queue(maxsize=32767)
+    q = Queue(maxsize=8191)
 
     util: ConsulUtil = ConsulUtil()
     cfg = _get_motr_fids(util)
--- a/hax/hax/hax.py
+++ b/hax/hax/hax.py
@@ -88,7 +88,7 @@ def main():
     # _run_qconsumer_thread function.
     #
     # [KN] Note: The server is launched in the main thread.
-    q = Queue(maxsize=8)
+    q = Queue(maxsize=32767)
 
     util: ConsulUtil = ConsulUtil()
     cfg = _get_motr_fids(util)
--- a/hax/hax/hax.py
+++ b/hax/hax/hax.py
@@ -88,7 +88,7 @@ def main():
     # _run_qconsumer_thread function.
     #
     # [KN] Note: The server is launched in the main thread.
-    q = Queue(maxsize=8191)
+    q = Queue(maxsize=8192)
 
     util: ConsulUtil = ConsulUtil()
     cfg = _get_motr_fids(util)


hax/hax/hax.py
```diff
@@ -88,7 +88,7 @@ def main():
     # _run_qconsumer_thread function.
     #
     # [KN] Note: The server is launched in the main thread.
-    q = Queue(maxsize=8)
+    q = Queue(maxsize=32767)
```
knekrasov, line 91: Looks like an overkill. I don't know what data structure is used under the hood of Queue, but it may happen that it's not an array but a linked list or something. So even maintenance of such a huge structure can bring additional performance penalties. That's why I'd recommend change sizes not that dramatically.

Why wouldn't it be 1K or 8K instead?

suvratjoshi: Thanks!
Changed to 8191 as per discussion today

hax/hax/hax.py
```diff
@@ -88,7 +88,7 @@ def main():
     # _run_qconsumer_thread function.
     #
     # [KN] Note: The server is launched in the main thread.
-    q = Queue(maxsize=8)
+    q = Queue(maxsize=8191)
```
mssawant, line 91: 1024 * 8 = 8192

hax/hax/hax.py
mssawant: Kindly squash the commits, just keep the latest one.

mssawant (approved):
Minor comments, kindly address them. I have no further comments.

knekrasov (approved):

ujjwalpl (approved):

knekrasov:
@mssawant, @NeeravChoudhari, @s-arya What are the latest thoughts on this patch? Don't we need it in `cortx-1.0` anymore?

NeeravChoudhari:
> @mssawant, @NeeravChoudhari, @s-arya What are the latest thoughts on this patch? Don't we need it in `cortx-1.0` anymore?

No, we have found the root cause and fixed it. This is not required in Cortx-1.0 and we should merge this in Main or Stable only if an issue is seen which requires this fix. For now, should be closed. @mssawant  Please confirm

suvratjoshi:
As it is discussed [here ](https://jts.seagate.com/browse/EOS-18485?focusedCommentId=2058619&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2058619)in this comment on the JIRA, we dont need this queue size related fix in the Cortx-1.0 branch.
So closing this PR.
Thanks!

PR #1537, opened 2021-03-09
EOS-18485: Not all the services are online during unboxing after max attempts

mssawant:
As part of fix (a354d38) to EOS-14988, Hare implemented a protocol fix with respect to motr communication. The fix introduced a problem of blocking motr thread in hax, that lead to delay in processing of motr events posted internally by Hare.
The pulled in commits from main mainly fix 2 problems,
1. Unblock in coming motr thread in hax corresponding to the entrypoint request.
2. Fix the wrong tag used in the halink message delivered call back in hax motr land.

The patch with these commits reduces the overall starting time of motr processes thus resolving the issues that are seen during cluster start/stop, standby/unstandby and boxing/unboxing.

Hare team has tested this patch by running at least 7 iterations of boxing/unboxing.

### DIFF
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -21,14 +21,18 @@ import time
 from queue import Empty, Queue
 from typing import List
 
-from hax.message import (BroadcastHAStates, EntrypointRequest, HaNvecGetEvent,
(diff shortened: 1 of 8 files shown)


knekrasov (approved):
Looks good to me.

suvratjoshi (approved):

PR #1539, opened 2021-03-10
Problem: Hare mini provisioner sets default values for some keys in CDF

SwapnilGaonkar7:
Solution: Reading values for pools and profiles from ConfStore

UT: Tested config, init, test sections for singlenode, execution is completes successful.
Sample CDF:
pools:
- allowed_failures: null
  name: StorageSet-1
  data_units: 1
  parity_units: 0
  disk_refs:
  - path: /dev/sdc
    node: ssc-vm-c-1822.colo.seagate.com
  type: sns
profiles:
- pools:
  - StorageSet-1
  name: Profile_the_pool
nodes:
- m0_servers:
  - io_disks:
      data: []
      meta_data: /dev/vg_metadata_srvnode-1/lv_raw_metadata
    runs_confd: true
  - io_disks:
      data:
      - /dev/sdc
      meta_data: /dev/vg_metadata_srvnode-1/lv_raw_metadata
    runs_confd: null
  data_iface: eth1
  hostname: ssc-vm-c-1822.colo.seagate.com
  m0_clients:
    other: 3
    s3: 0
  data_iface_type: tcp

Snippet from sample confstore file:
"cluster": {
    "cluster_id": "92f444df-87cc-4137-b680-aab3b35d1695",

"92f444df-87cc-4137-b680-aab3b35d1695": {
      "site": {
        "storage_set_count": "1"
      },
      "storage_set1": {
        "name": "StorageSet-1",
        "server_nodes": ["srvnode-1"],
        "durability": {
          "data": "1",
          "parity": "0",
          "spare": "0"
        }
      }
    }

### DIFF
--- a/provisioning/miniprov/hare_mp/cdf.py
+++ b/provisioning/miniprov/hare_mp/cdf.py
@@ -6,7 +6,8 @@ from typing import Dict, List, Optional
 import pkg_resources
 
 from hare_mp.store import ValueProvider
-from hare_mp.types import DList, Maybe, NodeDesc, Protocol, Text
+from hare_mp.types import (DList, Maybe, DiskRef, PoolDesc, ProfileDesc,
+                           NodeDesc, ClusterDesc, Protocol, Text)
 
 DHALL_PATH = '/opt/seagate/cortx/hare/share/cfgen/dhall'
 DHALL_EXE = '/opt/seagate/cortx/hare/bin/dhall'
@@ -29,6 +30,15 @@ class CdfGenerator:
             'hare_mp', resource_path)
         return raw_content.decode('utf-8')
 
+    def _get_cluster_id(self) -> str:
+        conf = self.provider
+
+        # We will read 'cluster_id' of 1st 'machine_id' present in server_node
+        server_node = conf.get('server_node')
+        machine_id = list(server_node.keys())[0]
+        cluster_id = server_node[machine_id]['cluster_id']
+        return cluster_id
+
     def _create_node_descriptions(self) -> List[NodeDesc]:
         nodes: List[NodeDesc] = []
         conf = self.provider
@@ -37,11 +47,70 @@ class CdfGenerator:
             nodes.append(self._create_node(node))
         return nodes
 
+    def _create_pool_descriptions(self) -> List[PoolDesc]:
+        pools: List[PoolDesc] = []
+        conf = self.provider
+        cluster_id = self._get_cluster_id()
+        storage_set_count = int(
+            conf.get(f'cluster>{cluster_id}>site>storage_set_count'))
+
+        for x in range(storage_set_count):
+            storage_set_name = conf.get(
+                f'cluster>{cluster_id}>storage_set{x+1}>name')
+
+            data_devices_count: int = 0
+            for node in conf.get(
+                    f'cluster>{cluster_id}>storage_set{x+1}>server_nodes'):
+                data_devices_count += len(
+                    conf.get(f'cluster>{node}>storage>data_devices'))
+
+            data_units_count = int(conf.get(
+                f'cluster>{cluster_id}>storage_set{x+1}>durability>data'))
+            parity_units_count = int(conf.get(
+                f'cluster>{cluster_id}>storage_set{x+1}>durability>parity'))
+            spare_units_count = int(conf.get(
+                f'cluster>{cluster_id}>storage_set{x+1}>durability>spare'))
+
+            if (data_devices_count != 0
+                and not data_devices_count >=
+                    data_units_count + parity_units_count + spare_units_count):
+                raise RuntimeError('Invalid storage set configuration')
+
+            pools.append(PoolDesc(
+                name=Text(storage_set_name),
+                disk_refs=Maybe(DList([
+                    DiskRef(path=Text(device), node=Maybe(Text(node), 'Text'))
+                    for node in conf.get(
+                        f'cluster>{cluster_id}>storage_set{x+1}>server_nodes')
+                    for device in conf.get(
+                        f'cluster>{node}>storage>data_devices')
+                ], 'List DiskRef'), 'List DiskRef'),
+                data_units=data_units_count,
+                parity_units=parity_units_count))
+
+        return pools
+
+    def _create_profile_descriptions(
+        self, pool_desc: List[PoolDesc]
+    ) -> List[ProfileDesc]:
+        profiles: List[ProfileDesc] = []
+
+        profiles.append(ProfileDesc(
+            name=Text('Profile_the_pool'),
+            pools=DList([pool.name
+                         for pool in pool_desc
+                         ], 'List Text')))
+
+        return profiles
+
     def _get_cdf_dhall(self) -> str:
         dhall_path = self._get_dhall_path()
         nodes = self._create_node_descriptions()
+        pools = self._create_pool_descriptions()
+        profiles = self._create_profile_descriptions(pools)
 
-        params_text = str(nodes)
+        params_text = str(ClusterDesc(
+            node_info=nodes, pool_info=pools, profile_info=profiles))
         gencdf = Template(self._gencdf()).substitute(path=dhall_path,
                                                      params=params_text)
         return gencdf
@@ -68,7 +137,6 @@ class CdfGenerator:
         yaml_out, err = to_yaml.communicate(input=dhall_out)
         if to_yaml.returncode:
             raise RuntimeError(f'dhall-to-yaml binary failed: {err}')
-
         return yaml_out
 
     def _get_iface(self, nodename: str) -> str:
--- a/provisioning/miniprov/hare_mp/types.py
+++ b/provisioning/miniprov/hare_mp/types.py
@@ -12,7 +12,7 @@ class Maybe:
         if self.value is None:
             return f'None {self.comment}'
 
-        return f'Some {self.value}'
+        return f'Some ({self.value})'
 
 
 @dataclass
@@ -87,3 +87,30 @@ class NodeDesc(DhallTuple):
     io_disks: DList  # [str]
     meta_data: Text
     s3_instances: int
+
+
+@dataclass(repr=False)
+class DiskRef(DhallTuple):
+    path: Text
+    node: Maybe
+
+
+@dataclass(repr=False)
+class PoolDesc(DhallTuple):
+    name: Text
+    disk_refs: Maybe  # [DList]
+    data_units: int
+    parity_units: int
+
+
+@dataclass(repr=False)
+class ProfileDesc(DhallTuple):
+    name: Text
+    pools: DList  # [str]
+
+
+@dataclass(repr=False)
+class ClusterDesc(DhallTuple):
+    node_info: List[NodeDesc]
+    pool_info: List[PoolDesc]
+    profile_info: List[ProfileDesc]
--- a/provisioning/miniprov/test/test_cdf.py
+++ b/provisioning/miniprov/test/test_cdf.py
@@ -27,7 +27,7 @@ from unittest.mock import MagicMock, Mock
 from hare_mp.cdf import CdfGenerator
 from hare_mp.store import ValueProvider
 from hare_mp.types import (DisksDesc, DList, M0Clients, M0ServerDesc, Maybe,
-                           NodeDesc, Protocol, Text)
+                           NodeDesc, PoolDesc, DiskRef, Protocol, Text)
 
 
 class TestTypes(unittest.TestCase):
@@ -42,14 +42,24 @@ class TestTypes(unittest.TestCase):
     def test_maybe_none(self):
         val = Maybe(None, 'P')
         self.assertEqual('None P', str(val))
-        self.assertEqual('Some P.tcp', str(Maybe(Protocol.tcp, 'P')))
+        self.assertEqual('Some (P.tcp)', str(Maybe(Protocol.tcp, 'P')))
 
     def test_disks_empty(self):
         val = M0ServerDesc(runs_confd=Maybe(True, 'Bool'),
                            io_disks=DisksDesc(meta_data=Maybe(None, 'Text'),
                                               data=DList([], 'List Text')))
         self.assertEqual(
-            '{ runs_confd = Some True, io_disks = { meta_data = None Text, data = [] : List Text } }',
+            '{ runs_confd = Some (True), io_disks = { meta_data = None Text, data = [] : List Text } }',
+            str(val))
+
+    def test_pooldesc_empty(self):
+        val = PoolDesc(
+            name=Text('storage_set_name'),
+            disk_refs=Maybe(DList([], 'List DiskRef'), []),
+            data_units=0,
+            parity_units=0)
+        self.assertEqual(
+            '{ name = "storage_set_name", disk_refs = Some ([] : List DiskRef), data_units = 0, parity_units = 0 }',
             str(val))
 
     def test_m0server_with_disks(self):
@@ -59,7 +69,7 @@ class TestTypes(unittest.TestCase):
                 meta_data=Maybe(None, 'Text'),
                 data=DList([Text('/disk1'), Text('/disk2')], 'test')))
         self.assertEqual(
-            '{ runs_confd = Some True, io_disks = { meta_data = None Text, data = ["/disk1", "/disk2"] } }',
+            '{ runs_confd = Some (True), io_disks = { meta_data = None Text, data = ["/disk1", "/disk2"] } }',
             str(val))
 
 
@@ -72,6 +82,7 @@ class TestCDF(unittest.TestCase):
                 'cluster>server_nodes': {
                     "blah": "srvnode_1"
                 },
+                'cluster>cluster_id': '92f444df-87cc-4137-b680-aab3b35d1695',
                 'cluster>srvnode_1>hostname':
                 'myhost',
                 'cluster>srvnode_1>network>data>interface_type':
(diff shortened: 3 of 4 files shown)


mssawant:
@SwapnilGaonkar7, kindly `squash` or `fixup` the commits appropriately.

knekrasov:
@SwapnilGaonkar7 Good work, it seems like you got the idea of Dhall expressions!

One comment: you have added new Dhall types and new functionality but I don't see new tests added. Will you add them?

knekrasov:
@SwapnilGaonkar7 The code in cdf.py is not formatted (flake8 will complain) and there are also problems with the types:
```
mypy hare_mp
hare_mp/cdf.py:52: error: Argument "node" to "DiskRef" has incompatible type "Maybe"; expected "Text"
hare_mp/cdf.py:52: error: Argument 2 to "Maybe" has incompatible type "None"; expected "str"
hare_mp/cdf.py:55: error: Argument 2 to "Maybe" has incompatible type "List[<nothing>]"; expected "str"
hare_mp/cdf.py:78: error: Argument "node_info" to "ClusterDesc" has incompatible type "List[NodeDesc]"; expected "DList"
hare_mp/cdf.py:78: error: Argument "pool_info" to "ClusterDesc" has incompatible type "List[PoolDesc]"; expected "DList"
hare_mp/cdf.py:78: error: Argument "profile_info" to "ClusterDesc" has incompatible type "List[ProfileDesc]"; expected "DList"
Found 6 errors in 1 file (checked 6 source files)
```

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -37,11 +38,44 @@ def _create_node_descriptions(self) -> List[NodeDesc]:
...
+
+        for x in range(storage_set_count):
+            storage_set_name = conf.get(f'cluster>{cluster_id}>storage_set{x+1}>name')
+            pools.append(PoolDesc(
+                name=Text(storage_set_name),
+                disk_refs=Maybe(DList([
+                    DiskRef(path=Text(device), node=Maybe(Text(node),None))
+                    for node in conf.get(f'cluster>{cluster_id}>storage_set{x+1}>server_nodes')
+                    for device in conf.get(f'cluster>{node}>storage>data_devices')
+                ], 'List Text'), []),
```
knekrasov, line 55: Note that 'List Text' is a label that will be used as type annotation in Dhall in case if the list becomes empty (Dhall enforces the writer to specify the type explicitly then). Example:
https://github.com/Seagate/cortx-hare/blob/main/provisioning/miniprov/test/test_cdf.py#L47-L53

In your case, the type of the element in the list is defintely not Text but DiskRef. You can write a unit test where this list becomes empty and see how Dhall complains 😉

provisioning/miniprov/test/test_cdf.py
```diff
@@ -52,6 +52,16 @@ def test_disks_empty(self):
             '{ runs_confd = Some True, io_disks = { meta_data = None Text, data = [] : List Text } }',
             str(val))
 
+    def test_pooldesc_empty(self):
```
knekrasov, line 55: I would suggest one more test case - Mock the confstore in such a way that list of disk refs becomes empty and make sure that Dhall will not complain (i.e. the construction is still valid in terms of Dhall types).

provisioning/miniprov/test/test_cdf.py
```diff
@@ -107,6 +124,12 @@ def ret_values(value: str) -> Any:
                 1,
                 'cluster>srvnode_1>network>data>interface_type':
                 'o2ib',
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>site>storage_set_count': 1,
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set>server_node_count': 1,
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set1>name': 'StorageSet-1',
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set1>server_nodes': ['srvnode_1'],
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set1>durability>data': 1,
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set1>durability>parity': 0
```
knekrasov, line 132: As far as I understand, there is no test case that checks that DiskRef, PoolDesc or ProfileDesc is actually generated by the given ConfStore values.

Here is the example from the existing tests:
```python
    def test_provided_values_respected(self):
        store = ValueProvider()

        def ret_values(value: str) -> Any:
            data = {
                'cluster>server_nodes': {
                    "blah": "srvnode_1"
                },
                'cluster>srvnode_1>hostname':
                'myhost',
                'cluster>srvnode_1>storage>data_devices': ['/dev/sdb'],
                'cluster>srvnode_1>network>data>private_interfaces':
                ['eth1', 'eno2'],
                'cluster>srvnode_1>storage>metadata_devices': ['/dev/meta'],
                'cluster>srvnode_1>s3_instances':
                1,
                'cluster>srvnode_1>network>data>interface_type':
                'o2ib',
                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>site>storage_set_count': 1,
                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set>server_node_count': 1,
                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set1>name': 'StorageSet-1',
                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set1>server_nodes': ['srvnode_1'],
                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set1>durability>data': 1,
                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set1>durability>parity': 0
            }
            return data[value]

        store._raw_get = Mock(side_effect=ret_values)

        ret = CdfGenerator(provider=store)._create_node_descriptions()
        self.assertIsInstance(ret, list)
        self.assertEqual(1, len(ret))
        self.assertEqual(Text('myhost'), ret[0].hostname)
        self.assertEqual(Text('eth1'), ret[0].data_iface)
        self.assertEqual(1, ret[0].s3_instances)
```
(note the lines like this: `self.assertEqual(Text('eth1'), ret[0].data_iface)` - Text is our own type that will be translated into Dhall expression; you added more such types but there is no test that checks that instances of such types ever get created)

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -37,11 +38,52 @@ def _create_node_descriptions(self) -> List[NodeDesc]:
             nodes.append(self._create_node(node))
         return nodes
 
+    def _create_pool_descriptions(self) -> List[PoolDesc]:
+        pools: List[PoolDesc] = []
+        conf = self.provider
+        cluster_id = conf.get('cluster>cluster_id')
+        storage_set_count = int(
+            conf.get(f'cluster>{cluster_id}>site>storage_set_count'))
+
```
mssawant, line 47: before going on to build disk refs for a storage set lets validate if the #data_devices >= data_units + parity_units + spare_units. If this condition is not satisfied we must fail logging a failure message as "Invalid storage set configuration".

provisioning/miniprov/test/test_cdf.py
mssawant: Also request you to add a test case where total number of data devices are < data_units + parity_units + spare_units to confirm that this case is handled after adding the validation as metioned above.

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -37,11 +38,52 @@ def _create_node_descriptions(self) -> List[NodeDesc]:
...
+        for x in range(storage_set_count):
+            storage_set_name = conf.get(
+                f'cluster>{cluster_id}>storage_set{x+1}>name')
+            pools.append(PoolDesc(
+                name=Text(storage_set_name),
+                disk_refs=Maybe(DList([
+                    DiskRef(path=Text(device), node=Maybe(Text(node), 'Text'))
+                    for node in conf.get(
+                        f'cluster>{cluster_id}>storage_set{x+1}>server_nodes')
```
mssawant, line 56: We are not catching any exception here, what if there does not exist a `cluster>{cluster_id}>storage_set{x+1}>` key?

mssawant (changes requested):

provisioning/miniprov/hare_mp/types.py
```diff
@@ -87,3 +87,30 @@ class NodeDesc(DhallTuple):
     io_disks: DList  # [str]
     meta_data: Text
     s3_instances: int
+
+
+@dataclass(repr=False)
+class DiskRef(DhallTuple):
+    path: Text
+    node: Maybe
```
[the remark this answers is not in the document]
SwapnilGaonkar7: Node is optional, in that case localhost will be considered

provisioning/miniprov/hare_mp/cdf.py
SwapnilGaonkar7: Done

provisioning/miniprov/hare_mp/cdf.py
SwapnilGaonkar7: Going to add new class for exception and handle all key related exception using that class.
Doing that in other task.

SwapnilGaonkar7:
@mssawant Please have a look

provisioning/miniprov/hare_mp/cdf.py
mssawant: Okay please create a ticket for the same and update here.

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -37,11 +38,68 @@ def _create_node_descriptions(self) -> List[NodeDesc]:
...
+                for device in conf.get(
+                        f'cluster>{node}>storage>data_devices'):
+                    data_devices_count += 1
+
+            data_units_count = int(conf.get(
+                f'cluster>{cluster_id}>storage_set{x+1}>durability>data'))
+            parity_units_count = int(conf.get(
+                f'cluster>{cluster_id}>storage_set{x+1}>durability>parity'))
+
+            #TODO We also need to add 'spare units' in below check
```
mssawant, line 64: we do have a key for spare units in conf-store `cluster>{cluster-id}>storage_set[N]>durability>spare` why not use it?
if not populated we can set it default to `parity_units`.

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -37,11 +38,68 @@ def _create_node_descriptions(self) -> List[NodeDesc]:
...
+            conf.get(f'cluster>{cluster_id}>site>storage_set_count'))
+
+        for x in range(storage_set_count):
+            storage_set_name = conf.get(
+                f'cluster>{cluster_id}>storage_set{x+1}>name')
+
+            data_devices_count: int = 0
+            for node in conf.get(
+                    f'cluster>{cluster_id}>storage_set{x+1}>server_nodes'):
+                for device in conf.get(
```
mssawant, line 55: why for? `conf.get(f'cluster>{node}>storage>data_devices')` will return a list right, you can do `len(conf.get(f'cluster>{node}>storage>data_devices')` or save the list and `len(dev_list)`, no?

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -37,11 +38,68 @@ def _create_node_descriptions(self) -> List[NodeDesc]:
...
+        conf = self.provider
+        cluster_id = conf.get('cluster>cluster_id')
+        storage_set_count = int(
+            conf.get(f'cluster>{cluster_id}>site>storage_set_count'))
+
+        for x in range(storage_set_count):
+            storage_set_name = conf.get(
+                f'cluster>{cluster_id}>storage_set{x+1}>name')
+
+            data_devices_count: int = 0
```
mssawant, line 52: [*suggestion*] nr_data_devices, we can follow a naming convention through the code.

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -37,11 +38,68 @@ def _create_node_descriptions(self) -> List[NodeDesc]:
...
+            for node in conf.get(
+                    f'cluster>{cluster_id}>storage_set{x+1}>server_nodes'):
+                for device in conf.get(
+                        f'cluster>{node}>storage>data_devices'):
+                    data_devices_count += 1
+
+            data_units_count = int(conf.get(
```
mssawant, line 59: [*suggestion*] just data_units and parity_units should be fine I think.

provisioning/miniprov/hare_mp/cdf.py
mssawant: Okay, kindly create a ticket and update here please.

provisioning/miniprov/hare_mp/types.py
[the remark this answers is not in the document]
mssawant: Okay.

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -37,11 +38,68 @@ def _create_node_descriptions(self) -> List[NodeDesc]:
             nodes.append(self._create_node(node))
         return nodes
 
+    def _create_pool_descriptions(self) -> List[PoolDesc]:
+        pools: List[PoolDesc] = []
+        conf = self.provider
+        cluster_id = conf.get('cluster>cluster_id')
```
mssawant, line 44: I don't see `cluster>cluster_id` but there is `server_node>{machine-id}>cluster_id` in [conf-store-keys](https://seagatetechnology.sharepoint.com/:x:/r/sites/gteamdrv1/tdrive1224/_layouts/15/Doc.aspx?sourcedoc=%7B46F15394-2BA7-4D44-8B7E-67F1AEAEA654%7D&file=LDR%20R2%20Specifications%20-%20Provisioning.xlsx&action=default&mobileredirect=true&cid=c993b23a-e712-46a5-8d38-ca987a646b31)

SwapnilGaonkar7: cluster_id is available at multiple path (e.g. server_node>{machine-id}>cluster_id, storage_enclosure>{enclosure-id}>cluster_id, etc.).
Also key I used is present in json file currently used in deployment and looks more logical than other places as it is at cluster level.
http://eos-jenkins.colo.seagate.com/job/Cortx-Main/job/centos-7.8.2003/job/Main%20Deploy/88/artifact/provisioner_cluster.json/*view*/

Let me know your thoughts.

Note: Not specific to this but in general, I am assuming that all the keys which are defined and used by us for single node cluster will be still valid and addition of new keys mentioned in 'ConfStore Keys (N-Node)'

provisioning/miniprov/hare_mp/cdf.py
knekrasov: I don't think that there is a specific naming rule in Python for these cases. But I also don't think that `nr` reads better than `count`. So I would suggest to keep it as is.

provisioning/miniprov/hare_mp/cdf.py
SwapnilGaonkar7: Done

provisioning/miniprov/hare_mp/cdf.py
SwapnilGaonkar7: Done

SwapnilGaonkar7:
@mssawant  I have addressed most comments, please have a look.

provisioning/miniprov/hare_mp/cdf.py
SwapnilGaonkar7: https://jts.seagate.com/browse/EOS-18813

provisioning/miniprov/hare_mp/cdf.py
SwapnilGaonkar7: Modified code to use 'server_node>{machine-id}>cluster_id'

mssawant (approved):

provisioning/miniprov/hare_mp/cdf.py
mssawant: > Not specific to this but in general, I am assuming that all the keys which are defined and used by us for single node cluster will be still valid after addition of new keys mentioned in 'ConfStore Keys (N-Node)'

Not necessarily, I see they have changed the keys altogether.

mssawant: It is possible that CI may fail until we have the correct conf-store file. But that is fine, we need to fix that.
Deployment failure is also expected but that too needs to be fixed by the provisioning team. I will confirm this.

provisioning/miniprov/test/test_cdf.py
```diff
@@ -119,6 +143,59 @@ def ret_values(value: str) -> Any:
...
+        self.assertEqual(Text('/dev/sdb'), disk_refs.value[0].path)
+
+        ret = CdfGenerator(provider=store)._create_profile_descriptions(ret)
+        self.assertIsInstance(ret, list)
+        self.assertEqual(1, len(ret))
+        self.assertEqual(Text('Profile_the_pool'), ret[0].name)
+        self.assertEqual(1, len(ret[0].pools.value))
+        self.assertEqual(Text('StorageSet-1'), ret[0].pools.value[0])
+
+    def test_invalid_storage_set_configuration(self):
```
knekrasov, line 163: Suggestion: rename the method to something like `test_invalid_storage_set_configuration_rejected`.

Reason: test names should (ideally) specify the statement about software unit or give a specific description of how the software is supposed to behave.

So if you have statement like "My software analyzes the correctness of storage set configuration AND it raises an exception if is incorrect", then it would be nice to give the corresponding clue in the test name. Note that unit tests can a kind of document the software and how it is supposed to work.

knekrasov: It would be nice to put a comment within test method describing why exactly the given configuration is incorrect, it is not that obvious to me. That would help to maintain your test in the future.

SwapnilGaonkar7: Done

knekrasov (approved):

knekrasov:
VM Deployment fails but that is expected since new keys ConfStore keys are now required; those keys are not provided by the CI yet.

There is a bug (not related to this PR though) that allows CI to proceed with `init` phase although no CDF file was generated. The bug will be addressed by  https://github.com/Seagate/cortx-hare/pull/1546

I'll merge this one.

PR #1546, opened 2021-03-17
Problem: Generate exception 'KeyError' for handling missing key scenario

SwapnilGaonkar7:
Solution: Modified code to raise 'KeyError' exception when key is missing

UT: Tested 'config', 'init' and 'test' sections of 'hare_setup' commands

In case key is not present following error is displayed,
(.py3venv) [root@ssc-vm-c-1902 hare]# /opt/seagate/cortx/hare/bin/hare_setup config --config 'json:///root/provisioner_cluster.json' --file '/var/lib/hare/cluster12.yaml'
Error performing configuration ('Required key cluster>srvnode-1>storage>data_devices not found')

[root@ssc-vm-c-1902 hare]# echo $?
255

### DIFF
--- a/utils/hare-shutdown
+++ b/utils/hare-shutdown
@@ -81,7 +81,8 @@ def shutdown_cluster():
     if leader_node and not utils.is_fake_leader_name(leader_node):
         logging.info(f'Shutting down RC Leader at {leader_node}... ')
         pkill: str = 'sudo pkill --exact -KILL consul &> /dev/null'
-        utils.exec_custom(utils.ssh_prefix(leader_node) + pkill)
+        utils.exec_custom(utils.ssh_prefix(leader_node) + pkill,
+                          show_err=False)
 
 
 def node_shutdown():
--- a/utils/utils.py
+++ b/utils/utils.py
@@ -188,12 +188,17 @@ def exec_silent(cmd: str) -> bool:
     return subprocess.call(cmd, shell=True) == 0
 
 
-def exec_custom(cmd: str) -> None:
+def exec_custom(cmd: str, show_err: bool = True) -> None:
     assert cmd
-    if exec_silent(cmd):
-        logging.info('OK')
-    else:
-        logging.error('**ERROR**')
+
+    def handle(result: bool) -> None:
+        if not show_err:
+            return
+        if result:
+            logging.info('OK')
+        else:
+            logging.error('**ERROR**')
+    handle(exec_silent(cmd))
 
 
 def process_stop(proc: Process) -> None:


provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -33,6 +33,8 @@ def _create_node_descriptions(self) -> List[NodeDesc]:
         nodes: List[NodeDesc] = []
         conf = self.provider
         server_dict: Dict[str, str] = conf.get('cluster>server_nodes')
+        if not server_dict:
+            raise MissingKey(Text('cluster>server_nodes'))
```
knekrasov, line 37: Why use `Text`?

Exceptions help to distinguish the nature of the problem (exception type) and help to provide human-friendly text that explains what was wrong (message).

In this case, I guess `raise MissingKey('Required key cluster>server_nodes not found')` will make way more sense.

Second point is.. Can't we move this `if` and this raising into the logic of `conf.get()`? Semantics of the call would be "I return the value or I raise the exception if the value is not present". That would simplify the code around, no extra checks will need to be added then.

SwapnilGaonkar7: 1. In this case, I guess raise MissingKey('Required key cluster>server_nodes not found') will make way more sense.

> Final output will be 'Error performing configuration. Missing ConfStore key ("cluster>server_nodes")_'.  Config() function which handles MissingKey exception will output 'Missing ConfStore key (%s)', key_error' by replacing 'key_error' with specific key.

If we move raising logic to conf.get() then it will always raise exception when key is not preset.
We may not always want to raise exception when key is missing, sometimes we can just take default value if key is missing.

We can still achieve this with moving raising logic to conf.get() and handling exception at caller but cases where we don't want to raise exception if key is missing needs to be handled separately.

@knekrasov  Let me know your inputs

knekrasov: @SwapnilGaonkar7 `Text` type is a kind of [DTO type](https://en.wikipedia.org/wiki/Data_transfer_object) (or [AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree), to be more precise) to express something in Dhall language. `'Missing ConfStore key (%s)', key_error` will invoke `__str__()` from `Text` class which will return the wrapped string. If you wrap a string with `Text` and then only thing that you do is implicitly unwrap it again into a string, then I don't see why `Text` is needed. From logical point of view, it makes no sense to me.

> We can still achieve this with moving raising logic to conf.get() and handling exception at caller but cases where we don't want to raise exception if key is missing needs to be handled separately.

On the contrary, I would like it to raise the exception whenever the key is incorrect. It will make the caller code more straightforward, there is no fear that somebody just forgets to check if the value is None (and thus will not notice a mistype in the key name). Please see [fail-fast](https://en.wikipedia.org/wiki/Fail-fast) principle for more details.

If by some reason the programmer will want another behavior, (s)he can add one more method like `if_key_exists` (or just implement `in` operator) but nowadays we don't have such a need.

knekrasov: Small update:
In my current PR I already realized the need of some values to be None (network_interface_type can be None). So I implemented it like this:

```python
class ValueProvider:
    def get(self, key: str, allow_null: bool = False) -> Any:
        ret = self._raw_get(key)
        if ret is None and not allow_null:
            raise RuntimeError(f'ConfStore key {key} not found')
        return ret

    def _raw_get(self, key: str) -> str:
        raise NotImplementedError()


class ConfStoreProvider(ValueProvider):
    def __init__(self, url: str):
        self.url = url
        conf = ConfStore()
        conf.load('hare', url)
        self.conf = conf

    def _raw_get(self, key: str) -> str:
        return self.conf.get('hare', key)
```

So when your PR is merged, I'll just update raw `RuntimeError` with your `MissingKey` type in my code above.

knekrasov: Just for reference: #1548

SwapnilGaonkar7: @knekrasov Addressed your review comments. Please have a look

provisioning/miniprov/hare_mp/store.py
```diff
@@ -22,11 +22,15 @@
 from typing import Any
 
 from cortx.utils.conf_store import ConfStore
+from hare_mp.types import MissingKey
 
 
 class ValueProvider:
     def get(self, key: str) -> Any:
-        return self._raw_get(key)
+        value = self._raw_get(key)
```
knekrasov, line 30: Do you unit tests pass for you?

SwapnilGaonkar7: It was failing. So I have modified test case.
@knekrasov Please have a look

provisioning/miniprov/test/test_cdf.py
```diff
@@ -210,5 +210,5 @@ def ret_values(value: str) -> Any:
 
         store._raw_get = Mock(side_effect=ret_values)
 
-        ret = CdfGenerator(provider=store)._create_node_descriptions()
-        self.assertEqual('None P', str(ret[0].data_iface_type))
+        with self.assertRaises(MissingKey):
+            CdfGenerator(provider=store)._create_node_descriptions()
```
knekrasov, line 214: I'm sorry but this is not correct. The fact is that the method name `test_iface_type_can_be_null` states explicitly what is the predicate: it is normal if iface_type is null, our application will not break, it is expected and handled. But your change assumes that the exception will be thrown.


Please have a look at my PR - I solved the same problem there: #1548
It is fine if you copy and paste the code from there, we just need to unblock your PR and get it merged. I'll resolve conflicts in my PR then.

mssawant: I agree, this does not simulate the actual scenario.  We need to specifically cat the the null KVs that are expected to be not null.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -35,6 +35,7 @@
 
 from hare_mp.cdf import CdfGenerator
 from hare_mp.store import ConfStoreProvider
+from hare_mp.types import MissingKey
```
mssawant, line 38: How about prefixing or suffixing with Exception or Error?
I believe there's already a KeyError exception for json, no? cannot we use that, why do we need to have our own exception in this case?

provisioning/miniprov/test/test_cdf.py
SwapnilGaonkar7: Done

SwapnilGaonkar7:
@mssawant @knekrasov I have addressed review comments.
Note : in main.py we have
def config(args):
    try:
        url = args.config[0]
        filename = args.file[0] or '/var/lib/hare/cluster.yaml'
    except Exception as error:
        logging.error('Error performing configuration (%s)', error)
        
 I haven't handled 'KeyError' separately since we are just displaying message which contains more info about exception..

provisioning/miniprov/hare_mp/store.py
```diff
@@ -25,8 +25,11 @@
 
 
 class ValueProvider:
-    def get(self, key: str) -> Any:
-        return self._raw_get(key)
+    def get(self, key: str, allow_null: bool = False) -> Any:
+        ret = self._raw_get(key)
+        if ret is None and not allow_null:
```
mssawant, line 30: I think we need to set expectations clearly here, `allow_null` is for us, hare. We know that this KV is not provided by conf-store and thus Hare is setting it to `null` by default. But if the KV is present in conf-store then it cannot be null I believe.  This routine is used to fetch KV from conf-store, that means Hare knows that the given key is present in conf-store. It cannot be `null` if the value is `None` we must raise an exception I think. cc @knekrasov.

mssawant:
@SwapnilGaonkar7 , kindly update the PR description and summary. I believe we are not adding a new class but using the one from json package, right?

provisioning/miniprov/hare_mp/main.py
knekrasov: @mssawant Because KeyError is probably not the only "symptom"  that the key is not found in ConfStore. For instance, ConfStore can contain an array with incomplete set of indices or something like that. Since we don't want to assume so many different exceptions at the top level of main.py, I suggested to create our own exception. So if we see exception MissingKey, it means that ConfStore doesn't match our expectations. No matter how the underlying layer came to that decision.

mssawant: >For instance, ConfStore can contain an array with incomplete set of indices or something like that.

Okay, so may be we got server_nodes count as 3 but only 2 of them are described. I somehow still think its a KeyError. But its okay, if we think that a separate exception class is more elaborate then let's use a separate exception class. I am fine with it.

provisioning/miniprov/hare_mp/store.py
knekrasov: @mssawant This piece of code is actually mine 😁 

The fact is that ConfStore's get() method returns None if the JSON file doesn't have that value. So either our code is incorrect (i.e. the key name is misspelled), or ConfStore is not full (we saw that as well) or our code doesn't correspond ConfStore. That's why I insisted to throw an exception if ConfStore returns None by default.

But there is at least one situation when ConfStore can return None - iface_type. It can be omitted. Then the caller will need to put `allow_null=True`. But that's a kind of special case and the caller code is quite explicit about that.

mssawant: Right, we need to raise an exception for `None` returned from conf-store.

But if `iface_type` key is present in conf-store then why would conf-store not have a value to assigned to it? I think we need to take this up with conf-store team to set a default value instead of returning `None` for it if the key is present. Problem here is that if we allow the keys to `None` we need to understand the circumstances, the contract needs to be more clear, no?

SwapnilGaonkar7:
> @SwapnilGaonkar7 , kindly update the PR description and summary. I believe we are not adding a new class but using the one from json package, right?

SwapnilGaonkar7:
@mssawant @knekrasov Please have a look

provisioning/miniprov/hare_mp/store.py
knekrasov: Not sure. According to Dhall types, iface_type is optional.  So it can be None. And thus it can be omitted in CDF file, CFGEN will take care of that and interpret it according to its own logic. I don't see a reason why we should make a decision on default values at the level of cdf generator.

knekrasov: > But if iface_type key is present in conf-store then why would conf-store not have a value assigned to it? 

ConfStore can miss that key. That is the reason why ConfStore API can return None.

knekrasov (approved):

PR #1547, opened 2021-03-18
EOS-18645: evaluate execution of mini provisioner setup command

mssawant:
Component mini-provisioning section, e.g. post-install, config, init, etc,
are expected to be executed on every node in the cluster. Thus it is
component's responsibility to evaluate if the command must do anything or
nothing on that node.
Hare needs to evaluate if the mini-provisioner init and test section
executions must be a no-op on a given node or not.

Solution:
- Check if the node is the first node of the first storage set in the
  cluster using information from the comf-store.
- Add a validator class to encapsulate common routines with respect to
  validations.
- Add routines to ConfStoreProvider for frequently fetched conf-store keys.

### DIFF
--- a/provisioning/miniprov/hare_mp/store.py
+++ b/provisioning/miniprov/hare_mp/store.py
@@ -28,10 +28,16 @@ class ValueProvider:
             raise RuntimeError(f'ConfStore key {key} not found')
         return ret
 
+    def _raw_get(self, key: str) -> str:
+        raise NotImplementedError()
+
     def get_machine_id(self) -> str:
         raise NotImplementedError()
 
-    def _raw_get(self, key: str) -> str:
+    def get_cluster_id(self) -> str:
+        raise NotImplementedError()
+
+    def get_storage_set_id(self) -> int:
         raise NotImplementedError()
 
 
@@ -42,10 +48,21 @@ class ConfStoreProvider(ValueProvider):
         conf.load('hare', url)
         self.conf = conf
 
+    def _raw_get(self, key: str) -> str:
+        return self.conf.get('hare', key)
+
     def get_machine_id(self) -> str:
-        with open('/etc/machine_id', 'r') as f:
+        with open('/etc/machine-id', 'r') as f:
             machine_id = f.readline().strip('\n')
             return machine_id
 
-    def _raw_get(self, key: str) -> str:
-        return self.conf.get('hare', key)
+    def get_cluster_id(self) -> str:
+        machine_id = self.get_machine_id()
+        cluster_id = self._raw_get(f'server_node>{machine_id}>cluster_id')
+        return cluster_id
+
+    def get_storage_set_id(self) -> int:
+        machine_id = self.get_machine_id()
+        storage_set_id = self._raw_get((f'server_node>{machine_id}>'
+                                        f'storage_set_id'))
+        return int(storage_set_id)
--- a/provisioning/miniprov/hare_mp/validator.py
+++ b/provisioning/miniprov/hare_mp/validator.py
@@ -1,0 +1,54 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+# Setup utility for Hare to configure Hare related settings, e.g. logrotate,
+# report unsupported features, etc.
+
+import socket
+from hare_mp.store import ValueProvider
+
+
+class Validator:
+    def __init__(self, provider: ValueProvider):
+        super().__init__()
+        self.provider = provider
+
+    def _get_machine_id(self) -> str:
+        machine_id = self.provider.get_machine_id()
+        if not self.is_local_machine_id_valid(machine_id):
+            raise RuntimeError(f'invalid machine id {machine_id}')
+        return machine_id
+
+    def is_first_node_in_storage_set(self) -> bool:
+        machine_id = self._get_machine_id()
+        cluster_id = self.provider.get_cluster_id()
+        storage_set_id = self.provider.get_storage_set_id()
+        server_nodes_key: str = (f'cluster>{cluster_id}>'
+                                 f'storage_set[{storage_set_id}]>server_nodes')
+        server_nodes = self.provider.get(server_nodes_key)
+        return server_nodes[0] == machine_id
+
+    def is_first_node_in_cluster(self) -> bool:
+        machine_id = self._get_machine_id()
+        cluster_id = self.provider.get_cluster_id()
+        server_nodes_key: str = (f'cluster>{cluster_id}>'
+                                 f'storage_set[0]>server_nodes')
+        server_nodes = self.provider.get(server_nodes_key)
+        return server_nodes[0] == machine_id
+
+    def is_local_machine_id_valid(self, machine_id: str) -> bool:
+        hostname = self.provider.get(f'server_node>{machine_id}>hostname')
+        return hostname == socket.gethostname()
--- a/provisioning/miniprov/test/test_validator.py
+++ b/provisioning/miniprov/test/test_validator.py
@@ -1,0 +1,159 @@
+# Copyright (c) 2020 Seagate Technology LLC and/or its Affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+# Setup utility for Hare to configure Hare related settings, e.g. logrotate,
+# report unsupported features, etc.
+
+# flake8: noqa
+import unittest
+import json
+import socket
+
+from hare_mp.store import ConfStoreProvider
+from hare_mp.validator import Validator
+
+
+URL = "json:///tmp/test-conf-store.json"
+
+conf_store_data = {
+  "provisioner": {
+    "cluster_info": {
+      "pillar_dir": "/opt/seagate/cortx/provisioner/pillar/groups/all/",
+      "num_of_nodes": 1
+    },
+    "cluster": {
+      "num_of_nodes": "1"
+    }
+  },
+  "storage": {
+  },
+  "system": {
+  },
+  "server_node": {
+    "srvmachine-1": {
+      "name": "srvnode-1",
+      "hostname": "ssc-vm-1623.colo.seagate.com",
+      "cluster_id": "e766bd52-c19c-45b6-9c91-663fd8203c2e",
+      "storage_set_id": 0
+    }
+  },
+  "cluster": {
+    "e766bd52-c19c-45b6-9c91-663fd8203c2e": {
+      "name": "cortx-cluster",
+      "storage_set": [
+          { "name": "StorageSet-1",
+           "server_nodes": [ "srvmachine-1" ]
+          }
+       ]
+    },
+    "storage_sets": {
+      "storage-set-1": [
+        "srvnode-1"
+      ]
+    },
+    "server_nodes": {
+      "7c4fd75dfedd7662e6a39b0a53274922": "srvnode-1"
+    },
+    "cluster_id": "e766bd52-c19c-45b6-9c91-663fd8203c2e",
+    "srvnode-1": {
+      "storage_set_id": 0,
+      "hostname": "ssc-vm-1623.colo.seagate.com",
+      "node_type": "VM",
+      "roles": [
+        "primary",
+        "openldap_server",
+        "kafka_server"
+      ],
+      "is_primary": "true",
+      "bmc": {
+      },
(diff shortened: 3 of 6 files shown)


mssawant:
Manually tested the code, unit tests are in-progress.

provisioning/miniprov/hare_mp/store.py
```diff
@@ -41,3 +42,26 @@ def __init__(self, url: str):
 
     def _raw_get(self, key: str) -> str:
         return self.conf.get('hare', key)
+
+    def get_machine_id(self) -> str:
+        with open('/etc/machine_id', 'r') as f:
+            machine_id = f.readline().strip('\n')
+            hostname = self.conf.get('hare',
```
knekrasov, line 49: Let's use `self._raw_get()` instead. When we have a single point where our class interacts to the real confstore, it is simpler to test (mock) and/or override the functionality.

provisioning/miniprov/hare_mp/store.py
```diff
@@ -41,3 +42,26 @@ def __init__(self, url: str):
...
+                                     f'server_node>{machine_id}>hostname')
+            if hostname != socket.gethostname():
+                raise RuntimeError(f'invalid machine id {machine_id}')
+            return machine_id
+        raise RuntimeError(f'cannot read machine_id')
+
+    def get_cluster_id(self) -> str:
+        machine_id = self.get_machine_id()
+        cluster_id = self.conf.get('hare',
```
knekrasov, line 58: The same thing. Let's replace self.conf.get with our `self._get_raw()`. It is more practical to have just one place that interacts with ConfStore.

provisioning/miniprov/hare_mp/store.py
```diff
@@ -41,3 +42,26 @@ def __init__(self, url: str):
...
+                                   f'server_node>{machine_id}>cluster_id')
+        return cluster_id
+
+    def get_storage_set_id(self) -> int:
+        conf = self.conf
+        machine_id = self.get_machine_id()
+        storage_set_id = conf.get('hare',
```
knekrasov, line 65: `self._get_raw()` please

provisioning/miniprov/hare_mp/validate.py
```diff
@@ -0,0 +1,43 @@
...
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+# Setup utility for Hare to configure Hare related settings, e.g. logrotate,
+# report unsupported features, etc.
+
+from hare_mp.store import ConfStoreProvider
+from hare_mp.store import ValueProvider
+
+class Validate:
```
knekrasov, line 23: Why Validate, not Validator? Per my understanding, it executes some action but not motivate other components, so verb is probably not suitable.

provisioning/miniprov/hare_mp/validate.py
```diff
@@ -0,0 +1,43 @@
...
+    def __init__(self, provider: ValueProvider):
+        super().__init__()
+        self.provider = provider
+
+    def is_first_node_in_storage_set(self) -> bool:
+        machine_id = self.provider.get_machine_id()
```
knekrasov, line 29: [optional] If Validator then invokes `self.provider.get()` anyway, does it really make sense to encapsulate two other keys into `get_machine()` and `get_cluster_id()`?

provisioning/miniprov/hare_mp/validate.py
```diff
@@ -0,0 +1,43 @@
...
+        cluster_id = self.provider.get_cluster_id()
+        storage_set_id = self.get_storage_set_id()
+        server_nodes_key: str = (f'cluster>{cluster-id}>'
```
knekrasov, line 32: `s/cluster-id/cluster_id` please

provisioning/miniprov/hare_mp/validate.py
knekrasov: It feels like it makes more sense to make `get_machine()` and `get_cluster_id()` be the part of this particular class instead of ValueProvider.

knekrasov (changes requested):

provisioning/miniprov/hare_mp/store.py
```diff
@@ -41,3 +42,26 @@ def __init__(self, url: str):
 
     def _raw_get(self, key: str) -> str:
         return self.conf.get('hare', key)
+
+    def get_machine_id(self) -> str:
+        with open('/etc/machine_id', 'r') as f:
+            machine_id = f.readline().strip('\n')
+            hostname = self.conf.get('hare',
+                                     f'server_node>{machine_id}>hostname')
+            if hostname != socket.gethostname():
```
knekrasov, line 51: It reminds me [why URL.equals() is broken](https://news.ycombinator.com/item?id=21765788) in Java (in short: because using DNS lookups to compare two URLs is not a good idea).

1. Why do we need to compare the hostname via socket.gethostname()
2. Are you sure it is a correct place for this check?

I agree that this functionality has some benefit but I'm a bit concerned about the fact that this method has two separate responsibilities. In other words, I just want to get the machine ID but instead the method will also resolve  the hostname and raise an exception if the hostname is not that good. I didn't ask for hostname, I wanted machine_id only 😁

provisioning/miniprov/hare_mp/store.py
mssawant: :), did use `_raw_get()` before, changed it to `conf.get()`. Yeah, makes sense to use `_raw_get()`.

provisioning/miniprov/hare_mp/store.py
mssawant: Yes, here we are verifying the machine id we got from conf-store if it is of the same node that we expect. We confirm this by comparing the hostname from conf-store for the given machine-id with the node's hostname. Its just another error case that we are trying to catch that may avoid additional debugging in case of an inconsistent conf-store.

provisioning/miniprov/hare_mp/store.py
mssawant: Done.

provisioning/miniprov/hare_mp/store.py
mssawant: Done.

provisioning/miniprov/hare_mp/validate.py
mssawant: Yep, makes sense.

provisioning/miniprov/hare_mp/validate.py
mssawant: >If Validator then invokes self.provider.get() anyway, does it really make sense to encapsulate two other keys into get_machine() and get_cluster_id()?
Idea is to keep all the conf-store related functions at one place, close to conf-store. Yes, we are still invoking provider but imagine get_cluster_id() and get_machine_id() as helpers. I expect we may need to access at least these 2 KVs frequently as most of the keys need them. Thus keeping a single function would help us to make future changes to the keys as they will be in one place instead of getting spread across multiple places.

I think having them in ValueProvider keeps all the conf-store related keys and functionality at one place.

provisioning/miniprov/hare_mp/validate.py
mssawant: ah, I did change this after testing, some how it is not updated in the patch.

mssawant: Done.

provisioning/miniprov/hare_mp/validate.py
knekrasov: Ok, fine

provisioning/miniprov/hare_mp/store.py
knekrasov: I see but can we do that check in a separate method?

provisioning/miniprov/hare_mp/store.py
mssawant: Done.

provisioning/miniprov/hare_mp/store.py
mssawant: Okay, separate method makes sense.

mssawant: Done.

mssawant:
@knekrasov, @SwapnilGaonkar7 , addressed comments, fixed some flake errors in mini provisioner code and added unit test.

provisioning/miniprov/hare_mp/store.py
```diff
@@ -41,3 +50,20 @@ def __init__(self, url: str):
 
     def _raw_get(self, key: str) -> str:
         return self.conf.get('hare', key)
+
+    def get_machine_id(self) -> str:
+        with open('/etc/machine_id', 'r') as f:
+            machine_id = f.readline().strip('\n')
+            return machine_id
+        raise RuntimeError('cannot read machine_id')
```
knekrasov, line 58: [Minor] this instruction looks not feasible (dead code). Can be easily removed.

provisioning/miniprov/hare_mp/validator.py
```diff
@@ -0,0 +1,56 @@
...
+        machine_id = self._get_machine_id()
+        cluster_id = self.provider.get_cluster_id()
+        server_nodes_key: str = (f'cluster>{cluster_id}>'
+                                 f'storage_set[0]>server_nodes')
+        server_nodes = self.provider.get(server_nodes_key)
+        return server_nodes[0] == machine_id
+
+    def is_local_machine_id_valid(self, machine_id: str) -> bool:
+        hostname = self.provider.get(f'server_node>{machine_id}>hostname')
+        if hostname == socket.gethostname():
```
knekrasov, line 54: Can be done simpler:
```python
return hostname == socket.gethostname()
```

provisioning/miniprov/test/test_validator.py
```diff
@@ -0,0 +1,163 @@
...
+
+class TestValidator(unittest.TestCase):
+    def test_is_cluster_first_node(self):
+        conf = ConfStoreProvider(URL)
+        hostname = socket.gethostname()
+        machine_id = conf.get_machine_id()
+        update_machine(machine_id, hostname)
+        validator = Validator(
+                        ConfStoreProvider("json:///tmp/temp-test-conf-store.json"))
+        assert validator.is_first_node_in_cluster()
```
knekrasov, line 151: [`self.assertEqual()`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual) please

`assert` instruction can be thrown away by the compiler. Anyway, I would suggest to use the framework-related API calls. `unittest` is our API for tests. Let's use it.

provisioning/miniprov/test/test_validator.py
```diff
@@ -0,0 +1,163 @@
...
+        conf = ConfStoreProvider(URL)
+        hostname = 'invalid-hostname'
+        machine_id = conf.get_machine_id()
+        update_machine(machine_id, hostname)
+        validator = Validator(
+                        ConfStoreProvider("json:///tmp/temp-test-conf-store.json"))
+        try:
+            machine_id = validator._get_machine_id()
+        except RuntimeError:
+            return
```
knekrasov, line 163: Unit tests should be fail-fast: if anything abnormal happens, shout loudly.

Why do we swallow this exception? Can't it hide a greater problem?

knekrasov: BTW you probably should take a look at [`assertRaises`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises)

It will pass if and only if the given exception is raised. Otherwise it will fail the test.

provisioning/miniprov/hare_mp/validator.py
```diff
@@ -0,0 +1,56 @@
...
+        return machine_id
+
+    def is_first_node_in_storage_set(self) -> bool:
+        machine_id = self._get_machine_id()
+        cluster_id = self.provider.get_cluster_id()
+        storage_set_id = self.provider.get_storage_set_id()
+        server_nodes_key: str = (f'cluster>{cluster_id}>'
+                                 f'storage_set[{storage_set_id}]>server_nodes')
+        server_nodes = self.provider.get(server_nodes_key)
+        return server_nodes[0] == machine_id
```
SwapnilGaonkar7, line 42: machine_id will be in format '0a31807456be14965f31b49314f5664e'
whereas 'server_nodes' will be in format as shown below,
cluster>{cluster-id}>storage_set[N]>server_nodes = ["srvmachine-1", "srvmachine-2", "srvmachine-3"] (taken from sheet).
@mssawant Please check if they are really in same format.

knekrasov: Per my understanding server_nodes contains machine id's. In other words, I believe Mandar's patch is correct.

mssawant: If you read the description of the `cluster>{cluster-id}>storage_set[N]>server_nodes` key it says its a list of machine-ids.
```
List of server node machine-ids belonging to Storage-Set N
```
We have not hardcoded anything, we will use whatever we get from the conf-store.

provisioning/miniprov/hare_mp/validator.py
mssawant: Ah yes. Fixed.

provisioning/miniprov/hare_mp/validator.py
SwapnilGaonkar7: You are correct,
Sample json generated by Nitin contains following lines,
"storage_set": [
        {
          "name": "storage-set-1",
          "server_nodes": [
            "a971a87aace34e05b4747bb3376d6661"
          ],

provisioning/miniprov/test/test_validator.py
mssawant: Agree, thanks. I missed this.

provisioning/miniprov/hare_mp/store.py
knekrasov: Yes. Otherwise `with` construct would have not much sense.

From [Python docs](https://docs.python.org/3/library/functions.html#open):
> Open file and return a corresponding file object. If the file cannot be opened, an OSError is raised. See Reading and Writing Files for more examples of how to use this function.

provisioning/miniprov/test/test_validator.py
mssawant: This is a negative test case and we are checking if the failure path is taken care of. So if the machine-id does not match the hostname (for which we have a handling in the code) we need to make sure that the code raise appropriate exception so that we do not go ahead and fail later. Thus we are putting some invalid values in the given conf-store and making sure that the main code raises the exception but test case itself must pass as our expectation matches.

Will look into assertRaises, thanks.

mssawant: Done.

knekrasov (approved):

suvratjoshi (approved):

PR #1548, opened 2021-03-18
Problem: Hare has no ConfStore templates

knekrasov:
Solution:
1. Add ConfStore templates with `conf` CLI tool (from cortx-py-utils)
2. Add unit test that verifies that the template (for config stage) is
sufficient for Hare needs (so we don't lie ourselves).
3. Improve ConfStore wrapper, make it fail-fast.

JIRA ticket: https://jts.seagate.com/browse/EOS-18695


provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -24,17 +24,17 @@ def _get_dhall_path(self) -> str:
...
             'hare_mp', resource_path)
         return raw_content.decode('utf-8')
 
     def _create_node_descriptions(self) -> List[NodeDesc]:
         nodes: List[NodeDesc] = []
         conf = self.provider
-        server_dict: Dict[str, str] = conf.get('cluster>server_nodes')
-        for _, node in server_dict.items():
-            nodes.append(self._create_node(node))
+        machines: Dict[str, Any] = conf.get('server_node')
```
mssawant, line 35: sorry, don't see this key in the list of conf-store keys, didn't get how do we get the machines?

knekrasov: The idea behind this is as follows.
The structure in the ConfStore is as follows `server_node>{machine-id}>everything-that-we-actually-need`. The problem is that there is no special key that lists all possible machine ids. While we need all of them to build CDF and specify all nodes.

So I tried this kind of hack, it turns out to be working with the current cortx-py-utils (installed from here http://cortx-storage.colo.seagate.com/releases/cortx/github/main/centos-7.8.2003/798/prod/cortx_iso/cortx-py-utils-2.0.0-182_6708fa0.noarch.rpm).

If there is a better way to work this around, I will be happy to follow.

knekrasov: In other words, I just need to know all valid `machine_id`s to substitute into `server_node>{machine-id}>everything-that-we-actually-need`. So no, this is not a recursive fetching. It is just a way to find all possible machine-ids.

mssawant: >So I tried this kind of hack, it turns out to be working with the current cortx-py-utils (installed from here

Ah, so you mean for something like this,
```
{
  "server_node": {
    "SOME_MACHINE_ID": {
      "hostname": "myhost",
      "network": {
        "data": {
```
if you query just `server_node` it will give all the keys under `server_node` right? similar to `jq`. But if there are additional keys than `machine_id` under `server` then this will fail. Why not iterate over storage-sets and corresponding server_nodes, `cluster>{cluster_id}>storage_set[{storage_set_id}]>server_nodes` to get the machine_ids for the entire cluster? It may be a bit expensive but I think will be safer, no?

knekrasov: Hm, yes, good idea. I 'll need to try it out on Monday.

knekrasov: @mssawant But it means that we'll need to know `cluster_id` beforehand :-) 

Can we merge this PR (given that currently all the keys under `server_node` are machine IDs) now and probably create an issue for the future?

Perhaps ConfStore keys will change again and all this activity will not make much sense.

mssawant: Okay, if the current method is working fine and relevant tests are passing., then let's merge this and create a separate issue/ticket for improvmment.

mssawant (approved):

SwapnilGaonkar7 (approved):
Looks good

PR #1549, opened 2021-03-18
EOS-18828: wait_for_all may wait for already delivered messages

mssawant:
It is possible that the messages from a halink promise were already delivered
and the sender invokes wait_for_all, which does not check if the promise was
already delivered or not and may block.

Solution:
Check the promise in recently_delivered list before blocking to confirm if
the messages are not already delivered.


knekrasov (approved):
Good catch @mssawant!

suvratjoshi (approved):

ujjwalpl (approved):

PR #1552, opened 2021-03-22
EOS-18751: hctl shutdown reports ERROR on VM

supritshinde:
Description: The **ERROR** message was logged onto the console due to a wrong exit code check. Re-initiating the killing of consul returned wrong exit code as consul is already being killed. And to ensure that in the end, all the processes containing "consul" as a pattern must be killed, we still have the "sudo pkill --exact -KILL consul &> /dev/null" command present.

### DIFF
--- a/utils/utils.py
+++ b/utils/utils.py
@@ -188,12 +188,17 @@ def exec_silent(cmd: str) -> bool:
     return subprocess.call(cmd, shell=True) == 0
 
 
-def exec_custom(cmd: str) -> None:
+def exec_custom(cmd: str, show_err: bool = True) -> None:
     assert cmd
-    if exec_silent(cmd):
-        logging.info('OK')
-    else:
-        logging.error('**ERROR**')
+
+    def handle(result: bool) -> None:
+        if not show_err:
+            return
+        if result:
+            logging.info('OK')
+        else:
+            logging.error('**ERROR**')
+    handle(exec_silent(cmd))
 
 
 def process_stop(proc: Process) -> None:
--- a/utils/hare-shutdown
+++ b/utils/hare-shutdown
@@ -81,7 +81,8 @@ def shutdown_cluster():
     if leader_node and not utils.is_fake_leader_name(leader_node):
         logging.info(f'Shutting down RC Leader at {leader_node}... ')
         pkill: str = 'sudo pkill --exact -KILL consul &> /dev/null'
-        utils.exec_custom(utils.ssh_prefix(leader_node) + pkill)
+        utils.exec_custom(utils.ssh_prefix(leader_node) + pkill,
+                          show_err=False)
 
 
 def node_shutdown():


utils/utils.py
```diff
@@ -174,10 +174,7 @@ def exec_silent(cmd: str) -> bool:
 
 def exec_custom(cmd: str) -> None:
     assert cmd
-    if exec_silent(cmd):
-        logging.info('OK')
-    else:
-        logging.error('**ERROR**')
+    exec_silent(cmd)
```
knekrasov, line 177: @supritshinde Can this command be more picky?

The fact is that we want to swallow non-zero exit code ONLY when it is the latest command ("killing RC leader"). With your changes applied all the commands will be touched, not only the one that kills RC leader.

Suggestion: add flag to this `exec_custom()` method which is False by default (e.g. `show_errors=True`). Pass False in one single case when no ERROR should be shown.

knekrasov (changes requested):

SwapnilGaonkar7:
Suggestion: Can we check if consul is alive/dead and skip killing command if consul is already dead.

suvratjoshi:
> Suggestion: Can we check if consul is alive/dead and skip killing command if consul is already dead.
Yes, would be nice if this can be done.
Additionally this info can be logged as well that consul already killed, in case of skipping.

supritshinde:
@SwapnilGaonkar7 @suvratjoshi Actually konstantin and i had this discussion. So, the pkill is still kept because, in future, if there is any such consul's process other than hare-consul-agent, which has "consul" pattern present in it.  Keeping this command will ensure killing of such processes too. Nothing else. @knekrasov any other thoughts?

utils/utils.py
```diff
@@ -172,12 +172,15 @@ def exec_silent(cmd: str) -> bool:
...
     assert cmd
-    if exec_silent(cmd):
+    if show_errors:
         logging.info('OK')
     else:
-        logging.error('**ERROR**')
+        if exec_silent(cmd):
+            logging.info('OK')
+        else:
+            logging.error('**ERROR**')
```
knekrasov, line 183: show_errors doesn't affect on logging. Did you test your change?

knekrasov: Suggestion:

```python
def exec_custom(cmd: str, show_errors: bool = True) -> None:
    assert cmd

    def handle(result: bool) -> None:
        if not show_errors:
            return

        if result:
            logging.info('OK')
        else:
            logging.error('**ERROR**')

    handle(exec_silent(cmd))
```

knekrasov (changes requested):

utils/utils.py
supritshinde: Done! @knekrasov Modified according to your suggestion.

knekrasov (approved):

supritshinde:
@knekrasov Done! Modified according to changes mentioned. And also resolved the flake rules error from second last commit.

mssawant (approved):

mssawant:
@supritshinde , please fixup/squash the commits appropriately.

supritshinde:
@mssawant Yeah, did it! rebased and squashed the commits into single.

PR #1553, opened 2021-03-22
EOS-18846: shutdown uses redundant ssh to local host

knekrasov:
Solution: reuse the approach from hare-bootstrap script - compare the
hostname with /var/lib/hare/node-name. If the file doesn't exist, then
fallback to older logic.

### DIFF
--- a/utils/utils.py
+++ b/utils/utils.py
@@ -142,9 +142,18 @@ def processes_by_consul_svc_name(cns: Consul) -> Dict[str, List[Process]]:
             from e
 
 
+def get_node_name() -> str:
+    with open('/var/lib/hare/node-name') as f:
+        return f.readline().strip('\n')
+
+
 def is_localhost(hostname: str) -> bool:
-    name = gethostname()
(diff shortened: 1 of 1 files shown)


utils/utils.py
```diff
@@ -142,9 +142,19 @@ def processes_by_consul_svc_name(cns: Consul) -> Dict[str, List[Process]]:
...
+    with open('/var/lib/hare/node-name') as f:
+        return f.readline().strip('\n')
+
+
 def is_localhost(hostname: str) -> bool:
-    name = gethostname()
-    return hostname in ('localhost', '127.0.0.1', name, f'{name}.local')
+    try:
+        node_name = get_node_name()
+        return node_name == hostname
```
mssawant, line 153: return get_node_name() == hostname?

mssawant (approved):
Looks good. Minor comment.

suvratjoshi (approved):

utils/utils.py
knekrasov: Yeah, thanks. Fixed.

PR #1555, opened 2021-03-23
Problem: Some keys used in mini-provisioner are outdated.

SwapnilGaonkar7:
Solution: Modified code to use new updated keys
Also, downloaded sample confstore file from https://seagate-systems.atlassian.net/wiki/spaces/PRIVATECOR/pages/85000402/Mini-Provisioner+ConfStore+Keys and uploaded in our repo
Note: Modified file to include 'durability' and changes value for data and parity


knekrasov (approved):

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -56,51 +56,62 @@ def _create_pool_descriptions(self) -> List[PoolDesc]:
...
+            )
+            spare_units_count = int(
+                conf.get(
+                    f'cluster>{cluster_id}>storage_set[{x}]>durability>spare'))
 
             min_pool_width = data_units_count + parity_units_count \
                 + spare_units_count
-            if (data_devices_count != 0) and not (
-                    data_devices_count >= min_pool_width):
+            if (data_devices_count !=
```
mssawant, line 81: I get flake errors in this case as `line break after an operator` did the build work for you?

SwapnilGaonkar7: Yes, build is passing.

SwapnilGaonkar7: In fact even if we move that operator to next light exact opposite error will be displayed(Line break occurred before a binary operator)

mssawant (approved):

provisioning/miniprov/hare_mp/cdf.py
mssawant: Yeah, make rpm works but if you do make in `provisioning/miniprov` you will see the errors. Just make sure that `make` in `provisioning/miniprov`.

mssawant:
@SwapnilGaonkar7 , please fix the DCO check.

SwapnilGaonkar7:
@mssawant Please check now

PR #1556, opened 2021-03-24
Removed hare_setup.py file

supriyachavan4398:
hare_setup.py was copied as is into hare_mp/main.py. We don't need to maintain duplicated code, so that hare_setup.py must be removed and all other components (like Provisioner) must start using hare_mp/main.py (and thus hare_setup executable) instead.


cortx-admin:

supriyachavan4398:
Some testing stuffs are going on.

PR #1558, opened 2021-03-24
EOS_18888: hare_setup.py is not used

supriyachavan4398:
Signed-off-by: Supriya Chavan <supriya.s.chavan@seagate.com>

EOS_18888: hare_setup.py is redundant and must be removed
Solution :
hare_setup.py was copied as is into hare_mp/main.py. We don't need to maintain duplicated code, 
so that hare_setup.py is removed and all other components (like Provisioner) start using hare_mp/main.py (and thus hare_setup executable) instead.


cortx-admin:

knekrasov (approved):

suvratjoshi:
> Thank you for your pull request and welcome to our community. We could not parse the GitHub identity of the following contributors: **supriyachavan4398**.
> This is most likely caused by a git client misconfiguration; please make sure to:
> 
> 1. check if your git client is configured with an email to sign commits `git config --list | grep email`
> 2. If not, set it up using `git config --global user.email email@example.com`
> 3. Make sure that the git commit email is configured in your GitHub account settings, see https://github.com/settings/emails

Looks like, correct setting of user name and user email is required. If root is being used while git operations, it should not be as per my knowledge.
These commands can be used :
```
$ git config --global user.name “Your Username”
$ git config --global user.email “Your Email ID”
and to view configuration : 
git config --list
```

Issue #1479, opened 2021-02-01
Problem: hare_setup.py is redundant and must be removed


knekrasov:
[`hare_setup.py`](https://github.com/Seagate/cortx-hare/blob/main/provisioning/hare_setup.py) was copied as is into [`hare_mp/main.py`](https://github.com/Seagate/cortx-hare/blob/main/provisioning/miniprov/hare_mp/main.py). We don't need to maintain duplicated code, so that `hare_setup.py` must be removed and all other components (like Provisioner) must start using `hare_mp/main.py` (and thus `hare_setup` executable) instead.

Note: `hare_mp/main.py` appeared as a result of #1467

knekrasov:
We can remove this file safely and we don't need to confirm this change with external teams (as per meeting 3/16/21)

PR #1560, opened 2021-03-25
EOS_18888: hare_setup.py is redundant and must be removed #1479

supriyachavan4398:
Signed-off-by: Supriya Chavan <supriya.s.chavan@seagate.com>
EOS-18888: hare_setup.py is redundant and must be removed #1479
Solution :
hare_setup.py was copied as is into hare_mp/main.py. We don't need to maintain duplicated code, 
so that hare_setup.py is removed and all other components (like Provisioner) start using hare_mp/main.py (and thus hare_setup executable) instead.


cortx-admin:

PR #1561, opened 2021-03-25
EOS-18904 Problem: Replace hardcoded keys for metadata volumes from CDF

SwapnilGaonkar7:
Solution: Modified code to hardcode newer version of metadata volumes for IO services

CDF file after changes,
```
[root@ssc-vm-c-1822 hare]# cat /var/lib/hare/cluster.yaml
pools:
- allowed_failures: null
  name: storage-set-1
  data_units: 1
  parity_units: 0
  disk_refs:
  - path: /dev/sdc
    node: ssc-vm-c-1822.colo.seagate.com
  type: sns
profiles:
- pools:
  - storage-set-1
  name: Profile_the_pool
nodes:
- m0_servers:
  - io_disks:
      data: []
      meta_data: /dev/vg_srvnode-1_md1/lv_raw_md1
    runs_confd: true
  - io_disks:
      data:
      - /dev/sdc
      meta_data: /dev/vg_srvnode-1_md2/lv_raw_md2
    runs_confd: null
  data_iface: eth3
  hostname: ssc-vm-c-1822.colo.seagate.com
  m0_clients:
    other: 3
    s3: 1
  data_iface_type: tcp
```

UT: Tested 'config', 'init' sections of mini-provisioner, also tested bootstrap and shutdown of cluster.


provisioning/miniprov/hare_mp/dhall/gencdf.dhall
```diff
@@ -9,7 +9,8 @@ let NodeInfo =
       , data_iface : Text
       , data_iface_type : Optional T.Protocol
       , io_disks : List Text
-      , meta_data : Text
+      , meta_data1 : Text
```
knekrasov, line 12: @SwapnilGaonkar7 Are you sure that 2 metadata items will suffice? And vice versa, do we think that there will always not less than 2 metadata volumes?

I just think that it makes more sense to have list of metadata volumes. But probably I miss something, please advice.

provisioning/miniprov/hare_mp/templates/hare.config.conf.tmpl.json
```diff
@@ -23,6 +23,7 @@
     "SOME_MACHINE_ID": {
       "cluster_id": "my-cluster",
       "hostname": "myhost",
+      "name": "mynodename",
```
knekrasov, line 26: Please also update create-template.sh script. In fact, this particular template is created by means of that script.

provisioning/miniprov/hare_mp/dhall/gencdf.dhall
SwapnilGaonkar7: Please find below discussion,
**meta_data for first ioservice: /dev/vg_{nodename}_md1/lv_raw_md1
meta_data for secondioservice: /dev/vg_{nodename}_md2/lv_raw_md2
for ex:
/dev/vg_srvnode-1_md1/lv_raw_md1
/dev/vg_srvnode-1_md2/lv_raw_md2**
So, currently number of metadata volumes are equal to number of IO services in particular node and number of IO services are fixed to 2 so 2 metadata items should suffice.

@knekrasov Let me know your input
.

knekrasov: For me that message sounds like something intermittent that can change again quite soon.

My point is that it is simpler and more flexible if from Dhall types level we use list. Then once the number of metadata volumes changes, we'll not need to add meta_data3 or meta_data9.

knekrasov (approved):

mssawant (approved):

PR #1564, opened 2021-03-26
Problem: Sample ConfStore files contain secrets

knekrasov:
JIRA ticket: [EOS-18815](https://jts.seagate.com/browse/EOS-18815)

Solution: replace "secret" values with STUB strings.

### DIFF
--- a/provisioning/miniprov/samples/exampleV1.json
+++ b/provisioning/miniprov/samples/exampleV1.json
@@ -32,7 +32,7 @@
   "corosync-pacemaker": {
     "cluster_name": "cortx_cluster",
     "user": "hacluster",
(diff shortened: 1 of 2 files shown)


provisioning/miniprov/samples/exampleV1.json
```diff
@@ -32,7 +32,7 @@
   "corosync-pacemaker": {
     "cluster_name": "cortx_cluster",
     "user": "hacluster",
-    "secret": "gAAAAABgDsD7pV2BxPXu6NeIIrVCwptUHbZaODtAAgo9Ik778Qz2lgu3xZzJUk7qNlU5ZMHZYq5HKoD500Df359jJ4JZ9z1KcQ=="
+    "secret": "STUB"
```
mssawant, line 35: On second thought, Hare does not read these keys, why not just remove them? we can have just skeletons for all these keys having secrets in them.

knekrasov: Because it is not us who created this example JSON. The less we need to tweak it, the closer it to the "original".

knekrasov:
@johnbent Would you mind to review this patch?

mssawant (approved):

PR #1566, opened 2021-03-26
Problem: Template files not deployed to the proper location

knekrasov:
JIRA ticket: [EOS-19019](https://jts.seagate.com/browse/EOS-19019)

## Solution: 
1. Rename all *.tmpl.json files to *.tmpl
2. Modify `make install` and `make devinstall` targets so that templates and setup.yaml get deployed to /opt/seagate/cortx/hare/conf/


suvratjoshi (approved):

SwapnilGaonkar7:
@knekrasov  Please see below lines from Confluence page,

> Example template files for s3server:
> config: cortx-s3server/s3.config.tmpl.3-node at main · Seagate/cortx-s3server (github.com)
> init: cortx-s3server/s3.init.tmpl.1-node at main · Seagate/cortx-s3server (github.com)
> test: cortx-s3server/s3.test.tmpl.1-node at main · Seagate/cortx-s3server (github.com)
> cleanup: cortx-s3server/s3.cleanup.tmpl.1-node at main · Seagate/cortx-s3server (github.com)

I think we need to follow this which is written in description,

> /opt/seagate/cortx/<comp>/conf/<comp>.post-install.tmpl.XXX
> (XXX: Are variants for 1-node and 3-nodes, resp. e.g. <comp>.post-install.tmpl.1-node)

Is my understanding correct?

knekrasov:
> I think we need to follow this which is written in description,

>>    /opt/seagate/cortx//conf/.post-install.tmpl.XXX
>>    (XXX: Are variants for 1-node and 3-nodes, resp. e.g. .post-install.tmpl.1-node)

> Is my understanding correct?

@SwapnilGaonkar7 I think you're right. But there are two problems:
1. There is no clear explanation why those XXX files are needed and why they should be different. I would prefer to make extra efforts if I understand why is it needed. I raised this question in Confluence as well but got no answer.
   - Per my humble opinion, templates were not designed well. That is also fine ("we all learn") but it doesn't explain why should we maintain so many different files.
   - Examples how "templates" could have been expressed: 
      1. [Dhall](https://dhall-lang.org/#) language. Concentrates on data types and configuration as a result of a function. If we understand template as something with placeholders, Dhall is probably the most suitable thing. And it is format-agnostic: Dhall expressions can be translated into JSON, XML or any other custom format (e.g. [cfgen](https://github.com/Seagate/cortx-hare/blob/main/rfc/3/README.md) uses Dhall to generate confd.xc files as well).
      2. [JSON schema](https://json-schema.org/) or [XSD](https://en.wikipedia.org/wiki/XML_Schema_(W3C)). I understand that ConfStore is designed as "format-agnostic" but it doesn't mean that internally it can't be bound to any specific format. Especially if there is a simple way to validate the structure and declare commitments.
   - Note that any of the options above can easily help to validate the ConfStore files automatically and define expectations and requirements. As far as I understand neither of the options were considered.
    
2. Hare supports the latest set of keys only. Latest set of keys was defined to address the needs of 3-node deployment. Does it mean that our templates must be marked as "3-node" ones? And consequently that will sound as if we are supporting the previous set of keys (while we don't). I wouldn't like to make such a commitment.

cc @s-arya, @NeeravChoudhari, @mssawant

knekrasov:
@SwapnilGaonkar7 your comment is addressed. Please review.

SwapnilGaonkar7 (approved):

PR #1567, opened 2021-03-29
Problem: no 'prepare' command is available

knekrasov:
JIRA ticket: [EOS-18909](https://jts.seagate.com/browse/EOS-18909)

According to the updated mini provisioner requirements, `hare_setup prepare --config $URL` command must be supported by our component.

Solution:
    1. Add prepare command to setup_hare (so that it is available through CLI, but effectively a no-op)
    2. Update setup.yaml to reflect the new command
    3. Fix flake8 issue in mini provisioner code.

### DIFF
--- a/provisioning/miniprov/hare_mp/cdf.py
+++ b/provisioning/miniprov/hare_mp/cdf.py
@@ -78,8 +78,7 @@ class CdfGenerator:
 
             min_pool_width = data_units_count + parity_units_count \
                 + spare_units_count
-            if (data_devices_count !=
-                    0) and not (data_devices_count >= min_pool_width):
+            if data_devices_count and (data_devices_count < min_pool_width):
                 raise RuntimeError('Invalid storage set configuration')
 
             pools.append(
--- a/Makefile
+++ b/Makefile
@@ -476,7 +476,7 @@ uninstall:
 PYTHON_SCRIPTS := utils/hare-shutdown utils/hare-status utils/gen-uuid utils/utils.py
 
 .PHONY: check
-check: check-cfgen check-hax flake8 mypy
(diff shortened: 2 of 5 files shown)


knekrasov:
@SwapnilGaonkar7 I refactored the code you had written, so it would be nice if you can review.

SwapnilGaonkar7 (approved):
Changes looks good.
@knekrasov Should we get 'setup.yaml' changes reviewed from privisioning team?

knekrasov:
@SwapnilGaonkar7 As far as I know we have no process defined with respect to setup.yaml reviews. Per my understanding they will eventually review the current setup.yaml in the main branch.

@s-arya, @mssawant, @suvratjoshi Any thoughts?

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -78,8 +78,7 @@ def _create_pool_descriptions(self) -> List[PoolDesc]:
 
             min_pool_width = data_units_count + parity_units_count \
                 + spare_units_count
-            if (data_devices_count !=
-                    0) and not (data_devices_count >= min_pool_width):
+            if data_devices_count and (data_devices_count < min_pool_width):
```
mssawant, line 81: I think it is good to have an explicit check for integers, `if data_devices_count > 0` in this case. It helps to clearly differentiate a boolean from an integer (I think it is a good practice even though a variable name has `count` in it).

mssawant (approved):
Minor comment, no more comments.

provisioning/miniprov/hare_mp/cdf.py
mssawant: Merging the patch.

PR #1571, opened 2021-04-01
EOS-18936: Provisioning: invoke cfgen in config phase of hare mini provisioner

Shreya-18:
### Description:  
jira ticket- https://jts.seagate.com/browse/EOS-18936

In order to hctl fetch-fids be working on all the nodes of the cluster, consul-kv.json must be present on every node in the cluster. Presently cfgen is invoked in the init phase of the hare mini provisioner which generates consul-kv.json.

**Note:** File : `utils/hare-fetch-fids ` had some linter warnings regarding formatting. **No code changes done.**

### Solution:
1. Invoking cfgen in config phase of hare mini provisioner.
2. hctl fetch-fids works on every node of the cluster post config phase execution of hare mini provisioner.
```
$ hctl fetch-fids -n ssc-vm-c-1813.colo.seagate.com
[
  {
    "name": "confd",
    "fid": "0x7200000000000001:0x9"
  },
  {
    "name": "ioservice",
    "fid": "0x7200000000000001:0xc"
  }
]
```
(Issue observed while running `hctl fetch-fids`, after the config stage. For fetching the **default node-name** (name as specified in the CDF file), we need node-name file . This file gets generated in the init phase. User has to explicitly specify nodename of local node.)
Addressed the issue, and creating the file `node-name`.

### Testing:

1. Ran /opt/seagate/cortx/hare/bin/hare_setup **config** --config
2. checked the files (consul-kv.json, consul-agents.json, confd.dhall ) being generated at `/var/lib/hare`.
3. Ran /opt/seagate/cortx/hare/bin/hare_setup **init**--config 
4. Ran /opt/seagate/cortx/hare/bin/hare_setup **test**--config 
All successfully.

Signed-off-by: Shreya Karmakar <shreya.karmakar@seagate.com>

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -283,11 +283,22 @@ def save(filename: str, contents: str) -> None:
         f.write(contents)
 
 
+def generate_config(url: str, path_to_cdf: str) -> None:
+    conf_dir = '/var/lib/hare'
+    os.environ['PATH'] += os.pathsep + '/opt/seagate/cortx/hare/bin/'
+    cmd = ['cfgen', '-o', conf_dir, path_to_cdf]
+    execute(cmd)
+    conf = ConfStoreProvider(url)
+    hostname = conf.get_hostname()
+    save(f'{conf_dir}/node-name', hostname)
+
+
 def config(args):
     try:
         url = args.config[0]
         filename = args.file[0] or '/var/lib/hare/cluster.yaml'
         save(filename, generate_cdf(url))
+        generate_config(url, filename)
     except Exception as error:
         logging.error('Error performing configuration (%s)', error)
         exit(-1)
--- a/provisioning/miniprov/hare_mp/store.py
+++ b/provisioning/miniprov/hare_mp/store.py
@@ -67,3 +67,8 @@ class ConfStoreProvider(ValueProvider):
         storage_set_id = self._raw_get((f'server_node>{machine_id}>'
                                         f'storage_set_id'))
         return int(storage_set_id)
+
+    def get_hostname(self) -> str:
+        machine_id = self.get_machine_id()
+        hostname = self._raw_get(f'server_node>{machine_id}>hostname')
+        return hostname


mssawant:
Commit message subject is too long, consider
```
EOS-18936: invoke cfgen in config phase of hare mini provisioner
```

provisioning/miniprov/hare_mp/main.py
```diff
@@ -287,11 +287,23 @@ def save(filename: str, contents: str) -> None:
...
+                             shell=False)
+    return process.returncode
+
+
 def config(args):
     try:
         url = args.config[0]
         filename = args.file[0] or '/var/lib/hare/cluster.yaml'
         save(filename, generate_cdf(url))
+        generate_config(filename)
```
mssawant, line 306: Return code is not used. In case of cfgen failure provisioner must error out.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -287,11 +287,23 @@ def save(filename: str, contents: str) -> None:
         f.write(contents)
 
 
+def generate_config(path_to_cdf: str):
```
mssawant, line 290: `def generate_config(path_to_cdf: str) -> int:` please.

mssawant (changes requested):

provisioning/miniprov/hare_mp/main.py
suvratjoshi, line 290: I think if we are creating this function, we need to use it at other places too wherever config is generated using CDF.
What do you think @mssawant and @knekrasov ?
Thanks!

provisioning/miniprov/hare_mp/main.py
knekrasov: That's why I prefer exceptions to error codes :-)

provisioning/miniprov/hare_mp/main.py
knekrasov: I wouldn't say it makes much sense in this case. The fact is that the current file doesn't have any other places where cfgen is executed like this. There is another one - in CdfGenerator class. If we want to refactor cfgen invocation so that both this main.py file and CdfGenerator reuse the same piece of code, then cfgen invocation should probably be refactored into a separate class (I don't like static functions, so let's don't consider that option) e.g. CfgenRunner or something. Anyway, I don't see much benefit. Potentially, it can be helpful if we want to test our business logic without real cfgen (e.g. we want to mock cfgen and emulate its behavior in a predictable and repeatable way). But at this moment we don't have such a need.

We don't save that much lines of code but require extra efforts.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -287,11 +287,23 @@ def save(filename: str, contents: str) -> None:
...
+    conf_dir='/var/lib/hare'
+    os.environ['PATH'] += os.pathsep + '/opt/seagate/cortx/hare/bin/'
+    process = subprocess.run(['cfgen', '-o', conf_dir, path_to_cdf],
```
knekrasov, line 293: There is `execute` function in this file. Why wouldn't you reuse it?

knekrasov (changes requested):

provisioning/miniprov/hare_mp/main.py
Shreya-18: Yes, thanks for pointing it out!

Shreya-18:
> Commit message subject is too long, consider
> 
> ```
> EOS-18936: invoke cfgen in config phase of hare mini provisioner
> ```

Done, updated commit message

provisioning/miniprov/hare_mp/main.py
Shreya-18: Done

provisioning/miniprov/hare_mp/main.py
Shreya-18: Yes, updated the code. Calling `execute() ` function present in this file, in `generate_config()`, which raises exception for failure.

provisioning/miniprov/hare_mp/main.py
Shreya-18: Updated the `generate_config()`, returning None now.

utils/hare-fetch-fids
```diff
@@ -168,23 +166,29 @@ def get_nodes_for_cluster() -> List[Node]:
...
-                   'Default: Local node',
-                   type=str,
-                   action='store')
-    p.add_argument('--all',
-                   help='Returns fids of all the services for nodes in cluster.',
-                   action='store_true')
+    p.add_argument(
+        '--node',
+        '-n',
+        help='node-name - Returns the fids of services for the given node. '
```
mssawant, line 184: @Shreya-18, as you rightly pointed out that `node-name` script reads `/var/lib/hare/node-name` which is created by `utils/hare-bootstrap`, I think anyway without fixing this the patch would not be anyway complete as it will not work as expected. The node-name file is anyway created using node names from `consul-agents.json`.
`consul-agents.json` is created by cfgen. I think we can move creation on `/var/lib/hare/node-name` file as well to config section but instead of using `consul-agents.json` I think let's use conf-store instead. Let's fetch the local node name from conf-store and save it in `/var/lib/hare/node-name`. Let the `/var/lib/hare/node-name` creation in `utils/hare-bootstrap` be as it is as of now.
cc @knekrasov.

utils/hare-fetch-fids
```diff
@@ -19,15 +19,16 @@
...
-import sys
+import argparse
+import json
 import re
+import sys
 from typing import List, NamedTuple, Optional
-import json
-import argparse
+
 import simplejson
```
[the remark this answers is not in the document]
Shreya-18: Actually, I get this linting error

> wrong-import-order: standard import "from typing import List, NamedTuple, Optional" should be placed before "import simplejson"

Still should I fix?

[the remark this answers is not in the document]
knekrasov: @Shreya-18 You can use [`isort`](https://pypi.org/project/isort/) to organize the imports automatically.

If you use vim editor, there is a brilliant [ALE plugin](https://github.com/dense-analysis/ale) that has integration with mypy, flake8 and isort just out of the box.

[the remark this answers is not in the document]
Shreya-18: As per suggestions, `isort` also gives the same sorting order for imports.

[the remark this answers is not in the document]
Shreya-18: Checked out that [PR: 1578](https://github.com/Seagate/cortx-hare/pull/1578/) already fixed this problem.

No changes needed.

utils/hare-fetch-fids
Shreya-18: Yes, it makes sense. Fetching from the `conf-store` file. Please check.

knekrasov:
@Shreya-18 please reword your last commit message. 'task' doesn't look good 😁

Shreya-18:
> @Shreya-18 please reword your last commit message. 'task' doesn't look good 😁

Yes, done. Squashed and updated the commit.

provisioning/miniprov/hare_mp/main.py
Shreya-18: done

knekrasov (approved):

mssawant (approved):

PR #1573, opened 2021-04-01
Problem: init and test mini provisioning phases fail

mssawant:
Because of newly added validation checks in init and test sections of Hare mini
provisioner, some conf-store keys are required in init and test sections too.
Present templates for init and test sections do not contain any keys with respect
to this.

Solution:
- Update init and test section conf-store templates.
- Add more nodes information in 3 node config section template.
- Add populated sample conf-store files for config, init and test sections.

### DIFF
--- a/provisioning/miniprov/hare_mp/templates/hare.config.conf.tmpl.3-node.sample
+++ b/provisioning/miniprov/hare_mp/templates/hare.config.conf.tmpl.3-node.sample
@@ -11,24 +11,24 @@
           "spare": "2",
           "name": "storage1",
           "server_nodes": [
-            "SERVER_NODE_0_MACHINE_ID",
-            "SERVER_NODE_1_MACHINE_ID",
-            "SERVER_NODE_2_MACHINE_ID",
+            "1114a50a6bf6f9c93ebd3c49d07d3fd4",
+            "9ec5de3a8b57493e8fc7bfae67ecd3b3",
+            "846fd26885f8423a8da0626538ed47bc"
           ]
         }
       ]
     }
   },
   "server_node": {
-    "SERVER_NODE_0_MACHINE_ID": {
+    "1114a50a6bf6f9c93ebd3c49d07d3fd4": {
       "cluster_id": "my-cluster",
-      "hostname": "myhost",
-      "name": "mynodename",
+      "hostname": "ssc-vm-1623.colo.seagate.com",
+      "name": "srvnode-1",
       "network": {
         "data": {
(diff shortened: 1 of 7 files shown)


knekrasov (approved):
Although the PR looks good, a few questions if I may.

1. The change is about template files only. Does it mean that somebody already use them for validation? If so, do you think it is possible to introduce the same checks into our tests so that we can be sure that we don't break anything? 
2. `SOME_MACHINE_ID` is replaced with `SERVER_NODE_{0,1,2}_MACHINE_ID`. Does the new name follows some requirement from CI? Can we see the set of conventions and requirements to templates then?
3. You added `.sample` files as well. How they differ from `.tmpl` ones? I just wonder how to maintain so many files properly...

SwapnilGaonkar7 (approved):

mssawant:
>The change is about template files only. Does it mean that somebody already use them for validation? If so, do you think it is possible to introduce the same checks into our tests so that we can be sure that we don't break anything?

I added multiple entries in 3-Node template files just to make it more explicit to provisioner as they will be using it populate. But I think we can maintain the template as per storage set size and need include all the all the nodes in the cluster. Having said that, for larger clusters it is good to get it clarified from provisioning team what they really expect from the template.
I think we can add some test cases for multi-node storage-set, e.g. validating the server_node count vs the layout params or actuall number of nodes.

> SOME_MACHINE_ID is replaced with SERVER_NODE_{0,1,2}_MACHINE_ID. Does the new name follow some requirement from CI? Can we see the set of conventions and requirements to templates then?

No, even I don't see any specific conventions as such. From [conf-store keys](https://seagatetechnology.sharepoint.com/:x:/r/sites/gteamdrv1/tdrive1224/_layouts/15/Doc.aspx?sourcedoc=%7B46F15394-2BA7-4D44-8B7E-67F1AEAEA654%7D&file=LDR%20R2%20Specifications%20-%20Provisioning.xlsx&action=default&mobileredirect=true&cid=9a305dec-c0fe-43e6-bffd-3bd309ff1c74&CID=76c04d49-f373-26cb-dc04-431fd67e7856) it seems to be some UUID. I don't see any specific requirements from CI.

>You added .sample files as well. How they differ from .tmpl ones? I just wonder how to maintain so many files properly...

Sample files are just a populated version of .tmpl files. I used them for my testing. Thought they would be a good reference to QA or anyone who wants to test by replacing hostnames and machine-ids as required. Yes, maintenance will be a nightmare but that is an accepted truth nowadays, no? ;)

PR #1574, opened 2021-04-02
EOS-19018: Pool type is not taken from ConfStore

knekrasov:
Solution:
1. Assume key structure as follows: `cluster>CLUSTER_ID>storage_set[M]>{type}>data`
2. Update tests and templates

### DIFF
--- a/provisioning/miniprov/hare_mp/cdf.py
+++ b/provisioning/miniprov/hare_mp/cdf.py
@@ -6,8 +6,9 @@ from typing import Any, Dict, List, Optional
 import pkg_resources
 
 from hare_mp.store import ValueProvider
-from hare_mp.types import (ClusterDesc, DiskRef, DList, Maybe, NodeDesc,
-                           PoolDesc, ProfileDesc, Protocol, Text)
+from hare_mp.types import (ClusterDesc, DiskRef, DList, Maybe, MissingKeyError,
+                           NodeDesc, PoolDesc, PoolType, ProfileDesc, Protocol,
+                           Text)
 
 DHALL_PATH = '/opt/seagate/cortx/hare/share/cfgen/dhall'
 DHALL_EXE = '/opt/seagate/cortx/hare/bin/dhall'
@@ -47,6 +48,41 @@ class CdfGenerator:
             nodes.append(self._create_node(machine_id))
         return nodes
 
+    def _get_pool_type(self, cluster_id: str, storage_ndx: int) -> PoolType:
+        conf = self.provider
+        for type_name in [p.name for p in PoolType]:
+            type_value = conf.get(
+                f'cluster>{cluster_id}>storage_set[{storage_ndx}]'
+                f'>durability>{type_name}',
+                allow_null=True)
+            if type_value:
+                return PoolType[type_name]
+        raise MissingKeyError(
+            'No pool type found under key '
+            f'cluster>{cluster_id}>storage_set[{storage_ndx}]')
+
+    def _get_pool_property(self, cluster_id: str, pool_type: str,
+                           storage_ndx: int, prop_name: str) -> int:
+        conf = self.provider
+        return int(
+            conf.get(f'cluster>{cluster_id}>storage_set[{storage_ndx}]>'
+                     f'durability>{pool_type}>{prop_name}'))
+
+    def _get_data_units(self, cluster_id: str, pool_type: str,
+                        storage_ndx: int) -> int:
+        return self._get_pool_property(cluster_id, pool_type, storage_ndx,
+                                       'data')
+
+    def _get_parity_units(self, cluster_id: str, pool_type: str,
+                          storage_ndx: int) -> int:
+        return self._get_pool_property(cluster_id, pool_type, storage_ndx,
+                                       'parity')
+
+    def _get_spare_units(self, cluster_id: str, pool_type: str,
+                         storage_ndx: int) -> int:
+        return self._get_pool_property(cluster_id, pool_type, storage_ndx,
+                                       'spare')
+
     def _create_pool_descriptions(self) -> List[PoolDesc]:
         pools: List[PoolDesc] = []
         conf = self.provider
@@ -64,19 +100,22 @@ class CdfGenerator:
                 data_devices_count += len(
                     conf.get(
                         f'server_node>{node}>storage>cvg[0]>data_devices'))
+            pool_type = self._get_pool_type(cluster_id, x)
+            type_str = pool_type.name
 
-            data_units_count = int(conf.get(
-                f'cluster>{cluster_id}>storage_set[{x}]>durability>sns>data'))
-            parity_units_count = int(conf.get(
-                f'cluster>{cluster_id}>'
-                f'storage_set[{x}]>durability>sns>parity'))
-            spare_units_count = int(conf.get(
-                f'cluster>{cluster_id}>storage_set[{x}]>durability>sns>spare'))
+            data_units_count = self._get_data_units(cluster_id, type_str, x)
+            parity_units_count = self._get_parity_units(
+                cluster_id, type_str, x)
+            spare_units_count = self._get_spare_units(cluster_id, type_str, x)
 
             min_pool_width = data_units_count + parity_units_count \
                 + spare_units_count
             if data_devices_count and (data_devices_count < min_pool_width):
-                raise RuntimeError('Invalid storage set configuration')
+                raise RuntimeError(
+                    'Invalid storage set configuration '
+                    f'(name={storage_set_name}):'
+                    f'data_devices_count ({data_devices_count}) must be not '
+                    f'less than N+K+S ({min_pool_width})')
 
             pools.append(
                 PoolDesc(
@@ -96,7 +135,8 @@ class CdfGenerator:
                                 f'storage>cvg[0]>data_devices')
                         ], 'List DiskRef'), 'List DiskRef'),
                     data_units=data_units_count,
-                    parity_units=parity_units_count))
+                    parity_units=parity_units_count,
+                    type=pool_type))
 
         return pools
 
--- a/provisioning/miniprov/hare_mp/cdf.py
+++ b/provisioning/miniprov/hare_mp/cdf.py
@@ -1,18 +1,38 @@
 import os.path as P
 import subprocess as S
+from dataclasses import dataclass
 from string import Template
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 import pkg_resources
 
 from hare_mp.store import ValueProvider
-from hare_mp.types import (ClusterDesc, DiskRef, DList, Maybe, MissingKeyError,
-                           NodeDesc, PoolDesc, PoolType, ProfileDesc, Protocol,
-                           Text)
+from hare_mp.types import (ClusterDesc, DiskRef, DList, Maybe, NodeDesc,
+                           PoolDesc, PoolType, ProfileDesc, Protocol, Text)
 
 DHALL_PATH = '/opt/seagate/cortx/hare/share/cfgen/dhall'
 DHALL_EXE = '/opt/seagate/cortx/hare/bin/dhall'
 DHALL_TO_YAML_EXE = '/opt/seagate/cortx/hare/bin/dhall-to-yaml'
+
+
+@dataclass
+class Layout:
+    data: int
+    parity: int
+    spare: int
+
+
+@dataclass
+class PoolHandle:
+    cluster_id: str
+    pool_type: str
+    storage_ndx: int
+
+    def tuple(self) -> Tuple[str, str, int]:
+        # We could have used dataclasses.astuple() instead here but explicit
+        # implementation is safer (we can be sure that the method is not
+        # broken after somebody has changed the order of fields)
+        return (self.cluster_id, self.pool_type, self.storage_ndx)
 
 
 class CdfGenerator:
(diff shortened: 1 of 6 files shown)


provisioning/miniprov/create-template.sh
```diff
@@ -23,9 +23,9 @@ FILENAME=$1
 URL="json://$(readlink -f "$FILENAME")"
 
 conf "$URL" set 'cluster>my-cluster>site>storage_set_count=1'
-conf "$URL" set 'cluster>my-cluster>storage_set[0]>data=1'
-conf "$URL" set 'cluster>my-cluster>storage_set[0]>parity=0'
-conf "$URL" set 'cluster>my-cluster>storage_set[0]>spare=0'
+conf "$URL" set 'cluster>my-cluster>storage_set[0]>sns>data=1'
```
SwapnilGaonkar7, line 26: According to below doc, actual path is '_cluster>{cluster-id}>storage_set[N]>{type}>durability>data_'.
https://seagate-systems.atlassian.net/wiki/spaces/PRIVATECOR/pages/85000402/Mini-Provisioner+ConfStore+Keys.

@knekrasov I guess you assuming that this will be handled by https://github.com/Seagate/cortx-hare/pull/1576/files. Is is correct?

provisioning/miniprov/create-template.sh
```diff
@@ -23,9 +23,9 @@ FILENAME=$1
...
+conf "$URL" set 'cluster>my-cluster>storage_set[0]>sns>parity=0'
+conf "$URL" set 'cluster>my-cluster>storage_set[0]>sns>spare=0'
```
suvratjoshi, line 28: I think similarly we also need to mention for >dix> as well here and modified references in other files

knekrasov: Then we come up to the question how many options should we show in the templates. We're not given with any requirement on them.

provisioning/miniprov/create-template.sh
knekrasov: Yes, I would prefer to keep the changes local. If your patch addresses missing "durability" key AND if my patch is landed before yours, then your patch should be rebased and updated in this place also.

provisioning/miniprov/create-template.sh
suvratjoshi: ok
I think here is what I understand from today's call:
For SNS (w.r.t. https://jts.seagate.com/browse/EOS-16763, but I am not sure if this is correct to refer here): 
data_units: 4
parity_units: 2
spare_units: 2

For DIX (https://jts.seagate.com/browse/EOS-18999 after knowing this requirement from Madhav): 
data_units: 1
parity_units: 1
spare_units: 1

For MD : My understanding from today's call is that, MD pool is going to be obselete soon. So, this PR wise, I think we can just have that pool type enum MD maintained, but we need not populate any value for MD pool now and we need to assume that it will not be there in ConfStore for now.

But please somebody confirm on my my understanding.
cc: @mssawant @s-arya

knekrasov: Addressed in the new commit. Please review @suvratjoshi

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -64,22 +100,22 @@ def _create_pool_descriptions(self) -> List[PoolDesc]:
                 data_devices_count += len(
                     conf.get(
                         f'server_node>{node}>storage>cvg[0]>data_devices'))
+            pool_type = self._get_pool_type(cluster_id, x)
```
mssawant, line 103: @knekrasov, How about instead of extracting a pool_type from conf-store, we iterate over supported pool types and fetch the information for the ones available in the conf-store?
SNS pool is a must have, an exception must be raised is not present. 
DIX is also required as per the latest discussions but for now let's not have it as a mandatory pool type. If none of the pool types are present, an exception is raised.
What do you think?

knekrasov: @mssawant If you look into `_get_pool_type()`, its algorithm is quite the same to what you described.

provisioning/miniprov/hare_mp/templates/hare.init.conf.tmpl.1-node
```diff
@@ -7,9 +7,11 @@
       "storage_set": [
         {
           "durability": {
-            "data": "1",
-            "parity": "0",
-            "spare": "0"
+            "sns": {
```
suvratjoshi, line 10: I think for 1-node, 1,0,0 needs to be maintained

provisioning/miniprov/create-template.sh
suvratjoshi: Not sure if I am missing something.
I was thinking, along with updates to values of layout N,K,S we also need to have >dix> too similar to sns.

provisioning/miniprov/create-template.sh
```diff
@@ -23,9 +23,9 @@ FILENAME=$1
 URL="json://$(readlink -f "$FILENAME")"
 
 conf "$URL" set 'cluster>my-cluster>site>storage_set_count=1'
-conf "$URL" set 'cluster>my-cluster>storage_set[0]>durability>data=1'
-conf "$URL" set 'cluster>my-cluster>storage_set[0]>durability>parity=0'
-conf "$URL" set 'cluster>my-cluster>storage_set[0]>durability>spare=0'
+conf "$URL" set 'cluster>my-cluster>storage_set[0]>durability>sns>data=4'
```
suvratjoshi, line 26: With this change, is 1-node provisioner going to work in the same way as it is working currently and these values will get reflected only for 3-node?

provisioning/miniprov/test/test_cdf.py
```diff
@@ -255,19 +345,26 @@ def ret_values(value: str) -> Any:
...
-                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set[0]>durability>parity': 0
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>site>storage_set_count':
+                1,
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set>server_node_count':
+                1,
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set[0]>name':
+                'StorageSet-1',
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set[0]>server_nodes':
+                ['srvnode_1'],
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set[0]>durability>data':
```
SwapnilGaonkar7, line 364: Should this path contain pooltype for e.g sns?

provisioning/miniprov/test/test_cdf.py
```diff
@@ -290,30 +387,40 @@ def ret_values(value: str) -> Any:
...
-                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set[0]>durability>data': 1,
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>site>storage_set_count':
+                1,
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set>server_node_count':
+                2,
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set[0]>name':
+                'StorageSet-1',
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set[0]>server_nodes':
+                ['srvnode_1', 'srvnode_2'],
+                'cluster>92f444df-87cc-4137-b680-aab3b35d1695>storage_set[0]>durability>data':
```
SwapnilGaonkar7, line 398: Is pooltype(sns) needed here?

suvratjoshi:
To my understanding, here are the main points :

1. Current 1-node miniprovisioner working should not get affected with our new changes
2. Currently spare_units JIRA EOS-16345 is not merged yet so, nothing should break even if spare_units functionality is not incorporated yet in our cfgen.
3. Do we need to update our conf-store?
4. New values for 3-node miniprovisioner as per my understanding are needed to be : 
For SNS (w.r.t. https://jts.seagate.com/browse/EOS-16763, but I am not sure if this is correct to refer here):
data_units: 4
parity_units: 2
spare_units: 2

For DIX (https://jts.seagate.com/browse/EOS-18999 after knowing this requirement from Madhav):
data_units: 1
parity_units: 1
spare_units: 1

5. I am not yet sure if main database of ConfStore contains layout configuration for both SNS and DIX (with pool type). At least not visible here : https://seagate-systems.atlassian.net/wiki/spaces/PRIVATECOR/pages/85000402/Mini-Provisioner+ConfStore+Keys

provisioning/miniprov/hare_mp/templates/hare.init.conf.tmpl.1-node
SwapnilGaonkar7: Agree, since only 2 data devices are there

provisioning/miniprov/test/test_cdf.py
knekrasov: It seems like this test doesn't go through pool pool creation that's why it can be even ignored in the test.

SwapnilGaonkar7: Ok, Got it.

knekrasov:
>     1. I am not yet sure if main database of ConfStore contains layout configuration for both SNS and DIX (with pool type). At least not visible here : https://seagate-systems.atlassian.net/wiki/spaces/PRIVATECOR/pages/85000402/Mini-Provisioner+ConfStore+Keys

@suvratjoshi Do you think that any of the problems should be addressed by this PR? If so, please list the ones which do you think are critical for this PR.

ConfStore structure will be dynamic quite long, I don't see a reason trying to polish everything up to perfect state today (because tomorrow the ideal solution would become invalid again). Let's concentrate on one single problem and address the single problem in a single PR.

suvratjoshi:
> @suvratjoshi Do you think that any of the problems should be addressed by this PR? If so, please list the ones which do you think are critical for this PR.
Ok. As we discussed over call, we only need to make changes for 1-node template files (for restoring to 1,0,0 layout) and we probably need to create a different JIRA ticket for have 3-node ConfStore .json file.
Thanks!

provisioning/miniprov/create-template.sh
knekrasov: This is discussed during 1:1 phone call with Suvrat. 
In short, 
1. This file is not used by any external team
2. It is used from time to time to re-generate some template files.
So it is not a problem that this file has some discrepancies. At the times when it was created there were no separate 3-node and 1-node templates. In order to address the issue properly, we need to understand whether 4,2,2 numbers should be some VAR_TMPL placeholders (given it is a template). Anyway it is not a must for this PR.

provisioning/miniprov/hare_mp/templates/hare.init.conf.tmpl.1-node
knekrasov: @suvratjoshi Addressed. Please see the updated patch.

suvratjoshi (approved):

provisioning/miniprov/hare_mp/cdf.py
mssawant: Not really, because there can be multiple pool types in a storage set, e.g. SNS, DIX, _get_pool_type() returns just 1 pool type per storage_set. We need to iterate over the pool_types and get the information for all the available pool_types in a  given storage set.

knekrasov: According to @ujjwalpl storage set = pool. One pool can have only one type.

Even if we imagine that a storage set that has types, say sns and dix, then we possibly can understand it as "I need to create two pools". But how will we map which drive will belong to which pool?

mssawant: So, there are data devices and metadata devices. Data devices belong to data pool, SNS. Metadata devices belong to metadata pool, in this case dix. SNS layout parameters, data, parity and spare apply only to the SNS pool. Important property of the layout parameters is that the total number of corresponding devices in the pool need to be >= data + parity + spare.
Thus SNS layout parameters cannot apply to DIX pool (comprising of metadata devices) in a storage set. Thus we have different layout parameters defined for DIX pool type corresponding to the metadata devices in the pool.
Presently the default layout parameters for DIX pool are 1 + 0 + 0 in Motr.
Thus, storage set (pool) comprises of 2 pools, data and metadata. Both will have different layout parameters.

knekrasov: Thanks @mssawant for explanation, now the picture looks pretty different to me.

knekrasov: @mssawant Your comment is addressed, please see.

@SwapnilGaonkar7 In the last two commits I refactored your code a bit, so it would be nice if you can review.

provisioning/miniprov/hare_mp/types.py
```diff
@@ -87,22 +92,22 @@ class M0Clients(DhallTuple):
...
 
 
 @dataclass(repr=False)
 class NodeDesc(DhallTuple):
     hostname: Text
     data_iface: Text
-    data_iface_type: Maybe  # [Protocol]
-    io_disks: DList  # [str]
+    data_iface_type: Maybe[Protocol]
+    io_disks: DList[Text]
```
azheregelya, line 110: Is `Text` really needed instead of `str`?  :)
From documentation i understood that it is just for compatibility with Python 2.

knekrasov: Yes, it is needed.

The fact is that Text, NodeDesc, Protocol and other guys are a kind of AST objects. Their main aim is to stringify themself into Dhall expressions. Use case: generate CDF file according to some values in ConfStore.  In order to generate a valid (and type safe) CDF file, we reuse Dhall types. In order to generate Dhall expressions we use objects like this.

So the main problem is that Dhall strings are enclosed into double quotes (`"like this"`). While str() for any string in Python will yield a string within single quotes. Even in Python 3

```

Python 3.6.8 (default, Aug  7 2019, 17:28:10)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-39)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> t = 'Some text goes here'
>>> str(t)
'Some text goes here'
>>>
```

azheregelya: So, I we create our own class whose `__str__` implementation prints double-quotes, right? 
I still don't understand how usage of  `Text` type hint helps in this situation :(

knekrasov: Exactly. Quote from the tests:

```python
    def test_m0server_with_disks(self):
        val = M0ServerDesc(
            runs_confd=Maybe(True, 'Bool'),
            io_disks=DisksDesc(
                meta_data=Maybe(None, 'Text'),
                data=DList([Text('/disk1'), Text('/disk2')], 'test')))
        self.assertEqual(
            '{ runs_confd = Some (True), io_disks = { meta_data = None Text, data = ["/disk1", "/disk2"] } }',
            str(val))

```

Note how `data = [ ... ]` array looks like in the assert.

azheregelya: Ah, now i got it. I mixed it with https://docs.python.org/3/library/typing.html#typing.Text
Didn't notice separate Text implementation :)

provisioning/miniprov/hare_mp/cdf.py
mssawant: @knekrasov , tried your branch, added following dix pool layout params,
```
          "durability": {
            "sns": {
              "data": "4",
              "parity": "2",
              "spare": "2"
            },
            "dix": {
              "data": "1",
              "parity": "1",
              "spare": "1"
            }
          },
```
getting error as,
```
[root@ssc-vm-1623:root] /opt/seagate/cortx/hare/bin/hare_setup config --config 'json:///opt/seagate/cortx/hare/conf/hare.config.conf.tmpl.3-node.sample' --file /root/config_cluster.cdf
Error performing configuration (Command ['cfgen', '-o', '/var/lib/hare', '/root/config_cluster.cdf'] exited with error code 1. Command output: Traceback (most recent call last):
  File "/opt/seagate/cortx/hare/bin/cfgen", line 1775, in <module>
    main()
  File "/opt/seagate/cortx/hare/bin/cfgen", line 133, in main
    validate_cluster_desc(cluster_desc)
  File "/opt/seagate/cortx/hare/bin/cfgen", line 245, in validate_cluster_desc
    validate_pools_desc(desc['pools'])
  File "/opt/seagate/cortx/hare/bin/cfgen", line 349, in validate_pools_desc
    f"{name}: disk_refs are only supported for pools of 'sns' type"
AssertionError: storage1__dix: disk_refs are only supported for pools of 'sns' type
)
```
with just sns things are working fine.

knekrasov: Hm, cfgen code has interesting `XXX` comment:
```python
        if 'disk_refs' not in pool:
            continue

        # XXX Do we want to support disk_refs for DIX and MD pools as well?
        assert t is PoolT.sns, \
            f"{name}: disk_refs are only supported for pools of 'sns' type"
        disk_refs = pool['disk_refs']
        # Allowing diskrefs to be empty until the CDF generation is fixed.
        if disk_refs:
            assert disk_refs, \
                f'Pool {name!r}: disk_refs must not be empty' \
                ' (it can be omitted though)'
            assert all_unique(repr(x) for x in disk_refs), \
                f'Pool {name!r}: disk_refs must be unique'
```

@mssawant Are you sure that DIX pool can (and should) have disk refs?
cc @s-arya, @suvratjoshi 

In the meantime I'll add one more commit to unblock this behaviour in cfgen...

knekrasov: Commit that unblocks DIX in cfgen has been added. I tested the change locally at a single node, bootstrap with dix and sns pool in the profile was successful. @mssawant please review.

mssawant: >@mssawant Are you sure that DIX pool can (and should) have disk refs?

I don't see a reason why should we restrict disk_refs for DIX pools as we have metadata devices.

mssawant: Yep, works, verified.

mssawant (approved):

mssawant:
@knekrasov, the branch has some conflicts, kindly resolve the same.
Also, don't we need to add dix pooltype to tests as well?

provisioning/miniprov/test/test_cdf.py
knekrasov: Addressed: unused keys were removed from tests.

knekrasov:
@mssawant I have resolved all the conflicts.

> Also, don't we need to add dix pooltype to tests as well?

There are at least two tests that validate dix-related functionality:
1. `test_both_dix_and_sns_pools_can_exist`
2. `test_dix_pool_uses_metadata_devices`

As a result 
1. we know that CdfGenerator doesn't fail when it sees "dix" pool
2. we know that CdfGenerator uses metadata devices in case of "dix" pool
3. we know that the one and the same storage set can have "sns" and "dix" pools at the same time

Do you think we need to test anything else?

mssawant:
@knekrasov, No,
```
1. test_both_dix_and_sns_pools_can_exist
2. test_dix_pool_uses_metadata_devices
```
should suffice, for some reason the diff for testcdf.py did not load automatically, I missed that.

PR #1570, opened 2021-04-01
EOS-19092 Problem : Makefile doesn't run linters against some of utils/hare-* scripts

supriyachavan4398:
Solution : The list of python scripts in utils/ is a kind of hard-coded in Makefile. So once somebody adds a script, the script will most probably be forgotten and not added to PYTHON_SCRIPTS in the makefile. This is error-prone.
So we try to make Makefile to fetch the filenames automatically or make that list of scripts be a kind of dynamic.

Signed-off-by: Supriya Chavan <supriya.s.chavan@seagate.com>

PR #1577, opened 2021-04-05
Revert "EOS-19092 Problem : Makefile doesn't run linters against some of utils/hare-* scripts"

mssawant:
Reverts Seagate/cortx-hare#1570

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -484,7 +484,7 @@ uninstall:
 # Linters --------------------------------------------- {{{1
 #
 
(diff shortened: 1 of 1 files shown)


knekrasov:
@mssawant Why did you revert the PR? So now we'll have to re-apply #1570. While the actual reason is that PR works and finds a problem in the code.

mssawant:
Jenkins build was failing in context of this PR. Yes the PR works as expected but the local testing of this PR did not find this and was revealed in Jenkins. Thus even if PR might be working, we are not sure if we need to anything else until the builds work fine in context of this PR.

knekrasov:
@mssawant Please review #1578

PR #1585, opened 2021-04-09
EOS-19399: Create design RFC for AUX multipools layout generation

suvratjoshi:
Need to create design document RFC, for generation of layout parameters for auxiliary (AUX) pools, during multipools cluster bootstrap.

-----
[View rendered rfc/22/README.md](https://github.com/suvratjoshi/cortx-hare/blob/e19399/rfc/22/README.md)
[View rendered rfc/README.md](https://github.com/suvratjoshi/cortx-hare/blob/e19399/rfc/README.md)

PR #1588, opened 2021-04-14
EOS-19399: Create design RFC for AUX multipools layout generation

suvratjoshi:
Need to create design document RFC, for generation of layout parameters
for auxiliary (AUX) pools, during multipools cluster bootstrap.
This PR is used to update the RFC indexes only.
Detailed document is published at confluence at link :
https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/184746420/23+MPAUX+Multipool+Auxiliary+Layout+Parameters+Generation
It addresses the comments from PR:
https://github.com/Seagate/cortx-hare/pull/1585/

-----
[View rendered rfc/README.md](https://github.com/suvratjoshi/cortx-hare/blob/new_e19399/rfc/README.md)

### DIFF
--- a/rfc/README.md
+++ b/rfc/README.md
@@ -23,6 +23,8 @@ processes.
   * [17/CICD](17/README.md) — Continuous Integration and Continuous Deployment
   * [19/EVERULES](19/README.md) — Events and Rules
   * [21/EVERULES](21/README.md) — HA Event Subscription
+  * [22/PARHAX](https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/178421786/22+PARHAX+Parallel+message+processing+in+HaX) - Parallel message processing in HaX
+  * [23/MPAUX](https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/184746420/23+MPAUX+Multipool+Auxiliary+Layout+Parameters+Generation) - Multipool Auxiliary Layout Parameters Generation
 * Draft
   * [4/KV](4/README.md) — Consul KV Schema
   * [5/HAX](5/README.md) — HAlink eXchange


mssawant:
@suvratjoshi, as discussed let's have the rfc in github and have a page in confluence with the github link.

suvratjoshi:
@knekrasov had some comments on confluence page. I will first try to answer those here to confirm answers from @mssawant so that I can update the confluence page accordingly : 
@mssawant, can you please help to get /correct answers (in case mine are wrong) to queries posted by @knekrasov .

[KN] After I read the document, I still don’t quite understand:
[KN] 1. Does motr identify auxiliary pool just automatically? Or do we need to use special naming or types for that?
[SJ] To my understanding, yes motr identifies it automatically. No special naming convention needs be followed.

[KN] 2. It sounds like the only task for Hare is to generate all probable subsets for every given pool.
[SJ] Yes. Correct.

[KN] 3. If [2] is correct, then we probably need to clarify:

[KN] 3a. How to calculate “allowed_failures” vectors for each such a generated aux pool. I guess it will not be equal to the source “allowed_failures”, right? A short formula or a schematic algorithm is highly appreciated.
[SJ] If I undersood this question correctly - the generated AUX pool itself comes into picture after disk failures, so again no additional allowed failures will be "allowed" than the original ones.

[KN] 3b. What pools need to be embellished with auxiliary pools? All pools that we have in CDF? Or mabye all “sns” pools in CDF? Else?
[SJ] I think creating AUX pools will be applicable for SNS pools only and not DIX. (Probable reason would be data_units allowed currently are not more than 1). AUX pools will be generated for all SNS pools.

[KN] 3c. Do we want to generate aux pools just always? Or maybe we’ll want to disable this generation by some flag in cfgen?
[SJ] Yes, always generated and no flag control.

knekrasov (approved):
Good enough. Thanks!

mssawant (approved):

PR #1591, opened 2021-04-15
Problem: private data interface template variable names mismatch

mssawant:
Private data interface names for node 1 in the 3-node conf store template files
are not the same as for other nodes.

Solution:
Match the private data interface variable name for node 1 with other nodes.


SwapnilGaonkar7:
In 'hare.init.conf.tmpl.3-node', we need to replace 'TEMPL_DATA_INTERFACE_TYPE' with 'TMPL_DATA_INTERFACE_TYPE'.
Can we address that in this PR?

knekrasov (approved):

mssawant:
@SwapnilGaonkar7 , good catch!, fixed.

suvratjoshi (approved):

PR #1593, opened 2021-04-19
EOS-19534 Consul Hare Agent to be started by Hare in Mini-Provisioner Init Phase

SwapnilGaonkar7:
Solution: Modified code to skip consul stop during shutdown, also note that consul will be restarted during bootstrap

### DIFF
--- a/utils/hare-bootstrap
+++ b/utils/hare-bootstrap
@@ -209,8 +209,8 @@ $debug_p && set -x
 
 cdf=${1:-}
 
-if sudo systemctl --quiet is-active hare-consul-agent; then
-    die 'hare-consul-agent is active ==> cluster is already running'
+if sudo systemctl --quiet is-active hare-hax; then
+    die 'hare-hax is active ==> cluster is already running'
 fi
 
 if [[ -z $conf_dir ]]; then
@@ -265,6 +265,26 @@ Bootstrap should be executed on a Consul server node.
 (cluster description file).
 EOF
     exit 1
+fi
+
+if sudo systemctl --quiet is-active hare-consul-agent; then
+    say 'Stopping Consul on this node...'
+    sudo systemctl stop hare-consul-agent
+    echo ' OK'
+
+    say 'Stopping Consul on other nodes...'
+    pids=()
+    while read node bind_ip; do
+        ssh $node "sudo systemctl stop hare-consul-agent" &
+        pids+=($!)
+    done < <(get_server_nodes | grep -vw $(node-name) || true)
+
+    while read node bind_ip; do
+        ssh $node "sudo systemctl stop hare-consul-agent" &
+        pids+=($!)
+    done < <(get_client_nodes)
+    wait4 ${pids[@]-}
+    echo ' OK'
 fi
 
 say 'Starting Consul server on this node...'
--- a/utils/hare-shutdown
+++ b/utils/hare-shutdown
@@ -55,13 +55,10 @@ def _get_processes_for_node() -> Dict[str, List[utils.Process]]:
     return processes
 
 
-def stop_process(processes: Dict[str, List[utils.Process]]):
+def stop_all_with_consul(processes: Dict[str, List[utils.Process]],
+                         leader_node: str):
     for svc in utils.shutdown_sequence:
-        if svc not in processes:
-            continue
-        if svc != 'consul':
-            utils.stop_parallel([proc for proc in processes[svc]])
-        else:
+        if svc in processes and svc == 'consul':
             # Consul process at the current node must stopped the last.
             # Reason: if something goes wrong and the user will want to
             # shutdown the cluster again, local Consul must be available.
@@ -72,17 +69,28 @@ def stop_process(processes: Dict[str, List[utils.Process]]):
             local_procs = [proc for proc in processes[svc] if proc.is_local]
             utils.stop_parallel(local_procs)
 
-
-def shutdown_cluster():
-    check_cluster()
-    (processes, leader_node) = _get_processes_for_cluster()
-    stop_process(processes)
-
     if leader_node and not utils.is_fake_leader_name(leader_node):
         logging.info(f'Shutting down RC Leader at {leader_node}... ')
         pkill: str = 'sudo pkill --exact -KILL consul &> /dev/null'
         utils.exec_custom(utils.ssh_prefix(leader_node) + pkill,
                           show_err=False)
+
+
+def stop_process(processes: Dict[str, List[utils.Process]]):
+    for svc in utils.shutdown_sequence:
+        if svc not in processes:
+            continue
+        if svc != 'consul':
+            utils.stop_parallel([proc for proc in processes[svc]])
+
+
+def shutdown_cluster(skip_consul_stop: bool):
+    check_cluster()
+    (processes, leader_node) = _get_processes_for_cluster()
+    stop_process(processes)
+
+    if not skip_consul_stop:
+        stop_all_with_consul(processes, leader_node)
 
 
 def node_shutdown():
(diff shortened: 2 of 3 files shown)


utils/hare-bootstrap
```diff
@@ -177,7 +178,7 @@ is_localhost() {
 # main
 
 TEMP=$(getopt --options hc: \
-              --longoptions help,mkfs,conf-dir:,debug \
+              --longoptions help,mkfs,skip_consul_start,conf-dir:,debug \
```
knekrasov, line 181: Optional: `s/skip_consul_start/skip-consul-start/`

utils/hare-shutdown
```diff
@@ -55,13 +55,14 @@ def _get_processes_for_node() -> Dict[str, List[utils.Process]]:
...
-def stop_process(processes: Dict[str, List[utils.Process]]):
+def stop_process(processes: Dict[str, List[utils.Process]],
+                 skip_consul_stop: bool):
     for svc in utils.shutdown_sequence:
         if svc not in processes:
             continue
         if svc != 'consul':
             utils.stop_parallel([proc for proc in processes[svc]])
-        else:
+        elif skip_consul_stop == False:
```
knekrasov, line 65: `eilf not skip_consul_stop:`

knekrasov: Although I would prefer using 2 different [strategies](https://en.wikipedia.org/wiki/Strategy_pattern): one works as it used to, another one never stops consul.
The code of each strategy will be quite simple, the decision whether consul should be stopped or not is taken just once, at the very beginning. So lots of code will not need to propagate this `skip_consul_stop` flag and check it every now and then.

But again, if this is a draft and temporarily solution, please ignore this comment.

utils/hare-bootstrap
```diff
@@ -209,8 +212,8 @@ $debug_p && set -x
 
 cdf=${1:-}
 
-if sudo systemctl --quiet is-active hare-consul-agent; then
```
knekrasov, line 212: Isn't it safer just create a copy of `hare-bootstrap` script and modify it - so the new one will be used by Mini Provisioner, but we'll not break the existing behavior for debugging and testing?

@mssawant what do you think?

utils/hare-bootstrap
```diff
@@ -40,6 +40,7 @@ Positional arguments:
 Options:
   --debug      Print commands and their arguments as they are executed.
   --mkfs       Execute m0mkfs.  *CAUTION* This wipes all Motr data!
+  --skip_consul_start   Consul will not be started and it is asuumed that it is already running
```
mssawant, line 43: s/asuumed/assumed please

utils/hare-bootstrap
mssawant: And you mean we invoke the new one if `skip-consul-start` is specified to `hctl` interface? Although not for long but we will have to maintain multiple scripts and make sure that changes to one, if any, are incorporated in to other one too. On the other hand it will be cleaner to have a separate script.

utils/hare-bootstrap
```diff
@@ -288,39 +293,41 @@ jq '[.[] | {key, value: (.value | @base64)}]' < $conf_dir/consul-kv.json |
...
-        exit 1
-    fi
-    echo -n '.'
-    sleep 1
-    (( count++ ))
-done
-echo ' OK'
+    # Waiting for the Consul agents to get ready...
+    count=1
+    while (( $(get_ready_agents_nr) != $agents_nr )); do
```
mssawant, line 317: Let's have this check unconditional to `skip_consul_start` value. Because even though Consul processes are running we need to make sure that all are running if not the we must error out.

utils/hare-bootstrap
```diff
@@ -288,39 +293,41 @@ jq '[.[] | {key, value: (.value | @base64)}]' < $conf_dir/consul-kv.json |
     consul kv import - > /dev/null
 echo ' OK'
 
-say 'Starting Consul on other nodes...'
-pids=()
-while read node bind_ip; do
-    ssh $node "PATH=$PATH $(which mk-consul-env) \
-                          --mode server --bind $bind_ip --join $join_ip &&
-               sudo systemctl start hare-consul-agent" &
+if [ "$skip_consul_start" = false ] ; then
```
mssawant, line 296: Now the question is if for some reason if the running Consul agent was stopped then who would restart it?
I would suggest the following,
- Continue iterating over all the nodes
- check if Consul is already running, if not start the Consul agent.

I think we also need to do this for node-start.

mssawant (changes requested):

utils/hare-bootstrap
knekrasov: > And you mean we invoke the new one if skip-consul-start is specified to hctl interface? 

Not necessary. The script name will be different, so probably we'll not even need a flag.

mssawant: I am thinking how would we invoke that different script, do you mean we just move to the new script and `hctl bootstrap` command invokes only the new script? or do you suggest to have a new `hctl command` altogether that invokes the new script?

knekrasov: No. We can leave `hare-bootstrap` script as is so everybody who have been using `hctl bootstrap` will still get just the same behavior they would expect.

But we can copy `hare-bootstrap` -> `hare-provision` and put @SwapnilGaonkar7's changes there. Then it will be invoked like this: `hctl provision --mkfs /my/cdf/file.yaml`

The latter script will not start Consul.

Thoughts?

utils/hare-shutdown
```diff
@@ -73,16 +74,17 @@ def stop_process(processes: Dict[str, List[utils.Process]]):
...
     (processes, leader_node) = _get_processes_for_cluster()
-    stop_process(processes)
+    stop_process(processes, skip_consul_stop)
 
-    if leader_node and not utils.is_fake_leader_name(leader_node):
-        logging.info(f'Shutting down RC Leader at {leader_node}... ')
-        pkill: str = 'sudo pkill --exact -KILL consul &> /dev/null'
-        utils.exec_custom(utils.ssh_prefix(leader_node) + pkill,
-                          show_err=False)
+    if skip_consul_stop == False:
```
mssawant, line 82: As @knekrasov suggested, if we take out the consul shutdown code into a separate function, may be `stop_consul()` then we need not pass the flag to `shutdown_cluster()` in `main()` itself we can check `if skip_consul_stop` and invoke `stop_consul()` or not.

SwapnilGaonkar7:
@mssawant @knekrasov I have modified code by just skipping consul stop during hctl shutdown, I have also addressed review comments which were valid after my changes.
Please have a look

utils/hare-bootstrap
mssawant: okay, so you are suggesting separate `hctl command`. I wonder if it would be confusing for the users of `hctl bootstrap`, we would need to document it well. In turn a flag like `skip-consul-start` seems more intuitive.
I think @SwapnilGaonkar7 is testing a "no change" option to hare-bootstrap (i.e. without `skip-consul-start`) May be we can just run `systemctl start hare-consul-agent` un-conditionally and handle the return code from systemd in case `hare-consul-agent` is already running.

utils/hare-shutdown
```diff
@@ -55,13 +55,9 @@ def _get_processes_for_node() -> Dict[str, List[utils.Process]]:
     return processes
 
 
-def stop_process(processes: Dict[str, List[utils.Process]]):
+def stop_consul(processes: Dict[str, List[utils.Process]], leader_node: str):
```
knekrasov, line 58: For me this name is misleading: we're stopping not only consul but all the processes.

`stop_all_with_consul()` maybe?

PS: that's why I like classes and overriden methods. Although the class name would be different (e.g. MiniProvFriendlyStopper and CommonStopper) but the main method would be named `stop()` in both cases 😀

knekrasov (approved):
Minor comment only.

mssawant (approved):

utils/hare-shutdown
SwapnilGaonkar7: Done

gowthamchinna:
run vm-deployment

gowthamchinna:
test vm-deployment

gowthamchinna:
test this please

PR #1594, opened 2021-04-19
EOS: 18914

supritshinde:
Issue: In a Cluster, Hare doesn't show for which node access is being denied while writing the node-name to the conf-dir.
Solution: Printng  out name of node where permission is denied for /var/lib/hare/node-name.

### DIFF
--- a/utils/hare-bootstrap
+++ b/utils/hare-bootstrap
@@ -242,6 +242,10 @@ EOF
     while read node _; do
         if is_localhost $node; then
             echo $node > $conf_dir/node-name
+            if [ $? -ne 0 ]
+                then
+                    echo "Permission Denied at" $node "for /var/lib/hare/node-name"
+            fi
         else
             # Redirect stdin to /dev/null (`-n`) to prevent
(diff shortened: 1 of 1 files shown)


utils/hare-bootstrap
```diff
@@ -242,6 +242,10 @@ EOF
     while read node _; do
         if is_localhost $node; then
             echo $node > $conf_dir/node-name
+            if [ $? -ne 0 ]
+                then
+                    echo "Permission Denied at" $node "for /var/lib/hare/node-name"
```
knekrasov, line 247: Sorry, but it seems to me that this code will never be executed just because of `set -e` setting at the very beginning. If echo at previous line fails to write the data, the script will exit immediately with non-zero code.

supritshinde: Um yeah. I just saw it. Yes this segment won't be executed.  Infact  if any line that fails, script will exit. Need to think of something else.

supritshinde: Closing this PR. It was a Blunder. Sorry for the inconvenience caused.

PR #1596, opened 2021-04-20
EOS 18914: The node for which conf_dir access is denied is unknown

supritshinde:
Solution: Displaying the node-name of those nodes for which /var/lib/hare/node-name is not accessible

### DIFF
--- a/utils/hare-bootstrap
+++ b/utils/hare-bootstrap
@@ -241,11 +241,17 @@ EOF
 
     while read node _; do
         if is_localhost $node; then
-            echo $node > $conf_dir/node-name
+            echo $node > $conf_dir/node-name || {
+                echo "Permission Denied at $node for $conf_dir/node-name"
+                exit 1
+            }
         else
             # Redirect stdin to /dev/null (`-n`) to prevent
             # accidental stealing of `get_all_nodes` output.
-	    ssh -n $node "echo $node > $conf_dir/node-name"
+	    ssh -n $node "echo $node > $conf_dir/node-name" || {
+                echo "Permission Denied at $node for $conf_dir/node-name"
+                exit 1
+            } 
         fi
     done < <(get_all_nodes)
 


knekrasov (commented):
I think the problem is more general: if _anything_ goes wrong, hare-bootstrap script will not help to understand at which node the problem appeared.

@andriytk is that correct?

In that case some more general solution is required.

mssawant:
The patch seems to solve the said problem though by displaying the node name but in general the actual problem can be something else and this can one of the symptoms of it. But just in case the problem is seen only during Hare deployment then I think the error message looks good. @supritshinde , can you post some data from the actual crime scene?

mssawant (approved):

utils/hare-bootstrap
```diff
@@ -241,11 +241,17 @@ EOF
...
+            echo $node > $conf_dir/node-name || {
+                echo "Permission Denied at $node for $conf_dir/node-name"
+                exit 1
+            }
         else
             # Redirect stdin to /dev/null (`-n`) to prevent
             # accidental stealing of `get_all_nodes` output.
-	    ssh -n $node "echo $node > $conf_dir/node-name"
+	    ssh -n $node "echo $node > $conf_dir/node-name" || {
+                echo "Permission Denied at $node for $conf_dir/node-name"
```
andriytk, line 252: It is not necessary always must be the "permission denied" error. In case there is some other error, this will cause confusion.

andriytk:
> I think the problem is more general: if anything goes wrong, hare-bootstrap script will not help to understand at which node the problem appeared.

@knekrasov I cannot confirm this. I think in some cases we can see where the problem is, but I'm not sure. It's better to revise the script(s) for this issue probably.

I'd suggest using this function for all ssh commands in the script:

```sh
myssh() {
    local hostname=$1
    shift
    ssh -n $hostname "$@" 2>&1 | sed "s/^/$hostname: /"
}
```

It will prefix each output line with the node name, for example:

```bash
$ set -o pipefail
$ myssh sage-tier1-1 ls bbb
sage-tier1-1: ls: cannot access bbb: No such file or directory
$ echo $?
2
```

utils/hare-bootstrap
supritshinde: Hi @andriytk Yeah I'll edit as suggested. Thanks.

supritshinde:
Hi @mssawant While analysing the issue by @andriytk The permission denial error was given out for Access issue to the node name file inside confdir. This was seen after "Generating cluster Config.... line" And the code block after that wasn't completely executed so, before printing 'OK'. System throwed the error and exited. And, in that part of code. We echo name of node into the node name file. (just this single operation is happening nothing else). Here's the link to JIRA ticket https://jts.seagate.com/browse/EOS-18914

supritshinde:
The issue is addressed in https://github.com/Seagate/cortx-hare/pull/1608 
And that's better solution. Hence, Closing this PR.

PR #1598, opened 2021-04-21
EOS-19637: node-start never works cluster-wise

knekrasov:
JIRA issue: [EOS-19637](https://jts.seagate.com/browse/EOS-19637)

Due to a mistake in bash script the script always thinks that it is
invoked with `--node` flag.

The problem is that in bash there is no boolean types so node=true and
node=false are understood in the same way (as a string with non-zero
length) - that turns out to be semantics of [[ ]] in bash.

Solution:
1. Don't evaulate true/false within double brackets.

### DIFF
--- a/utils/hare-start
+++ b/utils/hare-start
@@ -70,7 +70,7 @@ if  hctl status > /dev/null 2>&1; then
     die 'Cluster is up and running.'
 fi
 
-if [[ $node ]]; then
+if $node; then
     start-node
 else
     hctl bootstrap -c $conf_dir


Shreya-18 (approved):
Nice catch!! :-)

utils/hare-start
```diff
@@ -51,12 +51,12 @@ TEMP=$(getopt --options h \
 eval set -- "$TEMP"
 
 conf_dir=/var/lib/hare
-node=false
+node=
```
mssawant, line 54: This is how we use boolean in some of the scripts,
```
conf_create=false
opt_mkfs=

while true; do
    case "$1" in
        -h|--help)      usage; exit ;;
        -c|--conf-dir)  conf_dir=$2; shift 2 ;;
        --consul-addr)  CONSUL_ADDR=$2; shift 2 ;;
        --consul-port)  CONSUL_PORT=$2; shift 2 ;;
        --mkfs)         opt_mkfs=--mkfs; shift ;;
        --conf-create)  conf_create=true; shift ;;
        --)             shift; break ;;
        *)              break ;;
    esac
done
...
    if $conf_create; then
        say 'Generating node configuration...'
        PATH="$(dirname $(readlink -f $0)):$PATH" cfgen -o $conf_dir $cdf
        dhall text < $conf_dir/confd.dhall | m0confgen > $conf_dir/confd.xc
        while read node _; do
        if is_localhost $node; then
            echo $node > $conf_dir/node-name
        fi
        done < <(get_servers_from_conf)
        echo ' OK'
   fi
```

knekrasov: Updated.

suvratjoshi (approved):

mssawant (approved):

PR #1601, opened 2021-04-27
EOS-19796 Hare post_install need to use confstore keys instead of Provisoiner get_setup_info api

SwapnilGaonkar7:
Solution: Modified code to read server node type and use it during post_install

UT:
Tested on local setup deployed using Jenkin.
Will test it on local VM if required.

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -83,11 +83,28 @@ def _report_unsupported_features(features_unavailable):
         uf_db.store_unsupported_features('hare', features_unavailable))
 
 
-def logrotate():
+def get_server_type(url: str) -> str:
     try:
-        setup_info = get_data_from_provisioner_cli('get_setup_info')
-        if setup_info != 'unknown':
-            server_type = setup_info['server_type']
+        provider = ConfStoreProvider(url)
+        machine_id = provider.get_machine_id()
+        server_type = provider.get(
+            f'server_node>{machine_id}>type')
+
+        if server_type == 'VM':
+            return 'virtual'
+        else:
+            return 'physical'
+    except Exception as error:
+        logging.error('Cannot get server type (%s)', error)
+        return 'unknown'
+
+
+def logrotate(url: str):
+    try:
+        server_type = get_server_type(url)
+        logging.info('Server type (%s)', server_type)
+
+        if server_type != 'unknown':
             shutil.copyfile(
                 f'/opt/seagate/cortx/hare/conf/logrotate/{server_type}',
                 '/etc/logrotate.d/hare')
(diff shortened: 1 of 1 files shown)


provisioning/miniprov/hare_mp/main.py
```diff
@@ -83,31 +83,49 @@ def _report_unsupported_features(features_unavailable):
...
-            server_type = setup_info['server_type']
-            shutil.copyfile(
-                f'/opt/seagate/cortx/hare/conf/logrotate/{server_type}',
-                '/etc/logrotate.d/hare')
+        provider = ConfStoreProvider(url)
+        machine_id = provider.get_machine_id()
+        server_type = provider.get(
+            f'server_node>{machine_id}>type')
+
+        if server_type == 'VM':
```
mssawant, line 93: This is error prone, just in case this one string ('VM') changes the code will break. I suggest to have a cortx-py-utils interface that checks the type, e.g. is_setup_virtual() if not already present, @SwapnilGaonkar7 can you please confirm?
May be for now we can land these changes but we must propose to have a generic function. Or you can yourself submit a PR in cortx-py-utils to add such a function.
cc @ujjwalpl, @sachinpunadikar.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -83,31 +83,49 @@ def _report_unsupported_features(features_unavailable):
...
-        setup_info = get_data_from_provisioner_cli('get_setup_info')
-        if setup_info != 'unknown':
-            server_type = setup_info['server_type']
-            shutil.copyfile(
-                f'/opt/seagate/cortx/hare/conf/logrotate/{server_type}',
-                '/etc/logrotate.d/hare')
+        provider = ConfStoreProvider(url)
+        machine_id = provider.get_machine_id()
+        server_type = provider.get(
+            f'server_node>{machine_id}>type')
```
mssawant, line 91: Let's avoid the code duplication, please move the code in a function, e.g. get_setup_info(url: str) and invoke it from the main(), unconditionally and pass the type to both the functions (logrotate() and unsupported_feature()).

provisioning/miniprov/hare_mp/main.py
SwapnilGaonkar7: @mssawant Quick search on 'cortx-utils' suggests that there is no such functions as of now.
@sachinpunadikar Can you please confirm?

provisioning/miniprov/hare_mp/main.py
SwapnilGaonkar7: Done

SwapnilGaonkar7:
@mssawant Please have a look

provisioning/miniprov/hare_mp/main.py
```diff
@@ -83,28 +83,48 @@ def _report_unsupported_features(features_unavailable):
...
 
 
-def unsupported_feature():
+def unsupported_feature(url: str):
     try:
         features_unavailable = []
         path = '/opt/seagate/cortx/hare/conf/setup_info.json'
         with open(path) as hare_features_info:
             hare_unavailable_features = json.load(hare_features_info)
-            setup_info = get_data_from_provisioner_cli('get_setup_info')
```
suvratjoshi, line 104: Sorry, I have basic questions.
My understanding is that Hare used to call a provisioner command "get_setup_info", which has now started failing, and it is suggested to Hare that now this info needs to be retrieved directly from ConfStore.
1. Why `get_setup_info` failed?
2. Isn't it expected that `get_setup_info` needs to be updated / corrected to fetch the data correctly? (as this command might be called from elsewhere as well)?

SwapnilGaonkar7: @83bhp Any idea about the above comment from @suvratjoshi ?

mssawant: I think provisioner has deprecated get_setup_info and is no more supported, @83bhp, kindly confirm.

mssawant (approved):

provisioning/miniprov/hare_mp/main.py
suvratjoshi: We can keep this question open in public channels as discussed.
Approving this PR.

suvratjoshi (approved):

PR #1606, opened 2021-05-02
EOS-19683: Hax service times out during 'hctl shutdown'

mssawant:
Problem:
Hax starts m0_rconfc to invoke Spiel API.
Hax shutdown times out because m0_rconfc_stop() hangs and
m0_halon_interface_stop() also does not complete.
When rconfc starts, it requests entrypoint information
(endpoints of confds in the cluster and principal rm
endpoint). It establishes connection with confds and principal
rm. rconfc creates rm context, sets up rm creditor for read_lock
which includes setting up rpc connection and session with rm
endpoint.

Why m0_rconfc_stop() hangs?

Rconfc is designed to sustain confd failures where, in case a confd
fails and rconfc receives a confd failure notification,
- rconfc invalidates all the configuration objects in confc
- restarts rconfc version election, re-sends entrypoint request in
  a hope to get a new principal RM endpoint from Hare.
- This entrypoint request is a blocking request and rconfc will block
  if the reply is not sent.
- On receiving entrypoint reply with the RM endpoint, rconfc tries to
  establish a new rpc connection, initialize RM by requesting for read
  lock credit. But if the confd hosting principal RM service is down
  already, the further connection to RM endpoint timesout, rconfc then
  again requests for a new entrypoint information, this goes on till Hare
  replies a non-empty entrypoint reply. Thus, blocking rconfc finalization.
- If the HA state of corresponding RM service and confd process is not
  notified as M0_NC_FAILED by Hare, as part of rconfc re-initialization,
  RM from Hare motr land sends credit borrow requests for read lock resource.
  This also blocks as the corresponding rpc request is tried infinitey until
  it succeeds. This never ending rpcs will stop only if the corresponding
  session is cancelled (when the respective RM service is notified M0_NC_FAILED)
  or when the process is killed.
- As part of rconfc re-initialization, rconfc stops and starts again,
  thus, after entrypoint request, nvec request is sent, requesting for the processes 
  ha states. Now, if the nvec reply from Hare does not contain actual process states
  (M0_NC_FAILED/M0_NC_ONLINE) and when the ha states in nvec are applied at
  the receiver end, Hare motr land or any other peer motr process may try to send
  requests which may fail with timeout, again blocking the finalization.

How this patch solves this m0_rconfc_stop() problem?

- Hax failure event notification mechanism is modified a bit, where now Hax
  on each node maintains statuses for all the Motr processes in the cluster in
  Consul KV. This is updated in context of Consul's process failure notifications.
- On each node, Hax sends M0_NC_FAILED events for all the motr processes in the
  cluster to itself and local motr processes.
  So, during shutdown, Hax receives Consul notifications for Motr processes,
  Hax verifies the Consul's failure notifications, Hax then sends M0_NC_FAILED
  notifications to its Motr land and other alive motr processes if any.
- When Motr land of Hax receives the failure notifications. The halinks are
  disconnected and rconfc starts re-initialization (sends entrypoint request).
- Hax maintains the motr process statusses in Consul KV, this saved status is checked
  before replying to entrypoint request, So, if all the confds have stopped, the
  entrypoint reply will be sent with no confds and rm endpoint. This breaks the rconfc
  re-initialization loop.
- Also the nvec request reply is populated with the actual motr processes and
  services statuses.
- Now, when m0_rconfc_stop() is invoked (as part of Hax shutdown) rconfc stops cleanly
  as all the respective processes and services are marked M0_NC_FAILED.
- m0_halon_interface_stop() also completes successfully.


mssawant:
Presently the hax shutdown is crashing with the following stack,
```
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: motr[08992]:  a740  FATAL  [lib/assert.c:50:m0_panic]  panic: fatal signal delivered at unknown() (unknown:0)  [git: sage-base-1.0May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: Motr panic: fatal signal delivered at unknown() unknown:0 (errno: 0) (last failed: none) [git: sage-base-1.0-332-g2d26847] pid: 89May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: Motr panic reason: signo: 11
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(m0_arch_backtrace+0x2f)[0x7f4a15ffaecf]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(m0_arch_panic+0xf3)[0x7f4a15ffb0b3]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(+0x33b434)[0x7f4a15feb434]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(+0x34b0f8)[0x7f4a15ffb0f8]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libpthread.so.0(+0xf630)[0x7f4a2556a630]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(m0_mutex_is_locked+0x9)[0x7f4a15ff0db9]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(m0_ha_entrypoint_client_state_get+0x18)[0x7f4a15fc5e68]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(+0x2d4d3b)[0x7f4a15f84d3b]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(+0x33c8e7)[0x7f4a15fec8e7]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(+0x33c93a)[0x7f4a15fec93a]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(+0x3d7c51)[0x7f4a16087c51]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(m0_ha_entrypoint_client_stop+0x8e)[0x7f4a15fc5d7e]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(+0x319ed4)[0x7f4a15fc9ed4]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(m0_module_fini+0x76)[0x7f4a1603b546]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(m0_ha_disconnect+0x48)[0x7f4a15fc9358]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(+0x32229a)[0x7f4a15fd229a]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(m0_module_fini+0x76)[0x7f4a1603b546]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libmotr.so.2(m0_halon_interface_stop+0xa2)[0x7f4a15fd4262]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /opt/seagate/cortx/hare/lib64/python3.6/site-packages/hax/motr/../../libhax.cpython-36m-x86_64-linux-gnu.so(motr_api_stop+0x1a)[0xMay 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libffi.so.6(ffi_call_unix64+0x4c)[0x7f4a1d8aae2c]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libffi.so.6(ffi_call+0x1f5)[0x7f4a1d8aa755]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /usr/lib64/python3.6/lib-dynload/_ctypes.cpython-36m-x86_64-linux-gnu.so(_ctypes_callproc+0x502)[0x7f4a1dabd8e2]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /usr/lib64/python3.6/lib-dynload/_ctypes.cpython-36m-x86_64-linux-gnu.so(+0x97c6)[0x7f4a1dab67c6]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libpython3.6m.so.1.0(_PyObject_FastCallDict+0x90)[0x7f4a2581f160]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libpython3.6m.so.1.0(+0x15098c)[0x7f4a258c798c]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libpython3.6m.so.1.0(_PyEval_EvalFrameDefault+0x3a7)[0x7f4a258bc0f7]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libpython3.6m.so.1.0(+0x1505ca)[0x7f4a258c75ca]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libpython3.6m.so.1.0(+0x150903)[0x7f4a258c7903]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libpython3.6m.so.1.0(_PyEval_EvalFrameDefault+0x3a7)[0x7f4a258bc0f7]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libpython3.6m.so.1.0(+0x1505ca)[0x7f4a258c75ca]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libpython3.6m.so.1.0(+0x150903)[0x7f4a258c7903]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libpython3.6m.so.1.0(_PyEval_EvalFrameDefault+0x3a7)[0x7f4a258bc0f7]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libpython3.6m.so.1.0(+0x14fdee)[0x7f4a258c6dee]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libpython3.6m.so.1.0(_PyFunction_FastCallDict+0x24f)[0x7f4a258c89ff]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libpython3.6m.so.1.0(_PyObject_FastCallDict+0x10e)[0x7f4a2581f1de]
May 01 22:02:42 ssc-vm-1623.colo.seagate.com hare-hax[8992]: /lib64/libpython3.6m.so.1.0(_PyObject_Call_Prepend+0x61)[0x7f4a2581f2f1]
```
This problem is being debugged and can be tracked as a separate ticket to unblock the blocker ticket EOS-19683.

madhavemuri (approved):
Flow-wise steps seems fine as rconfc will be sending HA entrypoint requests continuously as it don't understand the cluster stop is triggered. In this case Hax is breaking this loop either using ha stop and also by not filling confd/rm values.
Ideally not sending confd values in the entrypoint reply should work fine. Need to see further why this didn't worked.

mssawant:
Found a bug in m0_halon_interface with respect to above crash. Will post a motr patch for the same.

hax/hax/handler.py
```diff
@@ -53,9 +53,11 @@ def __init__(self, q: Queue, motr: Motr, herald: DeliveryHerald,
                          name='qconsumer',
                          args=(q, motr))
         self.is_stopped = False
+        self.is_finalized = False
```
knekrasov, line 56: I can't see where this field is used. Is it really needed?

hax/hax/handler.py
```diff
@@ -53,9 +53,11 @@ def __init__(self, q: Queue, motr: Motr, herald: DeliveryHerald,
...
         self.consul = consul
         self.eq_publisher = EQPublisher()
         self.herald = herald
+        self.motr = motr
```
knekrasov, line 60: It seems like this assignment is not used also.

hax/hax/handler.py
```diff
@@ -232,6 +233,9 @@ def pull_msg():
                     # no op, swallow the exception
                     LOG.exception('**ERROR**')
         except StopIteration:
-            ffi.shun_motr_thread()
+            LOG.info('Consumer Stopped')
         finally:
+            LOG.info('Closing motr')
+            motr.close()
```
knekrasov, line 239: Previously, Motr context was started in main thread and stopped also there.
With this PR there is a kind of skewness: consumer thread invokes `motr.close()` although it didn't start it. Do we really need this `motr.close()` here in this thread?

hax/hax/hax.py
```diff
@@ -125,11 +125,9 @@ def main():
         run_server(q,
                    herald,
                    consul_util=util,
-                   threads_to_wait=[consumer, stats_updater, rconfc_starter])
+                   threads_to_wait=[stats_updater, consumer, rconfc_starter])
```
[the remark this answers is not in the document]
mssawant: Added a comment in the code.

hax/hax/motr/__init__.py
```diff
@@ -148,7 +153,8 @@ def _entrypoint_request_cb(self, reply_context: Any, req_id: Any,
             LOG.exception('Failed to notify failure for %s', process_fid)
 
         LOG.debug('enqueue entrypoint request for %s', remote_rpc_endpoint)
-        self.queue.put(
+        # self.queue.put(
```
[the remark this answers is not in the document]
mssawant: Restored the original line in the updated patch.

hax/hax/motr/__init__.py
```diff
@@ -148,7 +153,8 @@ def _entrypoint_request_cb(self, reply_context: Any, req_id: Any,
...
+        self.send_entrypoint_request_reply(
```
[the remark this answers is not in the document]
mssawant: Restored in the updated patch.

hax/hax/handler.py
mssawant: Removed.

hax/hax/handler.py
mssawant: Yeah, it is observed that during motr finalization, there are messages received from the motr land, thus if all the threads are stopped and motr finalization is invoked from the main thread then the messages seem to block. Hence invoking it from the consumer thread.

mssawant:
Update,
`m0_rconfc_stop()` and `m0_halon_interface_stop()` now completes successfully, there's just 1 panic seen during `m0_halon_interface_fini()`
```
motr[105577]:  da80  FATAL  [addb2/addb2.c:1144:m0_addb2__mach_print]  mach: 0xe0ed60 "main()" "main()".
May 13 15:52:40 ssc-vm-1624.colo.seagate.com hare-hax[105577]: motr[105577]:  d980  FATAL  [lib/assert.c:50:m0_panic]  panic: sys->sy_total == 0 at m0_addb2_sys_fini() (addbMay 13 15:52:40 ssc-vm-1624.colo.seagate.com hare-hax[105577]: Motr panic: sys->sy_total == 0 at m0_addb2_sys_fini() addb2/sys.c:257 (errno: 0) (last failed: none) [git: sagMay 13 15:52:40 ssc-vm-1624.colo.seagate.com hare-hax[105577]: Motr panic reason: 1
May 13 15:52:40 ssc-vm-1624.colo.seagate.com hare-hax[105577]: /lib64/libmotr.so.2(m0_arch_backtrace+0x2f)[0x7fcb56ffd31f]
May 13 15:52:40 ssc-vm-1624.colo.seagate.com hare-hax[105577]: /lib64/libmotr.so.2(m0_arch_panic+0xf3)[0x7fcb56ffd503]
May 13 15:52:40 ssc-vm-1624.colo.seagate.com hare-hax[105577]: /lib64/libmotr.so.2(+0x33b884)[0x7fcb56fed884]
May 13 15:52:40 ssc-vm-1624.colo.seagate.com hare-hax[105577]: /lib64/libmotr.so.2(+0x270f4f)[0x7fcb56f22f4f]
May 13 15:52:40 ssc-vm-1624.colo.seagate.com hare-hax[105577]: /lib64/libmotr.so.2(m0_subsystems_fini+0x2a)[0x7fcb5700857a]
May 13 15:52:40 ssc-vm-1624.colo.seagate.com hare-hax[105577]: /lib64/libmotr.so.2(m0_module_fini+0x76)[0x7fcb5703d6a6]
```
But this is happens on all the nodes, which is a lot better situation than 1 node finishing but other nodes used to hang.
There were some inconsistencies in hax communication of failure events which are fixed in the updated patch.
cc @max-seagate , @madhavemuri , @knekrasov

hax/hax/handler.py
ivan-alekhin: In this case we probably need to move `m0_addb2_global_thread_leave` at the end of `init_motr_api`. That should help with the `sys->sy_total == 0 ` panic (but I might be wrong here -- I'm assuming init_motr_api is called from the "main" thread and fini_motr_api is called from a non-main thread).

andriytk:
> Inorder for hax Motr land finalization to complete successfully it is
important that the failure events for cluster processes are propogated
to every node.

Sorry, which failure events?

mssawant:
@andriytk,
**Problem:**
Hax starts m0_rconfc to invoke Spiel API.
Hax shutdown times out because m0_rconfc_stop() hangs and m0_halon_interface_stop() also does not complete.

**Why m0_rconfc_stop() hangs?**
- rconfc requests entrypoint information (endpoints of confds in the cluster and principal rm endpoint).
- When rconfc starts, it establishes connection with confds and principal rm
- Creates rm context, sets up rm creditor for read_lock which includes setting up rpc connection and session with rm endpoint
Rconfc is designed to sustain confd failures where in case a confd fails 
- rconfc invalidates all the configuration objects in confc
- restarts rconfc version election, re-sends entrypoint request in a hope to get a new principal RM endpoint.
- This entrypoint request is a blocking request and rconfc will block if the reply is not sent.
- On receiving entrypoint reply with the RM endpoint, it tries to establish a new connection, so if the HA state of corresponding RM service and confd process is not set to FAILED, rconfc keeps sending entrypoint requests in a loop, this blocks rconfc.
- As part of rconfc re-initialization, rconfc stops and starts again, thus after entrypoint request, it also sends nvec request. Hax needs to reply to this nvec request appropriately with proper processes and services states for confd and RM.

**How this patch solves this m0_rconfc_stop() problem?**
- Hax failure event notification mechanism is modified a bit, where now Hax on each node maintains statuses for all the Motr processes in the cluster in Consul KV. This is updated in context of Consul process failure notifications.
- On each node, Hax sends M0_NC_FAILED events for all the motr processes in the cluster to itself and local motr processes.
   - So, during shutdown, Hax receives Consul notifications for Motr processes, Hax verifies the Consul's failure notifications, Hax then sends M0_NC_FAILED notifications to itself and other alive motr processes if any.
- When Motr land of Hax receives the failure notifications. The halinks are disconnected and rconfc starts re-initialization (sends entrypoint request).
- Hax also maintains a failed processes list, that is updated whenever a motr process stops or Consul reports failure of a Motr process. The failed list is checked before replying to entrypoint request, So, if all the confds have stopped, the entrypoint reply will be sent with no confds and rm endpoint. This breaks the rconfc re-initialization loop.
- Also the nvec request reply is populated with the actual motr processes and services statuses.
- Now, when m0_rconfc_stop() is invoked (as part of Hax shutdown) rconfc stops cleanly as all the respective processes and services are marked M0_NC_FAILED.
- m0_halon_interface_stop() also completes successfully.

Makes sense?

andriytk:
Yes! It's much better now. So, basically, from now on hax will know that there are no confd services available online and will reply to its own rconfc with empty reply without trying to establish connections with confds which are not available any more, right?

And why such empty request is not sent currently? Because some confds are not marked as failed yet at the time of hax shutdown or we just don't check their state in hax?

mssawant:
>So, basically, from now on hax will know that there are no confd services available online and will reply to its own rconfc with empty reply (so that it would not try to establish connections with confds which are not available any more), right?

Correct.

>And why such empty request is not sent currently? Because some confds are not marked as failed yet at the time of hax shutdown or we just don't check their state in hax?

Right, presently, Consul notifications are not handled properly and notifications for peer motr processes are not sent to local Motr processes on each node. Also Hax does not maintain any list of failed motr services which is why we are not able to send an empty reply to entrypoint request currently.

hax/hax/handler.py
```diff
@@ -111,16 +115,35 @@ def update_process_failure(self, q: Queue,
...
                           reply_to=None))
+                if current_status not in (ServiceHealth.UNKNOWN,
+                                          ServiceHealth.OFFLINE):
+                    # We also need to account and report the failure of remote
+                    # Motr processes to this node's hax and motr processes.
+                    # When Consul reports a remote process failure, hax
+                    # confirms its current status from Consul KV and updates
+                    # the list of failed services and also adds it to the
+                    # broadcast list.
+                    self.consul.update_failed_svcs(
```
andriytk, line 126: Looks like you put online services to the local list also? So how does the check for `if state.fid in self.consul.failed_svcs:` above works - will it check for online services also?

BTW, imagine some large cluster configuration with thousands ios processes - do we really want to maintain them in this list? This does not look like a scalable solution. I'd suggest to maintain only confd processes in it.

hax/hax/motr/__init__.py
```diff
@@ -148,52 +162,93 @@ def _entrypoint_request_cb(self, reply_context: Any, req_id: Any,
...
-            rm_fid = util.get_rm_fid()
+            # When stopping, there's a possibility that hax may receive
+            # an entrypoint request from motr land. In order to unblock
+            # motr land, reply with entrypoint request with no confds
+            # and RM endpoints as the processes might have already
+            # stopped.
+            rc_quorum = 0
+            if self.is_stopping:
+                confds = []
+                rm_fid = Fid(0, 0)
```
andriytk, line 201: Why isn't it enough to do this only? Why do we need some additional `self.consul.failed_svcs` list?

hax/hax/util.py
```diff
@@ -212,6 +212,8 @@ class ConsulUtil:
     def __init__(self, raw_client: Optional[Consul] = None):
         self.cns: Consul = raw_client or Consul()
         self.kv = ConsulKVBasic(cns=self.cns)
+        self.failed_svcs_lock = Lock()
+        self.failed_svcs: List[Fid] = []
```
andriytk, line 216: Simple list maybe enough for confds list only. But for all m0d processes it will make linear search in the list. Imagine large clusters with tens of thousands of processes - it will kill scalability and performance.

hax/hax/handler.py
mssawant: >Looks like you put online services to the local list also? So how does the check for if state.fid in self.consul.failed_svcs: above work - won't it check for the online services also?

We add the process fid on failure and remove it when the process comes online from the failed_svc lists. So, failed_svcs list will have only the failed process fids.

>BTW, imagine some large cluster configuration with thousands ios processes - do we really want to maintain them in this list? This does not look like a scalable solution. I'd suggest to maintain only confd processes in it.

Yes, I did think about it, it will grow but presently I don't have an alternative solution for this. We can revisit this at later point. I though rconfc or RM would check the ha states for processes and services but seems like this does not happen and RM still tries to connect. That is the reason we need to maintain the list of failed services. Ideally there should be no need for this if the ha state updates are percolated through all the modules.

>Also, why do we need to maintain it in hax instead of Consul?

Because we don't need to persist the list, it is dynamic and can remain in-memory. Also during shutdown Consul might not be accessible as confds are stopped (No cluster leader will be available) and hax will not be able to fetch the failure list.

hax/hax/motr/__init__.py
mssawant: So, even before shutdown starts (i.e. self.is_stopping = True), as hax notifies about confd failure (STOPPED), rconfc from hax motr land starts re-election and RM starts re-connecting, new entrypoint request is received before shutdown starts and hax then replies with a non-empty confd list and rm endpoint as `self.is_stopping` is still false. On confd failure, rconfc stops and starts again and this goes on in a loop,
```
static void rconfc_start(struct m0_rconfc *rconfc)
{
        struct m0_ha                   *ha  = m0_get()->i_ha;
        struct m0_ha_entrypoint_client *ecl = &ha->h_entrypoint_client;
        int                             rc  = 0;

        M0_PRE(rconfc->rc_local_conf == NULL);

        M0_ENTRY();

        while (rconfc_state(rconfc) != M0_RCS_FAILURE) {
                rconfc_state_set(rconfc, M0_RCS_ENTRYPOINT_WAIT);

                if (rconfc->rc_ha_entrypoint_rep.hae_control == M0_HA_ENTRYPOINT_QUERY) {
                        M0_LOG(M0_DEBUG, "Querying ENTRYPOINT...");
                        m0_ha_entrypoint_client_request(ecl);
                        break;
                }

                M0_LOG(M0_DEBUG, "ENTRYPOINT ready...");
                rc = rconfc_start_internal(rconfc);
                rconfc->rc_ha_entrypoint_rep.hae_control = M0_HA_ENTRYPOINT_QUERY;
                if (rc == 0)
                        break;
                m0_ha_entrypoint_rep_free(&rconfc->rc_ha_entrypoint_rep);
                M0_LOG(M0_DEBUG, "Try reconnecting after rc=%d", rc);
                M0_CNT_INC(rconfc->rc_ha_entrypoint_retries);
        }
```
This blocks rconfc shutdown and hax shutdown. In order to break this, we need to reply with only alive confds else empty.

hax/hax/util.py
mssawant: Okay, let me try to just add confds here.

hax/hax/handler.py
andriytk: > Also during shutdown Consul might not be accessible as confds are stopped (No cluster leader will be available) and hax will not be able to fetch the failure list.

Consul leader and RC leader and different things. Although, RC leader might not be available if there are no confds - true, Consul leader still will be available and you still will be able to access Consul KV even if all confds are down.

mssawant: Ah yes, here we need not get leader session, which fails is RC leader is not present. But even then the time required to fetch the list from Consul, look up or update and fetch again from consul is really not needed and this information can be maintained in memory. Let me try to reduce the list to just have confds.

hax/hax/motr/__init__.py
andriytk: This does not sound like a robust solution for breaking this. Because during the cluster shutdown confd may be already down or shutting down while hax still will think it's alive...

hax/hax/handler.py
andriytk: I thought the status of processes is updated in Consul KV anyway, isn't it so? If so, it will be only reads from Consul KV then.

The idea of maintaining processes status in Consul KV and in hax memory is not sound right.

hax/hax/hax.py
```diff
@@ -122,14 +122,20 @@ def main():
         stats_updater = _run_stats_updater_thread(motr, consul_util=util)
         # [KN] This is a blocking call. It will work until the program is
         # terminated by signal
+        # stats_updater and rconfc_started both invoke
+        # {adopt,shun}_motr_thread(), adopt_motr_thread() takes a reference
+        # on motr addb2 machine internally in motr and shun_motr_thread()
+        # releases it. We are invoke motr::stop() and motr::fini() when
+        # consumer thread stops, thus, it is important that all the motr
+        # references are dropped before invoking motr finalization.
+        # Hence, consumer thread is last to stop and join.
```
knekrasov, line 131: I'm afraid the order is still not guaranteed. Firstly `stop()` method is invoked for every thread, then we iterate again and perform `join()`. `stop()` is not blocking, it actually can't guarantee that the thread will stop immediately. That's why relying on the order of threads in the `threads_to_wait` list is not redundant. If we do need to guarantee the order between the threads, we'll most probably need to use other mechanisms (such as barriers or something).

hax/hax/handler.py
mssawant: Okay, it was mainly about performance but let me try and see if there are any performance issues.

hax/hax/motr/__init__.py
```diff
@@ -148,52 +162,93 @@ def _entrypoint_request_cb(self, reply_context: Any, req_id: Any,
...
+            # case, so we wait until bootstrapping is done that is all the
+            # motr services are up and then account for the motr process
+            # failures and lookup this list of failed motr processes before
+            # replying to entrypoint request.
+            if self.spiel_ready:
+                active_confds = confds
+                for confd in active_confds:
+                    with util.failed_svcs_lock:
+                        if confd.fid in util.failed_svcs:
+                            confds.remove(confd)
```
madhavemuri, line 230: I think there is no need to exclude failed confd's, this may impact quorum of confd's. Rconfc logic manages it internally.

suvratjoshi:
If I understand it correctly,
Regarding Consul delayed notification : 
Is this a consul bug?
Do we need to take that to hashicorp? (something like this : https://github.com/hashicorp/consul/issues/2970)

If not a bug, can this be improved by any chance, with some configuration settings of consul?

andriytk:
Consul might delay notifications until the status of the services is synchronised in its catalogue between the consul servers. And this synchronisation may take time depending on the number of consul servers configured in the cluster and many other factors, like network state/load, machines load, etc.

See more at https://github.com/hashicorp/consul/issues/7069#issuecomment-705461598 also.

hax/hax/motr/__init__.py
mssawant: With non-empty entrypoint reply I have observed that rconfc stop blocks and hax keeps getting entrypoint requests from motr land.

hax/hax/hax.py
mssawant: You are right, after sending stop signal each thread can finalize out of order. We could move motr_thread_shut() from filestats_updater and rconfc_starter to `stop_iteration` section but in case of exception it might be possible that `motr_thread_shun` is not invoked. Let me see and find a better place to do so.

hax/hax/motr/__init__.py
mssawant: Sorry didn't get you, we also invoke m0_rconfc_stop() and m0_rconfc_fini() through m0_spiel_rconfc_stop().

hax/hax/motr/__init__.py
mssawant: see,
```
May 20 16:35:54 ssc-vm-1623.colo.seagate.com hare-hax[118893]: motr[118893]:  4900   WARN  [ha/halon/interface.c:1073:m0_halon_interface_entrypoint_reply]  req_id=0:14 rc=0 confd_nr=3 confd_quorum=2 rm_fid=<7300000000000001:a> rm_eps=10.230.250.51@tcp:12345:2:1
May 20 16:35:54 ssc-vm-1623.colo.seagate.com hare-hax[118893]: motr[118893]:  4540   WARN  [ha/halon/interface.c:490:halon_interface_link_reused_cb]  req_id=0:14 hl=0x7fd7d4009360 ep=10.230.250.51@tcp:12345:1:1
May 20 16:35:54 ssc-vm-1623.colo.seagate.com hare-hax[118893]: 2021-05-20 16:35:54,265 [DEBUG] {Dummy-1} Entrypoint request has been replied to
May 20 16:35:55 ssc-vm-1623.colo.seagate.com hare-hax[118893]: motr[118893]:  4a20  ERROR  [conf/confc.c:557:m0_confc_init_wait]  <! rc=-110 confc=0x7fd86404cfa0 confd_addr=10.230.250.51@tcp:12345:2:1
May 20 16:35:56 ssc-vm-1623.colo.seagate.com hare-hax[118893]: motr[118893]:  4a20  ERROR  [conf/confc.c:557:m0_confc_init_wait]  <! rc=-110 confc=0x7fd86404fef0 confd_addr=10.230.250.53@tcp:12345:2:1
May 20 16:35:57 ssc-vm-1623.colo.seagate.com hare-hax[118893]: motr[118893]:  4a20  ERROR  [conf/confc.c:557:m0_confc_init_wait]  <! rc=-110 confc=0x7fd864051630 confd_addr=10.230.250.48@tcp:12345:2:1
May 20 16:35:57 ssc-vm-1623.colo.seagate.com hare-hax[118893]: motr[118893]:  4b30  ERROR  [conf/rconfc.c:1699:rconfc_herd_update]  <! rc=-2
May 20 16:35:57 ssc-vm-1623.colo.seagate.com hare-hax[118893]: motr[118893]:  4b30  ERROR  [conf/rconfc.c:1948:rconfc_entrypoint_consume]  <! rc=-2
May 20 16:35:57 ssc-vm-1623.colo.seagate.com hare-hax[118893]: motr[118893]:  4b30  ERROR  [conf/rconfc.c:1968:rconfc_start_internal]  <! rc=-2
May 20 16:35:57 ssc-vm-1623.colo.seagate.com hare-hax[118893]: motr[118893]:  5bb0   WARN  [ha/halon/interface.c:366:halon_interface_entrypoint_request_cb]  req_id=0:15 remote_rpc_endpoint=10.230.250.51@tcp:12345:1:1 process_fid=<7200000000000001:6> git_rev_id=51d02dac8c2096fbe82bd551dc0895d8967d0212 generation=18 pid=118893
May 20 16:35:57 ssc-vm-1623.colo.seagate.com hare-hax[118893]: 2021-05-20 16:35:57,273 [DEBUG] {Dummy-1} Received entrypoint request from remote endpoint '10.230.250.51@tcp:12345:1:1', process fid = 0x7200000000000001:0x6 The request will be processed in another thread.
May 20 16:35:57 ssc-vm-1623.colo.seagate.com hare-hax[118893]: 2021-05-20 16:35:57,273 [DEBUG] {Dummy-1} enqueue entrypoint request for 10.230.250.51@tcp:12345:1:1
May 20 16:35:57 ssc-vm-1623.colo.seagate.com hare-hax[118893]: 2021-05-20 16:35:57,274 [DEBUG] {Dummy-1} Processing entrypoint request from remote endpoint '10.230.250.51@tcp:12345:1:1', process fid 0x7200000000000001:0x6
May 20 16:35:57 ssc-vm-1623.colo.seagate.com hare-hax[118893]: 2021-05-20 16:35:57,288 [DEBUG] {Dummy-1} Passing the entrypoint reply to hax.c layer
May 20 16:35:57 ssc-vm-1623.colo.seagate.com hare-hax[118893]: motr[118893]:  4900   WARN  [ha/halon/interface.c:1073:m0_halon_interface_entrypoint_reply]  req_id=0:15 rc=0 confd_nr=3 confd_quorum=2 rm_fid=<7300000000000001:a> rm_eps=10.230.250.51@tcp:12345:2:1
```
Hax motr land keeps sending the entrypoint requests, other remaining confd's shutdown is also delayed.
```
May 20 16:37:37 ssc-vm-1623.colo.seagate.com motr-server[119434]: motr[119434]:  95e0   WARN  [reqh/reqh.c:561:m0_reqh_fop_handle]  fop "Rpc conn establish"@0x7ff2480a2110 disallowed: -111.
May 20 16:37:37 ssc-vm-1623.colo.seagate.com motr-server[119434]: motr[119434]:  95e0  ERROR  [reqh/reqh.c:568:m0_reqh_fop_handle]  <! rc=-108 Service shutdown.
May 20 16:37:40 ssc-vm-1623.colo.seagate.com motr-server[119434]: motr[119434]:  94e0  ERROR  [reqh/reqh.c:454:m0_reqh_fop_allow]  <! rc=-111
May 20 16:37:40 ssc-vm-1623.colo.seagate.com motr-server[119434]: motr[119434]:  95e0   WARN  [reqh/reqh.c:561:m0_reqh_fop_handle]  fop "Rpc conn establish"@0x7ff2480a2110 disallowed: -111.
May 20 16:37:40 ssc-vm-1623.colo.seagate.com motr-server[119434]: motr[119434]:  95e0  ERROR  [reqh/reqh.c:568:m0_reqh_fop_handle]  <! rc=-108 Service shutdown.
May 20 16:37:43 ssc-vm-1623.colo.seagate.com motr-server[119434]: motr[119434]:  94e0  ERROR  [reqh/reqh.c:454:m0_reqh_fop_allow]  <! rc=-111
May 20 16:37:43 ssc-vm-1623.colo.seagate.com motr-server[119434]: motr[119434]:  95e0   WARN  [reqh/reqh.c:561:m0_reqh_fop_handle]  fop "Rpc conn establish"@0x7ff2480a2110 disallowed: -111.
May 20 16:37:43 ssc-vm-1623.colo.seagate.com motr-server[119434]: motr[119434]:  95e0  ERROR  [reqh/reqh.c:568:m0_reqh_fop_handle]  <! rc=-108 Service shutdown.
May 20 16:37:46 ssc-vm-1623.colo.seagate.com motr-server[119434]: motr[119434]:  94e0  ERROR  [reqh/reqh.c:454:m0_reqh_fop_allow]  <! rc=-111
May 20 16:37:46 ssc-vm-1623.colo.seagate.com motr-server[119434]: motr[119434]:  95e0   WARN  [reqh/reqh.c:561:m0_reqh_fop_handle]  fop "Rpc conn establish"@0x7ff2480a2110 disallowed: -111.
May 20 16:37:46 ssc-vm-1623.colo.seagate.com motr-server[119434]: motr[119434]:  95e0  ERROR  [reqh/reqh.c:568:m0_reqh_fop_handle]  <! rc=-108 Service shutdown
```
and one of the confd, that is stopping keeps sending RM_BORROW requests,
```
771888  15640197.087.229544  97f9860  DEBUG   rpc/item.c:737:m0_rpc_item_change_state             0x7ff248009c10 [REQUEST/M0_RM_FOP_BORR771889  15640197.087.231982  97f98e0  CALL    rpc/rpc_machine.c:903:item_received                 <
771890  15640197.087.232276  97f9880  CALL    fop/fop.c:182:m0_fop_put                            > fop: 0x7ff248009bf0 Credit Borrow [1771891  15640197.087.233470  97f9850  CALL    fop/fop.c:152:m0_fop_release                        >
771892  15640197.087.234294  97f97f0  CALL    fop/fop.c:139:m0_fop_fini                           > fop: 0x7ff248009bf0 Credit Borrow
771893  15640197.087.235762  97f9760  CALL    rpc/item.c:396:m0_rpc_item_fini                     > 0x7ff248009c10 [REQUEST/M0_RM_FOP_BORROW(88)] rc=0
```

hax/hax/hax.py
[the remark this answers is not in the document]
mssawant: restored the order, no need to change now.

mssawant:
EOS-18434 and EOS-20579 did not reproduce with the patch after more than 20 iterations of `hare_setup init`, see
```
May 20 22:43:38 ssc-vm-1625.colo.seagate.com hare-hax[101180]: Motr panic: sys->sy_total == 0 at m0_addb2_sys_fini() addb2/sys.c:257 (erMay 20 22:43:38 ssc-vm-1625.colo.seagate.com hare-hax[101180]: Motr panic reason: 1
May 20 22:43:38 ssc-vm-1625.colo.seagate.com hare-hax[101180]: /lib64/libmotr.so.2(m0_arch_backtrace+0x2f)[0x7fd2185ecfbf]
May 20 22:43:38 ssc-vm-1625.colo.seagate.com hare-hax[101180]: /lib64/libmotr.so.2(m0_arch_panic+0xf3)[0x7fd2185ed1a3]
May 20 22:43:38 ssc-vm-1625.colo.seagate.com hare-hax[101180]: /lib64/libmotr.so.2(+0x33b524)[0x7fd2185dd524]
May 20 22:43:38 ssc-vm-1625.colo.seagate.com hare-hax[101180]: /lib64/libmotr.so.2(+0x270bef)[0x7fd218512bef]
May 20 22:43:38 ssc-vm-1625.colo.seagate.com hare-hax[101180]: /lib64/libmotr.so.2(m0_subsystems_fini+0x2a)[0x7fd2185f821a]
May 20 22:43:38 ssc-vm-1625.colo.seagate.com hare-hax[101180]: /lib64/libmotr.so.2(m0_module_fini+0x76)[0x7fd21862d346]
May 20 22:43:38 ssc-vm-1625.colo.seagate.com hare-hax[101180]: /lib64/libmotr.so.2(m0_halon_interface_fini+0x76)[0x7fd2185c5f06]
May 20 22:43:38 ssc-vm-1625.colo.seagate.com hare-hax[101180]: /opt/seagate/cortx/hare/lib64/python3.6/site-packages/hax/motr/../../libhMay 20 22:43:38 ssc-vm-1625.colo.seagate.com hare-hax[101180]: /lib64/libffi.so.6(ffi_call_unix64+0x4c)[0x7fd21fe72e2c]
May 20 22:43:38 ssc-vm-1625.colo.seagate.com hare-hax[101180]: /lib64/libffi.so.6(ffi_call+0x1f5)[0x7fd21fe72755]
```
now during hax shutdown.

hax/hax/motr/__init__.py
andriytk: Imagine confd is stopping. hax' rconfc starts re-election in Motr land. But `self.consul.failed_svcs` is not updated yet - because the notification did not come yet. hax will reply with non-empty confd in entry-point reply, but there is no confd any more. Is such a scenario possible?

knekrasov (approved):

suvratjoshi (approved):

hax/hax/handler.py
mssawant: You are right, because it addb reference was taken by main thread releasing it in main thread helped. Moved motr_fini to hax.py.
```
May 21 10:24:42 ssc-vm-1623.colo.seagate.com hare-hax[56895]: 2021-05-21 10:24:42,962 [DEBUG] {qconsumer} Shunning Motr thread
May 21 10:24:42 ssc-vm-1623.colo.seagate.com hare-hax[56895]: 2021-05-21 10:24:42,963 [INFO] {qconsumer} Handler thread has exited
May 21 10:24:42 ssc-vm-1623.colo.seagate.com hare-hax[56895]: 2021-05-21 10:24:42,963 [INFO] {MainThread} The http server has stopped
May 21 10:24:42 ssc-vm-1623.colo.seagate.com hare-hax[56895]: 2021-05-21 10:24:42,963 [INFO] {MainThread} Finalizing motr
May 21 10:24:43 ssc-vm-1623.colo.seagate.com hare-hax[56895]: 2021-05-21 10:24:43,034 [DEBUG] {MainThread} Motr API context is down. ByeMay 21 10:24:43 ssc-vm-1623.colo.seagate.com systemd[1]: Stopped HAX daemon for Hare.
```
Finally.

mssawant:
@knekrasov, @andriytk , have updated the patch and finally hax is shutting down cleanly,
```
May 21 10:24:42 ssc-vm-1623.colo.seagate.com hare-hax[56895]: 2021-05-21 10:24:42,962 [DEBUG] {qconsumer} Shunning Motr thread
May 21 10:24:42 ssc-vm-1623.colo.seagate.com hare-hax[56895]: 2021-05-21 10:24:42,963 [INFO] {qconsumer} Handler thread has exited
May 21 10:24:42 ssc-vm-1623.colo.seagate.com hare-hax[56895]: 2021-05-21 10:24:42,963 [INFO] {MainThread} The http server has stopped
May 21 10:24:42 ssc-vm-1623.colo.seagate.com hare-hax[56895]: 2021-05-21 10:24:42,963 [INFO] {MainThread} Finalizing motr
May 21 10:24:43 ssc-vm-1623.colo.seagate.com hare-hax[56895]: 2021-05-21 10:24:43,034 [DEBUG] {MainThread} Motr API context is down. ByeMay 21 10:24:43 ssc-vm-1623.colo.seagate.com systemd[1]: Stopped HAX daemon for Hare.
```
cc @NeeravChoudhari

hax/hax/handler.py
andriytk: Is it changed in the latest patch to read processes status from Consul?

hax/hax/util.py
andriytk: Is this changed?

hax/hax/motr/__init__.py
andriytk: Makes sense now?

andriytk:
@mssawant let's resolve all the above conversations.

hax/hax/handler.py
mssawant: Yes. Hax now does not maintain a failed services list and fetches the status from Consul.

hax/hax/motr/__init__.py
mssawant: Not maintaining failed_svc anymore, but fetching statuses from Consul. So now if the confd status in Consul is not STARTED then we do not include that confd in the entrypoint reply.

hax/hax/util.py
mssawant: Yes, in the updated patched we have removed this.

hax/hax/motr/__init__.py
andriytk: Don't you think the same situation may happen also in this case? Do we have a STOPPING state for Motr processes in Consul KV?

mssawant: Yes, Every state reported by a motr process is updated in consul kv.

andriytk:
@mssawant could you please update the description of this PR at the top with elaborated problem description and the proposed fix?

hax/hax/motr/__init__.py
andriytk: Then it should be fine, I suppose...

hax/hax/hax.py
mssawant: Moved adopt and shun to hax.c to their respective functions. Restored the order of the threads.

mssawant:
@andriytk , have updated the commit message and the description.

andriytk:
> Why m0_rconfc_stop() hangs?

@mssawant it's still hard to understand why it hangs from the text in the description which follows the question. 

> When rconfc starts, it...

Yes, but we are talking about stop, not start, right? The following text describes rather how rconfc re-start works, but why stop hangs - it's still not quite clear. Is it possible to clarify this somehow?

mssawant:
@andriytk , sure let me try to further elaborate the description.

mssawant:
@andriytk , have updated the description a bit, tried to elaborate the situation during rconfc stop. Mentioning about rconfc start gives a bit of context to the reader, thus added. Kindly check. Thank you.

hax/hax/motr/__init__.py
```diff
@@ -148,52 +162,94 @@ def _entrypoint_request_cb(self, reply_context: Any, req_id: Any,
...
+            # endpoints. While doing this we need to handle the bootstrapping
+            # case, so we wait until bootstrapping is done that is all the
+            # motr services are up, we check the confd status and exclude
+            # corresponding confd from the entrypoint reply.
+            active_confds = []
+            if self.spiel_ready:
+                for confd in confds:
+                    if not util.is_confd_failed(confd.fid):
+                        active_confds.append(confd)
+                confds = active_confds
```
madhavemuri, line 228: @mssawant : I think this may cause issues when a confd (out of 3) is stopped/failed and then another motr entity (may be due to restart) requests entrypoint from Hax, in this case hax will send two confd's and a ne principal RM. But here **quorum is not possible**, thus failed confd too needs to be passed and make sure that failed  confd is not set as principal RM.

andriytk:
@mssawant , it's much better now, thanks!

> Hax maintains the motr process statusses in Consul KV, this saved status is checked
before replying to entrypoint request,..

So such check is not there in the current code? That's strange, and that's a clear bug. I suppose only adding this check would have solved the problem, no?

Anyway, good job! Approving...

hax/hax/motr/__init__.py
andriytk: Good point!

@mssawant, I think the logic of removing confds from entry-point replies has to be ON only during the shutdown.

mssawant:
Tested single node and multinode deployments,
**Single node**:
http://eos-jenkins.colo.seagate.com/job/Cortx-Deployment/job/VM-Deployment-R2-1Node/808/console

Deployment was successful.
Verifies post deployment status, services started fine and hare shutdown before starting was successful.
```
May 26 16:27:48 ssc-vm-4710.colo.seagate.com hare-hax[74773]: 2021-05-26 16:27:48,599 [DEBUG] {qconsumer} Shunning Motr thread
May 26 16:27:48 ssc-vm-4710.colo.seagate.com hare-hax[74773]: 2021-05-26 16:27:48,599 [INFO] {qconsumer} Handler thread has exited
May 26 16:27:48 ssc-vm-4710.colo.seagate.com hare-hax[74773]: 2021-05-26 16:27:48,600 [INFO] {MainThread} The http server has stopped
May 26 16:27:48 ssc-vm-4710.colo.seagate.com hare-hax[74773]: 2021-05-26 16:27:48,600 [INFO] {MainThread} Finalizing motr
May 26 16:27:48 ssc-vm-4710.colo.seagate.com hare-hax[74773]: 2021-05-26 16:27:48,674 [DEBUG] {MainThread} Motr API context is down. ByeMay 26 16:27:48 ssc-vm-4710.colo.seagate.com systemd[1]: Stopped HAX daemon for Hare.
```
Also tried `pcs cluster stop --all` and `pcs cluster start --all` sequence. 
Cluster stopped and started successfully.
Hax shutdown was successful.

**3-Node deployment**

3 node deployment was tried manually due to infra issues.
deployment was tried on 2 setups, deployment was successful on both the setups.
Pacemaker cluster started fine on first setup,
```
[root@ssc-vm-4701 ~]# pcs status
Cluster name: cortx_cluster
Stack: corosync
Current DC: srvnode-1.data.private (version 1.1.23-1.el7-9acf116022) - partition with quorum
Last updated: Wed May 26 16:11:30 2021
Last change: Wed May 26 14:52:43 2021 by root via cibadmin on srvnode-1.data.private

3 nodes configured
30 resource instances configured

Online: [ srvnode-1.data.private srvnode-2.data.private srvnode-3.data.private ]

Full list of resources:

 motr-free-space-mon    (systemd:motr-free-space-monitor):      Started srvnode-1.data.private
 s3backprod     (systemd:s3backgroundproducer): Started srvnode-2.data.private
 Clone Set: io_group-clone [io_group]
     Started: [ srvnode-1.data.private srvnode-2.data.private srvnode-3.data.private ]
 Clone Set: monitor_group-clone [monitor_group]
     Started: [ srvnode-1.data.private srvnode-2.data.private srvnode-3.data.private ]
 Resource Group: management_group
     mgmt-vip   (ocf::heartbeat:IPaddr2):       Started srvnode-3.data.private
     kibana     (systemd:kibana):       Started srvnode-3.data.private
     csm-agent  (systemd:csm_agent):    Started srvnode-3.data.private
     csm-web    (systemd:csm_web):      Started srvnode-3.data.private
```
```
[root@ssc-vm-4701 ~]# hctl status
Data pool:
    # fid name
    0x6f00000000000001:0x93 'storage-set-1__sns'
Profile:
    # fid name: pool(s)
    0x7000000000000001:0xd8 'Profile_the_pool': 'storage-set-1__sns' None None
Services:
    ssc-vm-4701.colo.seagate.com
    [started]  hax        0x7200000000000001:0x6   192.168.80.71@tcp:12345:1:1
    [started]  ioservice  0x7200000000000001:0x9   192.168.80.71@tcp:12345:2:1
    [started]  confd      0x7200000000000001:0x24  192.168.80.71@tcp:12345:2:2
    [started]  s3server   0x7200000000000001:0x27  192.168.80.71@tcp:12345:3:1
    [unknown]  m0_client  0x7200000000000001:0x2a  192.168.80.71@tcp:12345:4:1
    [unknown]  m0_client  0x7200000000000001:0x2d  192.168.80.71@tcp:12345:4:2
    [unknown]  m0_client  0x7200000000000001:0x30  192.168.80.71@tcp:12345:4:3
    ssc-vm-4702.colo.seagate.com  (RC)
    [started]  hax        0x7200000000000001:0x36  192.168.79.184@tcp:12345:1:1
    [started]  ioservice  0x7200000000000001:0x39  192.168.79.184@tcp:12345:2:1
    [started]  confd      0x7200000000000001:0x54  192.168.79.184@tcp:12345:2:2
    [started]  s3server   0x7200000000000001:0x57  192.168.79.184@tcp:12345:3:1
    [unknown]  m0_client  0x7200000000000001:0x5a  192.168.79.184@tcp:12345:4:1
    [unknown]  m0_client  0x7200000000000001:0x5d  192.168.79.184@tcp:12345:4:2
    [unknown]  m0_client  0x7200000000000001:0x60  192.168.79.184@tcp:12345:4:3
    ssc-vm-4711.colo.seagate.com
    [started]  hax        0x7200000000000001:0x66  192.168.79.175@tcp:12345:1:1
    [started]  ioservice  0x7200000000000001:0x69  192.168.79.175@tcp:12345:2:1
    [started]  confd      0x7200000000000001:0x84  192.168.79.175@tcp:12345:2:2
    [started]  s3server   0x7200000000000001:0x87  192.168.79.175@tcp:12345:3:1
    [unknown]  m0_client  0x7200000000000001:0x8a  192.168.79.175@tcp:12345:4:1
    [unknown]  m0_client  0x7200000000000001:0x8d  192.168.79.175@tcp:12345:4:2
    [unknown]  m0_client  0x7200000000000001:0x90  192.168.79.175@tcp:12345:4:3
```
On second setup, deployment was successful but cluster start faced problems,
confd on one of the node failed, fom hung was seen,
```
May 26 16:23:54 ssc-vm-4744.colo.seagate.com motr-server[42062]: motr[42062]:  dbf0   WARN  [fop/fom.c:362:hung_fom_notify]  FOP HUNG[[2May 26 16:24:32 ssc-vm-4744.colo.seagate.com motr-server[42062]: motr[42062]:  dbf0   WARN  [fop/fom.c:362:hung_fom_notify]  FOP HUNG[[3May 26 16:25:12 ssc-vm-4744.colo.seagate.com motr-server[42062]: motr[42062]:  dbf0   WARN  [fop/fom.c:362:hung_fom_notify]  FOP HUNG[[3May 26 16:25:51 ssc-vm-4744.colo.seagate.com motr-server[42062]: motr[42062]:  dbf0   WARN  [fop/fom.c:362:hung_fom_notify]  FOP HUNG[[3May 26 16:26:20 ssc-vm-4744.colo.seagate.com systemd[1]: m0d@0x7200000000000001:0x84.service stop-sigterm timed out. Killing.
May 26 16:26:20 ssc-vm-4744.colo.seagate.com systemd[1]: m0d@0x7200000000000001:0x84.service: main process exited, code=killed, status=9May 26 16:26:20 ssc-vm-4744.colo.seagate.com systemd[1]: Stopped m0d helper for 0x7200000000000001:0x84 service-FID.
``` 
and confd timed out, which also caused ioservice to stop.
This seems to be related to https://jts.seagate.com/browse/EOS-20697

Thus, I think the patch is ready to merge.

mssawant:
@andriytk,
> I suppose only adding this check during the shutdown would have solved the problem, no?

No, because we still want to report M0_NC_FAILED for all the motr processes in the cluster in every node. So all the rpc connections from motr processes are also cancelled. Otherwise we see problems where Motr processes try to disconnect themselves from other motr processes and the connection/session termination rpc requests hang and delay the stopping processes.

hax/hax/motr/__init__.py
mssawant: @madhavemuri, yes, agree, but problem of rconfc happens even before hax shutdown is initiated. So when hax starts sending M0_NC_FAILED requests to itself and whichever alive motr processes, entrypointclient in hax's motr land starts trying to reconnect and starts sending entrypoint requests in a loop. Hax is not stopping yet, so we don't have a flag or condition that can be set here. Thus, we are checking motr statuses before replying to entrypoint requests.

mssawant: Now ideally, if all the processes are marked as failed rconfc must not send any entrypoint requests. But thats not how rconfc works right, it tries in a hope that at some point atleast one of the confds can come back up and rconfc will get the entrypoint reply. If you can let me know how to break this loop then I can happily revert this change.

andriytk:
> we still want to report M0_NC_FAILED for all the motr processes in the cluster in every node

That's what I thought we are doing today already...

hax/hax/motr/__init__.py
andriytk: > Hax is not stopping yet, so we don't have a flag or condition that can be set here.

But hax do know that it's going to stop, right? (Because this all happens when we are trying to stop it.) So why doesn't it check and remove confds from entrypoint replies only when it's going to stop?

mssawant: for e.g., I tried not filtering out failed confds at all, in that case rconfc_stop hangs and hax times out.
rconfc after receiving failed event for RC, sends entrypoint request to hax, now if the remaining confds have not stopped yet and entrypoint is replied and hax sends failed event after that, it does not help as rconfc already has started re-initialization and has sent RM borrow request, which hangs rconfc stop later.
```
May 26 12:55:37 ssc-vm-1624.colo.seagate.com hare-hax[11622]: 2021-05-26 12:55:37,954 [DEBUG] {qconsumer} Broadcasting HA states [HAState(fid=0x7200000000000001:0x9, status=STOPPED), HAState(fid=0x7200000000000001:0x37, status=OFFLINE), HAState(fid=0x7200000000000001:0x65, status=STOPPED)] over ha_link
May 26 12:55:37 ssc-vm-1624.colo.seagate.com hare-hax[11622]: 2021-05-26 12:55:37,959 [DEBUG] {qconsumer} Process fid=0x7200000000000001:0x9 encloses 2 services as follows: [FidWithType(fid=0x7300000000000001:0xb, service_type='confd'), FidWithType(fid=0x7300000000000001:0xa, service_type='rms')]
May 26 12:55:37 ssc-vm-1624.colo.seagate.com hare-hax[11622]: 2021-05-26 12:55:37,965 [DEBUG] {qconsumer} Process fid=0x7200000000000001:0x65 encloses 2 services as follows: [FidWithType(fid=0x7300000000000001:0x67, service_type='confd'), FidWithType(fid=0x7300000000000001:0x66, service_type='rms')]
May 26 12:55:37 ssc-vm-1624.colo.seagate.com hare-hax[11622]: motr[11622]:  b900   WARN  [ha/halon/interface.c:1111:m0_halon_interface_send]  hl=0x7fd904009360 ep=10.230.250.53@tcp:12345:1:1 epoch=0 tag=71 type=2
May 26 12:55:37 ssc-vm-1624.colo.seagate.com hare-hax[11622]: motr[11622]:  b900   WARN  [ha/halon/interface.c:1111:m0_halon_interface_se_send]  hl=0x7fd904046150 ep=10.230.250.53@tcp:12345:2:1 epoch=0 tag=65 type=2
May 26 12:55:37 ssc-vm-1624.colo.seagate.com hare-hax[11622]: motr[11622]:  27e0   WARN  [ha/halon/interface.c:335:halon_interface_process_failure_check]  disconnecting all links to no_id=<7200000000000001:9>
May 26 12:55:37 ssc-vm-1624.colo.seagate.com hare-hax[11622]: motr[11622]:  27e0   WARN  [ha/halon/interface.c:335:halon_interface_process_failure_check]  disconnecting all links to no_id=<7200000000000001:65>

May 26 12:55:37 ssc-vm-1624.colo.seagate.com hare-hax[11622]: 2021-05-26 12:55:37,977 [DEBUG] {Dummy-1} Received entrypoint request from remote endpoint '10.230.250.53@tcp:12345:1:1'
May 26 12:55:37 ssc-vm-1624.colo.seagate.com hare-hax[11622]: 2021-05-26 12:55:37,977 [DEBUG] {Dummy-1} enqueue entrypoint request for 10.230.250.53@tcp:12345:1:1
May 26 12:55:37 ssc-vm-1624.colo.seagate.com hare-hax[11622]: 2021-05-26 12:55:37,977 [DEBUG] {Dummy-1} Processing entrypoint request from remote endpoint '10.230.250.53@tcp:12345:1:1'
May 26 12:55:38 ssc-vm-1624.colo.seagate.com hare-hax[11622]: motr[11622]:  1900   WARN  [ha/halon/interface.c:1073:m0_halon_interface_entrypoint_reply]  req_id=0:d rc=0 confd_nr=3 confd_quorum=2 rm_fid=<7300000000000001:a> rm_eps=10.230.250.51@tcp:12345:2:1

May 26 12:55:39 ssc-vm-1624.colo.seagate.com hare-hax[11622]: 2021-05-26 12:55:39,171 [DEBUG] {qconsumer} Setting process status in KV: ssc-vm-1624.colo.seagate.com/processes/0x7200000000000001:0x37:{"state": "M0_CONF_HA_PROCESS_STOPPED"}
May 26 12:55:39 ssc-vm-1624.colo.seagate.com hare-hax[11622]: 2021-05-26 12:55:39,176 [DEBUG] {qconsumer} Broadcasting HA states [HAState(fid=0x7200000000000001:0x37, status=FAILED)] over ha_link
May 26 12:55:39 ssc-vm-1624.colo.seagate.com hare-hax[11622]: 2021-05-26 12:55:39,182 [DEBUG] {qconsumer} Process fid=0x7200000000000001:0x37 encloses 2 services as follows: [FidWithType(fid=0x7300000000000001:0x39, service_type='confd'), FidWithType(fid=0x7300000000000001:0x38, service_type='rms')]
May 26 12:55:39 ssc-vm-1624.colo.seagate.com hare-hax[11622]: motr[11622]:  b300   WARN  [ha/halon/interface.c:1111:m0_halon_interface_send]  hl=0x7fd904009360 ep=10.230.250.53@tcp:12345:1:1
May 26 12:55:39 ssc-vm-1624.colo.seagate.com hare-hax[11622]: 2021-05-26 12:55:39,182 [DEBUG] {qconsumer} Broadcast HA state complete with the following message_ids = [MessageId(0x7fd904009360, 75)]

May 26 13:00:37 ssc-vm-1624.colo.seagate.com hare-hax[11622]: 2021-05-26 13:00:37,273 [INFO] {Dummy-1} Delivered to endpoint'10.230.250.
May 26 13:00:37 ssc-vm-1624.colo.seagate.com hare-hax[11622]: 2021-05-26 13:00:37,273 [DEBUG] {qconsumer} Stopping rconfc
May 26 13:02:07 ssc-vm-1624.colo.seagate.com systemd[1]: hare-hax.service stop-sigterm timed out. Killing.
May 26 13:02:07 ssc-vm-1624.colo.seagate.com systemd[1]: hare-hax.service: main process exited, code=killed, status=9/KILL
May 26 13:02:07 ssc-vm-1624.colo.seagate.com systemd[1]: Stopped HAX daemon for Hare.
May 26 13:02:07 ssc-vm-1624.colo.seagate.com systemd[1]: Unit hare-hax.service entered failed state.
May 26 13:02:07 ssc-vm-1624.colo.seagate.com systemd[1]: hare-hax.service failed.

```

m0trace
```
163461  31519887.955.007864  77fc0e0  CALL    rm/rm.c:2650:borrow_send                            > incoming: 0x7fd8e8010460 credit: 1
407317  31520650.561.903180  effb960  CALL    spiel/spiel.c:93:m0_spiel_rconfc_stop               >
407329  31520650.561.904414  d3b5930  CALL    fop/fop.c:182:m0_fop_put                            > fop: 0x7fd8f00601d0 HA link msg fop

422945  31520848.466.817366  d3b5930  CALL    fop/fop.c:182:m0_fop_put                            > fop: 0x7fd99805cb50 Credit Borrow [5422946  31520848.466.818890  d3b5930  CALL    fop/fop.c:188:m0_fop_put                            <
422947  31520848.466.823502  d3b5950  DEBUG   rpc/session.c:797:m0_rpc_session_release            session 0x7fd8e800fed8 1 -> 0
422948  31520848.466.827382  d3b58f0  INFO    rpc/session.c:137:session_state_set                 Session 0x7fd8e800fed8: Busy -> Idle
422949  31520848.466.829924  d3b5910  DEBUG   rpc/item.c:737:m0_rpc_item_change_state             0x7fd99805cb70 [REQUEST/M0_RM_FOP_BORROW(88)] xid=1 SENT -> WAITING_FOR_REPLY sm=Outgoing-RPC-Item-sm
```

mssawant:
@andriytk,
>I thought we are doing this in the current code already...

We were doing it but not for all the processes in the cluster on every node, thats what is changed.

hax/hax/motr/__init__.py
madhavemuri: I thought this was due to Hax is not sending FAILED event as part of nvec reply.  Have you fixed note vec part and checked it ?

mssawant: Yes, that is fixed. Now hax is fetching processes and services statuses from Consul and reporting in nvec reply.
The problem here is when nvec is replied, it is possible that not all the processes have stopped. Thus the nvec reply can still contain M0_NC_ONLINE for some of the services.

andriytk: Mandar, it's not the only problem here. Going back to the beginning of this thread... Currently, the semantics of entrypoint reply implies that all configured confds are listed there, not only those which are online. If we are about to break/change this semantics for the normal cluster work (i.e. not shutdown) - we should carefully revise the rconfc logic and test all the possible scenarios.

For example, consider that one of 3 confds goes down (the main one) and rconfc in all processes now gets reply with only 2 confds which are online. Now, when the failed confd restarts - how will it be updated in rconfcs list in all processes?

mssawant: @andriytk, whenever principal rm confd fails, rconfc invalidates and restarts. As part of restarting it requests new entrypoint information, in response to this new entrypoint request, rconfc will get the restarted confd.
But in general, we need not filter out confds and everything must work fine after marking the respective confds and their services as FAILED. But this is not the case presently. 

This is what is happening,
Tried again without filtering out the failed/stopped confds,
M0_NC_FAILED, was sent in nvec reply to confd,
```
324770  31728444.966.378542  0ff86d0  DEBUG   ha/note.c:225:m0_ha_msg_accept                      nv_note[0]=(no_id=<7200000000000001:9>
g_accept                      nv_note[1]=(no_id=<7300000000000001:a> no_state=2)
g_accept                      nv_note[2]=(no_id=<7200000000000001:65> no_state=2)
g_accept                      nv_note[3]=(no_id=<7200000000000001:37> no_state=2)
g_accept                      nv_note[4]=(no_id=<7400000000000001:0> no_state=1)
_accept                       > confc=0x16dd548 note->nv_nr=5
_accept                       nv_note[0]=(no_id=<7200000000000001:9> no_state=2) obj=0x7f633404b6a0 obj->co_status=2
nfc_herd_link__on_death_cb    > lnk=0x7f6334045730 ep:10.230.250.51@tcp:12345:2:1 rc 0 state:0
post                          grp=0x173b3e0 forkq_push: 0x7f6334046e38 -> 0x7f634184b480
nfc_herd_link__on_death_cb    < lnk=0x7f6334045730 co_ha_state = M0_NC_FAILED, return false
_accept                       nv_note[1]=(no_id=<7300000000000001:a> no_state=2) obj=0x7f6334043b40 obj->co_status=2
_accept                       nv_note[2]=(no_id=<7200000000000001:65> no_state=2) obj=0x7f633405b0e0 obj->co_status=2
nfc_herd_link__on_death_cb    > lnk=0x7f6334053520 ep:10.230.250.48@tcp:12345:2:1 rc 0 state:0
post                          grp=0x173b3e0 forkq_push: 0x7f6334054c28 -> 0x7f6334046e38
nfc_herd_link__on_death_cb    < lnk=0x7f6334053520 co_ha_state = M0_NC_FAILED, return false
_accept                       nv_note[3]=(no_id=<7200000000000001:37> no_state=2) obj=0x7f63340533f0 obj->co_status=2
nfc_herd_link__on_death_cb    > lnk=0x7f633404b9a0 ep:10.230.250.53@tcp:12345:2:1 rc 0 state:2
post                          grp=0x173b3e0 forkq_push: 0x7f633404d0a8 -> 0x7f6334054c28
nfc_herd_link__on_death_cb    < lnk=0x7f633404b9a0 co_ha_state = M0_NC_FAILED, return false
_accept                       nv_note[4]=(no_id=<7400000000000001:0> no_state=1) obj=0x7f6288003790 obj->co_status=0
```
still borrow fop was sent,
```
327468  31728444.973.059312  07f7500  CALL    rpc/item.c:1141:m0_rpc_item_send                    > 0x7f633405bb20 [REQUEST/M0_RM_FOP_BORROW(88)] dest_ep=10.230.250.51@tcp:12345:2:1 ri_session=0x7f6288018
```
This timed out rconfc stop
Here is where the borrow request is sent,
```
327341  31728444.972.843844  07f7c40  CALL    rm/rm.c:915:m0_rm_owner_creditor_reset              <
327342  31728444.972.846292  0ff8700  CALL    ha/halon/interface.c:535:halon_interface_link_disconnected_cb   > hii=0x16ceac0 ha=0x16d82
327343  31728444.972.846372  07f7c90  CALL    conf/rconfc.c:1053:rlock_ctx_creditor_setup         < rc=0
327344  31728444.972.846734  07f7c90  CALL    conf/rconfc.c:1873:rconfc_read_lock_get             > rconfc = 0x16dc530
327345  31728444.972.847666  0ff8700  WARN    ha/halon/interface.c:537:halon_interface_link_disconnected_cb   hl=0x7f629c048000 ep=10.23
327346  31728444.972.866520  07f7c40  DEBUG   conf/rconfc.c:1189:rconfc_state_set                 rconfc: 0x16dc530, M0_RCS_CREDITOR_SET -> M0_RCS_GET_RLOCK
327347  31728444.972.868874  07f7c40  CALL    rm/rm_rwlock.c:554:m0_rm_rwlock_req_init            >
327348  31728444.972.869342  07f7bd0  CALL    rm/rm.c:1066:m0_rm_incoming_init                    > incoming: 0x7f6288019090 for owner: 
327349  31728444.972.871446  07f7b80  CALL    rm/rm.c:967:m0_rm_credit_init                       > credit: 0x7f6288019110 with owner: 
0327350  31728444.972.875430  07f7b80  CALL    rm/rm.c:981:m0_rm_credit_init                       <
327351  31728444.972.876738  07f7bd0  CALL    rm/rm.c:1078:m0_rm_incoming_init                    <
327352  31728444.972.877010  07f7c40  CALL    rm/rm_rwlock.c:563:m0_rm_rwlock_req_init            <
327353  31728444.972.878918  07f7c30  CALL    rm/rm.c:1763:m0_rm_credit_get                       > owner: 0x7f6288018cc0 credit: 1
327354  31728444.972.879498  07f7ba0  DEBUG   sm/sm.c:158:m0_sm_asts_run                          grp=0x7f62880194d0 forkq_pop: 0x7f6341
327355  31728444.972.880886  07f7890  CALL    rm/rm.c:1947:owner_balance
327357  31728444.972.883752  07f7890  CALL    rm/rm.c:2048:incoming_check                         >
327358  31728444.972.883888  07f7840  CALL    rm/rm.c:967:m0_rm_credit_init                       > credit: 0x7f63307f7b90 with owner: 0327359  31728444.972.884044  07f7840  CALL    rm/rm.c:981:m0_rm_credit_init                       <
327360  31728444.972.884808  07f7890  CALL    rm/rm.c:2359:incoming_check_with                    > incoming: 0x7f6288019090 credit: 1
327361  31728444.972.885374  07f7890  CALL    rm/rm.c:2650:borrow_send                            > incoming: 0x7f6288019090 credit: 1
```
This blocks rconfc stop.
I will create another ticket to investigate why we need to filter out confds and why it just does not work with notifying M0_NC_FAILED for the services.
One descrepency that I see here is motr is not requesting state of confd service in the nvec request but just confd process and rm service. Need to further investigate if that would somehow make rconfc miss that the confd is already down. Probably some process state checking is missing.
But I think, let's complete this PR and investigate this issue in another ticket.

madhavemuri: This will be investigated further in motr  https://jts.seagate.com/browse/EOS-20589 and from Hare point of as well.

PR #1608, opened 2021-05-04
EOS-18914: hctl bootstrap permissions error

supritshinde:
Solution: Added two new functions named ssh_safe and conf_dir_check. one for generic ssh purpose that would also show on which node in cluster is problematic and confdir for checking write access to the peer nodes. 

### DIFF
--- a/utils/hare-bootstrap
+++ b/utils/hare-bootstrap
@@ -29,14 +29,11 @@ usage() {
     cat <<EOF
 Usage: $PROG [<option>]... <CDF>
        $PROG [<option>]... --conf-dir <dir>
-
 Bootstrap the cluster.
-
 Positional arguments:
   <CDF>                  Path to the cluster description file.
   -c, --conf-dir <dir>   Don't generate configuration files, use existing
                          ones from the specified directory.
-
 Options:
   --debug      Print commands and their arguments as they are executed.
   --mkfs       Execute m0mkfs.  *CAUTION* This wipes all Motr data!
@@ -51,6 +48,32 @@ die() {
 
 say() {
     echo -n "$(date '+%F %T'): $*"
+}
+
+ssh_error_info() {
+    local hostname=$1
+    shift
+    ssh -n $hostname "$@" ||
+    {
+        echo "Error at $hostname with command $@";
+        exit 1
+    }
+}
+
+conf_dir_perm_check() {
+    ssh -n $node """
+    if ! [[ -d $conf_dir ]]; then
+        echo "$conf_dir directory does not exist on $node \
+Try reinstalling Hare on the above mentioned node.";
+        exit 1
+    fi
+
+    if ! [[ -w $conf_dir ]]; then
+        echo "Cannot write to $conf_dir directory of $node";
+        exit 1
+    fi
+    echo $node > $conf_dir/node-name;
+"""
 }
 
 get_server_nodes() {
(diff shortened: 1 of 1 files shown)


supritshinde:
Once Review is done. I'll convert all the commits into one and push a final patch. I had to do few updations.

supritshinde:
This PR is final alternative to https://github.com/Seagate/cortx-hare/pull/1596 
Hence, Closing that PR.

utils/hare-bootstrap
```diff
@@ -53,6 +50,29 @@ say() {
     echo -n "$(date '+%F %T'): $*"
 }
 
+ssh_safe() {
```
mssawant, line 53: suggestion: s/ssh_safe/ssh_error_info ( we are not doing anything from security perspective as such :))

utils/hare-bootstrap
```diff
@@ -53,6 +50,29 @@ say() {
...
+    ssh -n $node $* ||
```
mssawant, line 54: where is this `node` argument coming from, this function is not taking a node argument?

utils/hare-bootstrap
```diff
@@ -117,7 +137,7 @@ abort_if_RC_leader_election_is_impossible() {
         if [[ $node == $(node-name) ]]; then
             ssh=
         else
-            ssh="ssh $node"
+            ssh="ssh_safe"
```
mssawant, line 140: don't you need to pass `$node` to `ssh_safe` too?

utils/hare-bootstrap
```diff
@@ -275,12 +291,12 @@ if sudo systemctl --quiet is-active hare-consul-agent; then
     say 'Stopping Consul on other nodes...'
     pids=()
     while read node bind_ip; do
-        ssh $node "sudo systemctl stop hare-consul-agent" &
+        ssh_safe "sudo systemctl stop hare-consul-agent" &
```
mssawant, line 294: same here, don't you need to pass `$node` here as well?

utils/hare-bootstrap
supritshinde: Ah yes. missed this. blunder. here it must be ssh $node or sshsafe with node passed to it. right.

utils/hare-bootstrap
supritshinde: yeah. This seems much better. I'll change it to ssh_error_info

utils/hare-bootstrap
supritshinde: I don't think so it explicitly needs to be passed. It can access the $node in that while do logic. the jq - ... if i'm not wrong. Still i'll test it myself.

utils/hare-bootstrap
supritshinde: Done.

utils/hare-bootstrap
supritshinde: Changed the ssh funtion.

utils/hare-bootstrap
supritshinde: Made changes to ssh function.

utils/hare-bootstrap
supritshinde: Altered the ssh function to take explicit args and refined the script.

supritshinde:
Final Changes as suggested are made. Please Review and Approve.

mssawant (approved):

PR #1612, opened 2021-05-06
EOS-19373 Parallel message processing in Hax

knekrasov:
Jira ticket: [EOS-19373](https://jts.seagate.com/browse/EOS-19373) 

Currently we have only one Motr-aware thread (ConsumerThread) that reads out `queue.Queue`, gets BaseMessage from it and interacts with Motr layer accordingly. In bigger installations one thread simply doesn't manage to process the flow of the messages fast (so that entrypoint requests can start delaying which is critical for Motr API).

As a solution. we want to add more consumer threads. There is one problem though: we can't blindly process the ordered messages in parallel to each other, sometimes the order matters. This patch proposes splitting the whole stream of such messages into continous partitions (or groups). Messages within the same partition can be handled in parallel without an issue. If a message belongs to one of the "next" partitions, that processing gets delayed until the previous partitions are processed fully.

More details on the algorithm and terminology can be found here: https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/178421786/22%2BPARHAX%2BParallel%2Bmessage%2Bprocessing%2Bin%2BHaX#Proposed-solution (Seagate internal).

### DIFF
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -17,8 +17,6 @@
 #
 
 import logging
-import time
-from queue import Empty, Queue
 from typing import List
 
 from hax.message import (BroadcastHAStates, EntrypointRequest,
@@ -30,6 +28,7 @@ from hax.message import (BroadcastHAStates, EntrypointRequest,
                          StobIoqError)
 from hax.motr import Motr
 from hax.motr.delivery import DeliveryHerald
+from hax.motr.planner import WorkPlanner
 from hax.queue.publish import EQPublisher
 from hax.types import (ConfHaProcess, HaLinkMessagePromise, HAState, MessageId,
                        ObjT, ServiceHealth, StoppableThread, m0HaProcessEvent,
@@ -48,11 +47,11 @@ class ConsumerThread(StoppableThread):
     The thread exits gracefully when it receives message of type Die (i.e.
     it is a 'poison pill').
     """
-    def __init__(self, q: Queue, motr: Motr, herald: DeliveryHerald,
-                 consul: ConsulUtil):
+    def __init__(self, planner: WorkPlanner, motr: Motr,
+                 herald: DeliveryHerald, consul: ConsulUtil):
         super().__init__(target=self._do_work,
                          name='qconsumer',
-                         args=(q, motr))
+                         args=(planner, motr))
         self.is_stopped = False
         self.consul = consul
         self.eq_publisher = EQPublisher()
@@ -62,7 +61,7 @@ class ConsumerThread(StoppableThread):
         self.is_stopped = True
 
     @repeat_if_fails(wait_seconds=1)
-    def _update_process_status(self, q: Queue, motr: Motr,
+    def _update_process_status(self, p: WorkPlanner, motr: Motr,
                                event: ConfHaProcess) -> None:
         # If a consul-related exception appears, it will
         # be processed by repeat_if_fails.
@@ -78,7 +77,7 @@ class ConsumerThread(StoppableThread):
                 notify_devices=False)
 
     @repeat_if_fails(wait_seconds=1)
-    def update_process_failure(self, q: Queue,
+    def update_process_failure(self, planner: WorkPlanner,
                                ha_states: List[HAState]) -> List[HAState]:
         new_ha_states: List[HAState] = []
         for state in ha_states:
@@ -118,7 +117,7 @@ class ConsumerThread(StoppableThread):
                     # the real time status of the process before
                     # broadcasting failure.
                     current_status = ServiceHealth.UNKNOWN
-                    q.put(
+                    planner.add_command(
                         BroadcastHAStates(states=[
                             HAState(fid=state.fid, status=ServiceHealth.FAILED)
                         ],
@@ -147,29 +146,19 @@ class ConsumerThread(StoppableThread):
                 new_ha_states.append(state)
         return new_ha_states
 
-    def _do_work(self, q: Queue, motr: Motr):
+    def _do_work(self, planner: WorkPlanner, motr: Motr):
         LOG.info('Handler thread has started')
         motr.adopt_motr_thread()
-
-        def pull_msg():
-            try:
-                return q.get(block=False)
-            except Empty:
-                return None
 
         try:
             while True:
                 try:
                     LOG.debug('Waiting for the next message')
 
-                    item = pull_msg()
-                    while item is None:
-                        time.sleep(0.2)
-                        if self.is_stopped:
-                            raise StopIteration()
-                        item = pull_msg()
+                    item = planner.get_next_command()
+                    planner.ensure_allowed(item)
 
-                    LOG.debug('Got %s message from queue', item)
+                    LOG.debug('Got %s message from planner', item)
                     if isinstance(item, FirstEntrypointRequest):
                         LOG.debug('first entrypoint request, broadcast FAILED')
                         ids: List[MessageId] = motr.broadcast_ha_states(
@@ -197,7 +186,7 @@ class ConsumerThread(StoppableThread):
                         # hence will need to make new attempt by itself
                         motr.send_entrypoint_request_reply(item)
                     elif isinstance(item, ProcessEvent):
-                        self._update_process_status(q, motr, item.evt)
+                        self._update_process_status(planner, motr, item.evt)
                     elif isinstance(item, HaNvecGetEvent):
                         fn = motr.ha_nvec_get_reply
                         # If a consul-related exception appears, it will
@@ -209,7 +198,8 @@ class ConsumerThread(StoppableThread):
                         decorated(item)
                     elif isinstance(item, BroadcastHAStates):
                         LOG.info('HA states: %s', item.states)
-                        ha_states = self.update_process_failure(q, item.states)
+                        ha_states = self.update_process_failure(
+                            planner, item.states)
                         result: List[MessageId] = motr.broadcast_ha_states(
                             ha_states)
                         if item.reply_to:
@@ -264,6 +254,8 @@ class ConsumerThread(StoppableThread):
                 except Exception:
                     # no op, swallow the exception
                     LOG.exception('**ERROR**')
+                finally:
+                    planner.notify_finished(item)
         except StopIteration:
             LOG.info('Consumer Stopped')
             motr.stop()
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -48,9 +48,9 @@ class ConsumerThread(StoppableThread):
     it is a 'poison pill').
     """
     def __init__(self, planner: WorkPlanner, motr: Motr,
-                 herald: DeliveryHerald, consul: ConsulUtil):
+                 herald: DeliveryHerald, consul: ConsulUtil, idx: int):
         super().__init__(target=self._do_work,
-                         name='qconsumer',
+                         name=f'qconsumer-{idx}',
                          args=(planner, motr))
         self.is_stopped = False
         self.consul = consul
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -156,9 +156,6 @@ class ConsumerThread(StoppableThread):
                     LOG.debug('Waiting for the next message')
 
                     item = planner.get_next_command()
-                    # TODO maybe it makes sense to move ensure_allowed() into
-                    # get_next_command()?
-                    item = planner.ensure_allowed(item)
 
                     LOG.debug('Got %s message from planner', item)
                     if isinstance(item, FirstEntrypointRequest):
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -21,18 +21,19 @@ import time
 from queue import Empty, Queue
 from typing import List
 
-from hax.message import (BroadcastHAStates, FirstEntrypointRequest,
-                         EntrypointRequest, HaNvecGetEvent,
-                         ProcessEvent, SnsRebalancePause, SnsRebalanceResume,
+from hax.message import (BroadcastHAStates, EntrypointRequest,
+                         FirstEntrypointRequest, HaNvecGetEvent, ProcessEvent,
+                         SnsRebalancePause, SnsRebalanceResume,
                          SnsRebalanceStart, SnsRebalanceStatus,
                          SnsRebalanceStop, SnsRepairPause, SnsRepairResume,
-                         SnsRepairStart, SnsRepairStatus, SnsRepairStop)
+                         SnsRepairStart, SnsRepairStatus, SnsRepairStop,
+                         StobIoqError)
 from hax.motr import Motr
 from hax.motr.delivery import DeliveryHerald
 from hax.queue.publish import EQPublisher
-from hax.types import (ConfHaProcess, HAState, MessageId, StobIoqError,
-                       ServiceHealth, m0HaProcessEvent, m0HaProcessType,
-                       StoppableThread, HaLinkMessagePromise, ObjT)
+from hax.types import (ConfHaProcess, HaLinkMessagePromise, HAState, MessageId,
+                       ObjT, ServiceHealth, StoppableThread, m0HaProcessEvent,
+                       m0HaProcessType)
 from hax.util import ConsulUtil, dump_json, repeat_if_fails
 
 LOG = logging.getLogger('hax')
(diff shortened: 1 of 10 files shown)


hax/hax/motr/planner.py
```diff
@@ -0,0 +1,207 @@
...
+                    return self._create_poison()
+                if is_current(command, state):
+                    state.taken_commands.remove(command)
+                    state.active_commands.add(command)
+                    return command
+                LOG.log(TRACE, '[WP]Cmd %s not allowed for now.'
+                        ' Current state: %s', command, self.state)
+                self.b_lock.wait()
+
+    def _inc_group(self):
```
mssawant, line 131: - How many groups are expected to present at a given instance?
- Will the number of groups vary? If not then it would be good have an invariant that asserts the expected state of the planner at any given instance during its lifetime.

hax/hax/motr/planner.py
```diff
@@ -0,0 +1,207 @@
...
+            if state.active_commands:
+                return
+            for c in self.backlog:
+                if c.group == state.current_group_id:
+                    return
+            for c in state.taken_commands:
+                if c.group == state.current_group_id:
+                    return
+            # if we're here, command was the only one belonging to group
+            self._inc_group()
```
mssawant, line 154: When does the active group change? is there an invariant that must be satisfied at any given instance of time?

hax/hax/motr/planner.py
```diff
@@ -0,0 +1,207 @@
...
+            LOG.log(TRACE, '[WP]Active group changed to %s',
+                    state.current_group_id)
+            # The group changed, let's unblock those who are waiting for
+            # this group
+            self.b_lock.notifyAll()
+
+    def _should_increase_group(self, cmd: BaseMessage) -> bool:
```
mssawant, line 161: How about fix set of groups with an allowed_message bitmap? We have fixed set of commands, we can have a fixed set of groups where each group takes only those messages (commands) which are allowed by the allowed_message bitmap in the group. This can be more efficient than searching in the list and we can also check the invariant easily.

hax/hax/hax.py
```diff
@@ -28,6 +27,7 @@
 from hax.motr import Motr
 from hax.motr.delivery import DeliveryHerald
 from hax.motr.ffi import HaxFFI
+from hax.motr.planner import WorkPlanner
```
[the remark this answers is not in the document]
knekrasov: In design document this thing is called WorkPlanner, I didn't come up to a better name :-)

Scheduler sounds to me quite the same to planner (scheduler is more like connected with a calendar - we don't have it, planner is just about making a plan). This thing is a synchronization point that a kind of "plans" which work can be done right now and which cannot. And when a piece of work becomes eligible for processing, it will wake up the blocked thread.

hax/hax/motr/planner.py
knekrasov: That's a good point. After I wrote this I realized that the group index can grow too fast which is a problem namely in Python. It uses long integer maths just by default, so, for instance, if an integer number becomes greater than 4 bytes can hold, more memory will be used (no integer overflow will happen). And the fact is that we definitely don't need such huge numbers.

I'll update the PR to address that problem.

hax/hax/motr/planner.py
knekrasov: When group has no more commands, we consider this group as finished. It may happen that the next (n+1) group already has commands. So by incrementing group index, we unblock those commands and they start executing.

hax/hax/motr/planner.py
knekrasov: I'll check if standard pythonic `set` can hold types. That would be simpler than maintaining bitmap, map the types to integer constants etc.

hax/hax/motr/planner.py
```diff
@@ -0,0 +1,207 @@
...
+    #
+    # group being executed currently
+    current_group_id: int
+    #
+    # group that is being populated by next add_command() invocation
+    next_group_id: int
+    #
+    # commands that are being processed now
+    active_commands: LinkedSet[BaseMessage]
+    taken_commands: LinkedSet[BaseMessage]
```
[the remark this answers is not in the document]
knekrasov: Added comments

hax/hax/motr/planner.py
```diff
@@ -0,0 +1,207 @@
...
+        with self.b_lock:
+            cmd = self._assign_group(command)
+            LOG.log(TRACE, '[WP]Cmd %s is added. Current state: %s', cmd,
+                    self.state)
+            self.backlog.append(cmd)
+            # Some threads may be waiting because of an empty backlog - let's
+            # notify them
+            self.b_lock.notifyAll()
+
+    def _create_poison(self) -> BaseMessage:
```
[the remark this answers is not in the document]
knekrasov: Added docstrings

hax/hax/motr/planner.py
knekrasov: Addressed, improved unit tests

hax/hax/motr/planner.py
knekrasov: Addressed. Standard `set()` is used for types given they are hashable.

hax/hax/hax.py
```diff
@@ -100,15 +100,16 @@ def main():
...
-                consul_util=util)
+    motr = Motr(planner=planner, ffi=ffi, herald=herald, consul_util=util)
 
     # Note that consumer thread must be started before we invoke motr.start(..)
     # Reason: hax process will send entrypoint request and somebody needs
     # to reply it.
-    consumer = _run_qconsumer_thread(q, motr, herald, util)
+
+    # TODO make the number of threads configurable
+    consumer_threads = [
```
mssawant, line 110: I think we can still configure number of consumer threads instead of a constant, anyway execution of threads is restricted to the number of cpus, so how about creating threads equal to `os.cpu_count()`?

mssawant: But we should make sure that the thread execution does not block.

hax/hax/motr/planner.py
```diff
@@ -0,0 +1,297 @@
...
+    #
+    # group that is being populated by next add_command() invocation
+    next_group_id: int
+    #
+    # commands that are being processed now
+    active_commands: LinkedList[BaseMessage]
+    #
+    # The commands that have already been taken by the worker threads
+    # but not approved yet for execution (so either threads haven't invoked
+    # ensure_allowed() yet or ensure_allowed() has not succeded for them).
```
mssawant, line 50: So what happens to the commands for whom ensure_allowed() fails? do they get assigned to a different group then?

hax/hax/motr/planner.py
```diff
@@ -0,0 +1,297 @@
...
+__all__ = ['WorkPlanner']
+
+
+@dataclass
+class State:
+    #
+    # group being executed currently
+    current_group_id: int
+    #
+    # group that is being populated by next add_command() invocation
```
mssawant, line 42: s/group that../Group that... .

hax/hax/motr/planner.py
```diff
@@ -0,0 +1,297 @@
...
+           mode to stop worker threads gracefully.
+        '''
+
+        cmd = Die()
+        # Since it is a special case, no _assign_group() will be invoked to
+        # find the proper group. In fact, we want to stop the threads as soon
+        # as possible, that's why it doesn't make sense to postpone the
+        # command. In other words, it must belong to the group which is
+        # currently active.
+        cmd.group = self.state.current_group_id
```
mssawant, line 133: Can current_group_id change while the Die command is being assigned the group?
How about creating a group0 that can execute such high priority commands?

hax/hax/handler.py
```diff
@@ -107,45 +106,39 @@ def update_process_failure(self, q: Queue,
...
-                    item = pull_msg()
-                    while item is None:
-                        time.sleep(0.2)
-                        if self.is_stopped:
-                            raise StopIteration()
-                        item = pull_msg()
+                    item = planner.get_next_command()
+                    # TODO maybe it makes sense to move ensure_allowed() into
+                    # get_next_command()?
+                    item = planner.ensure_allowed(item)
```
mssawant, line 133: Yes, I think it would be good to merge get_next_command() and ensure_allowed() so the the user of get_next_command() need not have to have knowledge about one more step. The command returned by get_next_command() can be executed directly.

hax/hax/motr/planner.py
```diff
@@ -0,0 +1,297 @@
...
+        if not self.state.next_group_commands:
+            # current group is empty -> join it freely
+            return False
+        if isinstance(cmd, HaNvecGetEvent):
+            # HaNvecGetEvent can be done in parallel to any other commands.
+            # No need to form the new group for it.
+            return False
+        if isinstance(cmd, AnyEntrypointRequest):
+            # if the current group has a BroadcastHAStates request,
+            # then this entrypoint request should be placed to a next group.
```
mssawant, line 266: So broadcast and entrypoint requests will not be part of the same group. Which seems to be okay but is it possible that one of the consumer thread picked up the last cmd in the current thread and other thread starts executing the next_group as it does not find any more commands in the current group?

mssawant:
I tried to test with 11 s3servers per node in a 3 node setup, but my VM does not take that much load. Need to try it on on some hardware setup.

hax/hax/motr/planner.py
knekrasov: There is no definition of 'fails' in our code.

From WorkPlanner's point of view, the command can be planned for processing ('taken'), it can be in progress now (being processed) and finished.

If an exception happens while processing a command, it should be handled within ConsumerThread (and [it is done](https://github.com/Seagate/cortx-hare/blob/3ff7c323438c67772dd433505986dfe70a44996f/hax/hax/handler.py#L62-L63) [like that](https://github.com/Seagate/cortx-hare/blob/3ff7c323438c67772dd433505986dfe70a44996f/hax/hax/handler.py#L224-L230) already). It is up to the ConsumerThread logic to decide whether the operation should be retried or just skipped. If the operation is retried, we do not leverage any special WorkPlanner mechanisms. All the retries (e.g. because of Consul issues) are looking as one long processing of the same command.

hax/hax/motr/planner.py
knekrasov: Maintaining two different groups with different priorities looks like a complication to me and there is a risk to introduce a bug.

The case you're describing is done in a bit different way. 
1. We don't assume that `Die` commands will be put to WorkPlanner from outside (it is not hardly restricted though)
2. Instead, WorkPlanner will start generating `Die` commands automatically when it is  set to "shutting down" mode (see `shutdown()` method)
3. In "shutting down" mode WorkPlanner will always return `Die` command instance any time `get_next_command()` is invoked (although the backlog doesn't have to contain it). Returned `Die` command instance will always belong to the currently active group (reason: no need to wait, as the whole application is shutting down).
4. In "shutting down" mode WorkPlanner will always return `Die` command instance any time `ensure_allowed(cmd)` is invoked. In this case `ensure_allowed` works as if it "poisons" the given command turning it into `Die` (that is the reason why `ensure_allowed` has signature `BaseCommand -> BaseCommand` instead of `BaseCommand -> None`). Returned command also belongs to the current group.

Long story short, `Die` commands is a special case when groups stop making any sense. Effectively `Die` commands get executed as quick as possible ("on priority").

hax/hax/motr/planner.py
```diff
@@ -0,0 +1,297 @@
...
+    def _should_increase_group(self, cmd: BaseMessage) -> bool:
+        ''' Predicate function. Returns True if and only if cmd command CANNOT
+            be added to group number next_group_id.
+            Assumes that b_lock is acquired already.
+        '''
+        def has(cmd_type: Type[BaseMessage]) -> bool:
+            ''' Checks if the group being currently formed (i.e. next_group)
+                contains a message of the given type.
+            '''
+            return cmd_type in self.state.next_group_commands
```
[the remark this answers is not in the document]
knekrasov: I guess that `Set[Type[BaseCommand]]` is still more efficient than `Dict[[Type[BaseCommand]], int]`. Effectively we don't need counters. We just need to know what types are already there in the next group.

Please note that this structure is needed and used when the command is just being added to the backlog. It is not about execution time. When a command is being added to the queue, its group id is evaluated. So all the commands stored in the backlog have some meaningful group_id assigned.

After that. when ConsumerThreads start requesting the work, those simple group ids just highlight which command can be processed right away and which one must wait until its group becomes active. So we don't even have to maintain the list of types with counters, increase or decrease them.

hax/hax/motr/planner.py
```diff
@@ -0,0 +1,297 @@
...
+            self.state.is_shutdown = True
+            self.b_lock.notifyAll()
+
+    def ensure_allowed(self, command: BaseMessage) -> BaseMessage:
+        ''' The method which must be invoked by every worker thread before
+        starting the command execution. If WorkPlanner allows this command
+        to be executed right now then this method will return early.
+
+        If the execution is not allowed at the moment, the thread will be
+        blocked until the execution is allowed.
```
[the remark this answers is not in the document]
knekrasov: Do you think it is worth repeating [the design document](https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/178421786/22%2BPARHAX%2BParallel%2Bmessage%2Bprocessing%2Bin%2BHaX#Group-forming-rules)?

Ok, I'll try to list it shortly.

hax/hax/motr/planner.py
knekrasov: The thread which picked up a command from `next_group_id` will be allowed to process it if and only if `current_group_id` becomes equal to the command's group id.
Current group id gets changed when all these criteria are true:
1. in backlog there is no command with `group_id == current_group_id` (backlog is used as a source for `get_next_command()`)
2. in `taken_commands` list there is no command `group_id == current_group_id` (which means that there is no ConsumerThread that has already taken a command from the backlog with such an id but has not started processing it yet).
3. in `active_commands` list there is no command `group_id == current_group_id` (this list contains the commands being processed at this moment, so WorkPlanner has already allowed their execution)

hax/hax/motr/planner.py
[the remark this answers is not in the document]
knekrasov: Updated the comments.

hax/hax/hax.py
knekrasov: As far as I remember, the formula usually looks like this:
`Number of threads = Number of Available Cores / (1 - Blocking Coefficient)` ([source](https://superuser.com/questions/664550/optimum-number-of-threads-while-multitasking))

So it really depends how much of time the threads can do something meaningful and how much of time they will be waiting for an IO operation. In other words, we probably need to experiment and propose a rule of a thumb for this setting.

mssawant: In worst case every thread may block (although the best efforts should be to avoid executing a blocking operation), in such cases if a work item is going to block a thread, an additional handler thread must be created which must be destroyed again after completion of the blocking operation via some routines, e.g. block_enter(), block_leave(). But this can complicate the design, thus for now we can come up with a formula to derive number of threads and try to avoid blocking the consumer thread.

knekrasov: Even adding a new thread like you described can decrease the performance (in case when maintaining a thread, context switching and GIL acquiring brings more problems than solves anything).

When threads block on IO the most of time, then obviously number of threads should be `round(N * os.cpu_count())` where N > 1. When the load of Motr-related messages is not that high, then too much of parallelism doesn't make sense (even on a machine with 32 CPU cores). That brings us to `round(M * os.cpu_count())` where 0 < M <= 1.

I think that we need to gather some statistics (like mean or max time that commands spend in the backlog) and warn if the values are suspiciously big (i.e. processing is too slow). That can helps us to held the experiments and thus calibrate the recommended coefficients.

hax/hax/hax.py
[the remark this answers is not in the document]
mssawant: Okay, fair enough.

hax/hax/motr/planner.py
mssawant: Okay

hax/hax/handler.py
knekrasov: @mssawant I've addressed the latest comments and rebased the PR against the latest main branch (there were conflicts, hopefully I resolved them properly).

Please have a look.

hax/hax/motr/planner.py
knekrasov: Addressed.

mssawant (approved):

suvratjoshi (approved):

PR #1614, opened 2021-05-10
EOS-18907: Implement "reset" and "cleanup" phase for Hare mini-provisioner

SwapnilGaonkar7:
'Reset' will delete entries from consul KV which are added by Hare
'Cleanup' will delete config(/var/lib/hare) and logs(/var/log/hare)

### DIFF
--- a/hax/hax/util.py
+++ b/hax/hax/util.py
@@ -36,7 +36,7 @@ from urllib3.exceptions import HTTPError
 from hax.exception import HAConsistencyException, InterruptedException
 from hax.types import (ConfHaProcess, Fid, FsStatsWithTime,
                        ObjT, ServiceHealth, Profile, m0HaProcessEvent,
-                       m0HaProcessType, HaNoteStruct)
+                       m0HaProcessType, KeyDelete, HaNoteStruct)
 
 __all__ = ['ConsulUtil', 'create_process_fid', 'create_service_fid',
            'create_sdev_fid', 'create_drive_fid']
@@ -206,6 +206,26 @@ class ConsulKVBasic:
             return False
         except (ConsulException, HTTPError, RequestException) as e:
             raise HAConsistencyException('Failed to put value to KV') from e
+
+    @repeat_if_fails(max_retries=5)
+    def kv_delete_in_transaction(self, tx_payload: List[KeyDelete]) -> bool:
+        def to_payload(v: KeyDelete) -> Dict[str, Any]:
+            return {'KV': {'Key': v.name, 'Verb':
+                           'delete-tree' if v.recurse else 'delete'}}
+
+        try:
+            self.cns.txn.put([to_payload(i) for i in tx_payload])
+            return True
+        except ClientError:
+            # If a transaction fails, Consul returns HTTP 409 with the
+            # JSON payload describing the reason why the transaction
+            # was rejected.
+            # The library transforms HTTP 409 into generic ClientException.
+            # Unfortunately, we can't easily extract the payload from it.
+            return False
+        except (ConsulException, HTTPError, RequestException) as e:
+            raise HAConsistencyException(f'Failed to delete key(s)'
+                                         f' from KV, error: {e}')
 
 
 class ConsulUtil:
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -36,6 +36,8 @@ from cortx.utils.product_features import unsupported_features
 from hare_mp.cdf import CdfGenerator
 from hare_mp.store import ConfStoreProvider
 from hare_mp.validator import Validator
+from hax.util import ConsulUtil
+from hax.types import KeyDelete
 
 
 def execute(cmd: List[str]) -> str:
@@ -201,6 +203,65 @@ def test(args):
         exit(-1)
 
 
+def reset(args):
+    try:
+        rc = 0
+        util: ConsulUtil = ConsulUtil()
+
+        keys: List[KeyDelete] = [KeyDelete(name='epoch', recurse=False),
+                                 KeyDelete(name='eq-epoch', recurse=False),
+                                 KeyDelete(name='last_fidk', recurse=False),
+                                 KeyDelete(name='leader', recurse=False),
+                                 KeyDelete(name='m0conf/', recurse=True),
+                                 KeyDelete(name='processes/', recurse=True),
+                                 KeyDelete(name='stats/', recurse=True)]
+
+        logging.info('Deleting Hare KV entries (%s)', keys)
+        if not util.kv.kv_delete_in_transaction(keys):
+            logging.error('Error during key delete in transaction')
+            rc = -1
+
+        exit(rc)
+    except Exception as error:
+        logging.error('Error during reset (%s)', error)
+        exit(-1)
+
+
+def cleanup(args):
+    try:
+        rc = -1
+
+        if config_cleanup() == 0 and logs_cleanup() == 0:
+            rc = 0
+
+        exit(rc)
+    except Exception as error:
+        logging.error('Error during cleanup (%s)', error)
+        exit(-1)
+
+
+def logs_cleanup():
+    try:
+        logging.info('Cleaning up hare log directory(/var/log/hare)')
+        os.system('rm -rf /var/log/hare')
+
+        return 0
+    except Exception as error:
+        logging.error('Error during logs cleanup (%s)', error)
+        return -1
+
+
+def config_cleanup():
+    try:
+        logging.info('Cleaning up hare config directory(/var/lib/hare)')
+        os.system('rm -rf /var/lib/hare/*')
+
+        return 0
+    except Exception as error:
+        logging.error('Error during config cleanup (%s)', error)
+        return -1
+
+
 def generate_support_bundle(args):
     try:
         # Default target directory is /tmp/hare
@@ -224,13 +285,11 @@ def checkRpm(rpm_name):
                                 stdin=subprocess.PIPE,
                                 stdout=subprocess.PIPE,
                                 stderr=subprocess.PIPE,
-                                env={},
                                 encoding='utf8')
     rpm_search = subprocess.Popen(["grep", "-q", rpm_name],
                                   stdin=rpm_list.stdout,
                                   stdout=subprocess.PIPE,
                                   stderr=subprocess.PIPE,
-                                  env={},
                                   encoding='utf8')
     out, err = rpm_search.communicate()
     logging.debug("Output: {}".format(out))
@@ -411,13 +470,13 @@ def main():
     add_subcommand(subparser,
                    'reset',
                    help_str='Resets temporary Hare data and configuration',
-                   handler_fn=noop, config_required=False)
+                   handler_fn=reset, config_required=False)
 
     add_subcommand(
         subparser,
         'cleanup',
         help_str='Resets Hare configuration, logs and formats Motr metadata',
-        handler_fn=noop, config_required=False)
+        handler_fn=cleanup, config_required=False)
 
     add_subcommand(subparser,
                    'prepare',
--- a/provisioning/miniprov/mypy.ini
+++ b/provisioning/miniprov/mypy.ini
@@ -1,0 +1,9 @@
+[mypy]
+python_version = 3.6
+check_untyped_defs = True
+
+[mypy-urllib3]
+ignore_missing_imports = True
+
+[mypy-urllib3.exceptions]
+ignore_missing_imports = True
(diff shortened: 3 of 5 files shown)


provisioning/miniprov/hare_mp/main.py
```diff
@@ -201,6 +202,97 @@ def test(args):
...
+    return not os.path.isfile('/var/lib/hare/consul-agents.json')
+
+
+def logs_cleanup(url: str, nodes: List[str]):
+    try:
+        for node in nodes:
+            if is_localhost(node, url):
+                os.system('rm -rf /var/log/hare')
+            else:
+                os.system('ssh ' + node + ' rm -rf /var/log/hare')
```
knekrasov, line 243: In my opinion, it is better to think about logging here.

If, by any reason, this instruction fails, how would we learn what was going to happen and what went wrong? And at which node?

provisioning/miniprov/hare_mp/main.py
```diff
@@ -201,6 +202,97 @@ def test(args):
         exit(-1)
 
 
+def reset(args):
+    try:
+        rc = 0
+        exit(rc)
```
knekrasov, line 208: It's not great that every small function can suddenly perform exit(). Small function does some small job but exti() looks like something existential. Usually it is good to have just one single point in the code that makes a decision whether the application should be finished and what exit code should it issue.

It was fine when the number of such functions was not great, but I don't like that this habit is growing.

But perhaps it should be addressed with a separate PR.
@mssawant What do you think?

provisioning/miniprov/hare_mp/main.py
```diff
@@ -201,6 +202,97 @@ def test(args):
...
+            nodes = get_all_nodes()
+            rc = config_cleanup(url, nodes) & logs_cleanup(url, nodes)
+
+        exit(rc)
+    except Exception as error:
+        logging.error('Error during cleanup (%s)', error)
+        exit(-1)
+
+
+def is_cleanup_done() -> bool:
```
knekrasov, line 233: Suggestion: `is_cleanup_needed()`. We don't know whether it is DONE. But we know whether it makes sense to do it.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -201,6 +202,97 @@ def test(args):
...
+        return -1
+
+
+def config_cleanup(url: str, nodes: List[str]):
+    try:
+        for node in nodes:
+            if is_localhost(node, url):
+                os.system('rm -rf /var/lib/hare/*')
+            else:
+                os.system('ssh ' + node + ' rm -rf /var/lib/hare/*')
```
knekrasov, line 257: `/var/lib/hare/*` is a glob. Are you sure that it will be expanded at _remote_ machine? I'm not that sure. Usually bash linters recommend to wrap such expressions into quotes to prevent globbing BEFORE the command is transmitted to the proper machine.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -201,6 +202,97 @@ def test(args):
...
+        return None
+    with open(path) as f:
+        return f.readline().strip()
+
+
+def is_localhost(node: str, url: str) -> bool:
+    conf = ConfStoreProvider(url)
+    hostname = conf.get_hostname()
+
+    localhost_list = ['localhost', '127.0.0.1', hostname, minion_id(),
```
knekrasov, line 277: Some time back I was fixing redundant ssh'ing into localhost: #1557 Don't we need to consider node-name file here as well?

provisioning/miniprov/hare_mp/main.py
knekrasov, line 243: Ssh'ing is done in copy-paste manner. It may become a problem if, say, somebody will ask to change the user that is used for ssh.

Is it possible that a specific user may be required in the future? Or some specific options to `ssh` command will be wanted (like `-vv`)?

provisioning/miniprov/hare_mp/main.py
```diff
@@ -201,6 +202,97 @@ def test(args):
...
+
+def minion_id():
+    path = '/etc/salt/minion_id'
+    if not os.path.isfile(path):
+        return None
+    with open(path) as f:
+        return f.readline().strip()
+
+
+def is_localhost(node: str, url: str) -> bool:
```
mssawant, line 273: No need for this method.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -201,6 +202,97 @@ def test(args):
...
+    try:
+        rc = 0
+        url = args.config[0]
+        validator = Validator(ConfStoreProvider(url))
+        if validator.is_first_node_in_cluster():
+            if is_cleanup_done():
+                logging.info('No data to cleanup')
+                exit(0)
+
+            nodes = get_all_nodes()
```
mssawant, line 224: No need for this.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -201,6 +202,97 @@ def test(args):
...
+    conf = ConfStoreProvider(url)
+    hostname = conf.get_hostname()
+
+    localhost_list = ['localhost', '127.0.0.1', hostname, minion_id(),
+                      socket.gethostname(), socket.getfqdn()]
+
+    return node in localhost_list
+
+
+def get_all_nodes() -> List[str]:
```
mssawant, line 283: No need for this.

mssawant (changes requested):

SwapnilGaonkar7:
@mssawant @knekrasov I have addressed all the review comments which were still valid and also added code for 'reset'
Please have a look

provisioning/miniprov/hare_mp/main.py
```diff
@@ -201,6 +202,63 @@ def test(args):
...
+
+        exit(rc)
+    except Exception as error:
+        logging.error('Error during reset (%s)', error)
+        exit(-1)
+
+
+def cleanup(args):
+    try:
+        rc = config_cleanup() & logs_cleanup()
```
knekrasov, line 232: Bitwise 'and' operator doesn't have a short circuit. So even in case of an error the latter function will be executed. Is that a desired behavior?

provisioning/miniprov/hare_mp/main.py
```diff
@@ -201,6 +202,63 @@ def test(args):
...
+                'eq-epoch': False,
+                'last_fidk': False,
+                'leader': False,
+                'm0conf/': True,
+                'processes/': True,
+                'stats/': True}
+
+        logging.info('Deleting Hare KV entries from consul KV store')
+        for key in keys:
+            if not consul.kv.delete(key, keys[key]):
```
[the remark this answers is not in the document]

SwapnilGaonkar7:
@knekrasov @mssawant I have addressed review comments. Please check

hax/hax/util.py
```diff
@@ -207,6 +207,28 @@ def to_payload(v: TxPutKV) -> Dict[str, Any]:
         except (ConsulException, HTTPError, RequestException) as e:
             raise HAConsistencyException('Failed to put value to KV') from e
 
+    @repeat_if_fails(max_retries=5)
+    def kv_delete_in_transaction(self, tx_payload: Dict[str, str]) -> bool:
```
knekrasov, line 211: Can we use an explicit type for `tx_payload` instead of raw `Dict[str, str]`?

The problem is that it is too easy to provide a dict with incorrect structure that your function will fail to process. If you use `@dataclass` or `NamedTuple`, the requirements to such a dict become explicit and can be found during static analysis.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -201,6 +202,62 @@ def test(args):
...
+        keys = {'epoch': False,
+                'eq-epoch': False,
+                'last_fidk': False,
+                'leader': False,
+                'm0conf/': True,
+                'processes/': True,
+                'stats/': True}
+
+        logging.info('Deleting Hare KV entries (%s)', keys)
+        if not util.kv.kv_delete_in_transaction(keys):
```
knekrasov, line 219: The function has signature `kv_delete_in_transaction(self, tx_payload: Dict[str, str]) -> bool` while you feed it with `keys` of type `Dict[str, bool]`.

Did you really check this code with mypy? 😉

provisioning/miniprov/mypy.ini
```diff
@@ -0,0 +1,8 @@
+[mypy]
+python_version = 3.6
```
knekrasov, line 2: Please look into https://github.com/Seagate/cortx-hare/blob/main/hax/mypy.ini

There is an important directive there:
```
check_untyped_defs = True
```
that appeared when it turned out that mypy (by default) doesn't try to validate the source code which IS NOT type-hinted. We missed some issues because of that. Please see [this doc](https://mypy.readthedocs.io/en/stable/common_issues.html#no-errors-reported-for-obviously-wrong-code) for more details.

knekrasov (changes requested):

provisioning/miniprov/hare_mp/main.py
```diff
@@ -201,6 +202,62 @@ def test(args):
         exit(-1)
 
 
+def reset(args):
```
suvratjoshi, line 205: I think we need to cleanup installation data as well. We can remove installed things in reset `reset` option.
1. `/opt/seagate/cortx/hare` path.
2. All installations that hare component does.

provisioning/miniprov/mypy.ini
SwapnilGaonkar7: Done

hax/hax/util.py
SwapnilGaonkar7: Done

provisioning/miniprov/hare_mp/main.py
SwapnilGaonkar7: Done

SwapnilGaonkar7:
@knekrasov I have addressed your review comments please have a look

hax/hax/util.py
```diff
@@ -207,6 +207,28 @@ def to_payload(v: TxPutKV) -> Dict[str, Any]:
         except (ConsulException, HTTPError, RequestException) as e:
             raise HAConsistencyException('Failed to put value to KV') from e
 
+    @repeat_if_fails(max_retries=5)
+    def kv_delete_in_transaction(self, tx_payload: List[KeyDelete]) -> bool:
+        def to_payload(v: KeyDelete) -> Dict[str, Any]:
+            if v.recurse is True:
```
knekrasov, line 213: `if v.recurse:` is sufficient.

`is` operator compares the identity (memory address) of two objects. Perhaps `True` and `False` objects are kind of singletons within Python interpreter which makes your code working, but in general case this habit is a bit dangerous.

And I guess these four lines can be replaced by ternary operator like this (please double check, I didn't try it in REPL):
```suggestion
            return {'KV': {'Key': v.name, 'Verb': 'delete-tree' if v.recurse else 'delete'}}
```

provisioning/miniprov/hare_mp/main.py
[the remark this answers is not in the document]
knekrasov: @suvratjoshi That's a good point but probably that's not in scope of this PR. Hopefully we'll come up to an idea of KV schema (see [EOS-17956](https://jts.seagate.com/browse/EOS-17956)) and thus have an idea how to express it in the code.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -224,13 +285,13 @@ def checkRpm(rpm_name):
...
                                 stderr=subprocess.PIPE,
-                                env={},
+                                env=None,
                                 encoding='utf8')
     rpm_search = subprocess.Popen(["grep", "-q", rpm_name],
                                   stdin=rpm_list.stdout,
                                   stdout=subprocess.PIPE,
                                   stderr=subprocess.PIPE,
-                                  env={},
+                                  env=None,
```
knekrasov, line 294: What is the meaningful difference between `env=None` and `env={}`?

If `env=None` makes subprocess to inherit the environment of the current process, then it is worth removing `env=` kwarg at all.

SwapnilGaonkar7: Done, default value of env in None anyway. In that case it will inherit parent

hax/hax/util.py
SwapnilGaonkar7: DOne

provisioning/miniprov/hare_mp/main.py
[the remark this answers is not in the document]
suvratjoshi: ok.
Approving PR.

suvratjoshi (approved):

knekrasov (approved):

PR #1616, opened 2021-05-10
EOS-20534: hax message can be delivered before waiting for delivery

mssawant:
It is possible that the hax message can be delivered even before
intiating wait on the delivery for a given ha message. In this
particular case, it is observed that the message is already delivered
before hax starting to wait on the delivery. Thus, as delivery callback
is already received, `notify waiting clients dict_items([])` and
hax starts waiting after the notification,
`waiting for broadcast of [MessageId(0x7f21a8008f00, 1)]`, the notification
is missed and hax continues to wait, thus timing out the entrypoint reply.

Solution:
Add the delivered `MessageID` to `herald.recently_delivered` list unconditionally
so that the delivered message is already present in the `recently delivered` messages
list when hax start to wait for then same.

(cherry picked from commit 6e22c54cca87f8338134d8ca34865f1fdf11a844)

### DIFF
--- a/hax/hax/motr/delivery.py
+++ b/hax/hax/motr/delivery.py
@@ -110,11 +110,10 @@ class DeliveryHerald:
     def notify_delivered(self, message_id: MessageId):
         # [KN] This function is expected to be called from Motr.
         with self.lock:
-            LOG.log(TRACE, 'notify waiting clients %s',
-                    self.waiting_clients.items())
+            LOG.debug('received msg id %s, notify waiting clients %s',
+                      message_id, self.waiting_clients.items())
             for promise, client in self.waiting_clients.items():
-                LOG.log(TRACE, 'received msg id %s, waiting promise %s',
-                        message_id, promise)
+                LOG.debug('waiting promise %s', promise)
                 if message_id in promise:
                     LOG.log(TRACE, 'Found a waiting client for %s: %s',
                             message_id, promise)
@@ -124,3 +123,24 @@ class DeliveryHerald:
                     with client:
                         client.notify()
                     return
+            # If notify_delivered() was invoked before wait_for_all(), i.e.
+            # if the ha message is already delivered before the sender starts
+            # waiting on the same, append the delivered message to the list of
+            # recently_delivered, so that the delivery is found before
+            # wait_for_{all, any}() starts the conditional wait.
+            self.recently_delivered[HaLinkMessagePromise(
+                [message_id])] = [message_id]
+
+    # This function must be invoked with the self.lock held.
+    def check_if_delivered_locked(
+            self,
+            promise: HaLinkMessagePromise) -> HaLinkMessagePromise:
+        if not self.lock.locked():
+            raise RuntimeError('DeliveryHerald.lock not acquired')
+        if promise in self.recently_delivered:
+            confirmed_msgs = self.recently_delivered.pop(promise)
+            LOG.debug('Thread unblocked - %s just received',
+                      confirmed_msgs)
+            del self.waiting_clients[promise]
+            promise.exclude_ids(confirmed_msgs)
+        return promise


hax/hax/motr/delivery.py
```diff
@@ -124,3 +123,24 @@ def notify_delivered(self, message_id: MessageId):
                     with client:
                         client.notify()
                     return
+            # If notify_delivered() was invoked before wait_for_all(), i.e.
+            # if the ha message is already delivered before the sender starts
+            # waiting on the same, append the delivered message to the list of
+            # recently_delivered, so that the delivery is found before
+            # wait_for_{all, any}() starts the conditional wait.
+            self.recently_delivered[HaLinkMessagePromise(
```
knekrasov, line 131: This change will most probably cause `recently_delivered` dictionary to grow unconditionally.

The fact is that if a message is issued from our `hax.c` and nobody invokes `wait_for_all()`/`wait_for_any()`, the corresponding promise will never be removed from this dictionary.

I understand that the problem you're solving is real and we need this kind of fix. But we also need a mechanism that will remove these orphan promises from this dict eventually.

Options:
1. Try [weak references](https://docs.python.org/3/library/weakref.html#module-weakref) to get promises garbage-collected eventually.
   - Cons: this removal is a kind of uncontrolled. We don't know exactly when a key or value will be removed. So there is even a chance that a "good" (not orphaned, i.e. the one being awaited by somebody) will be removed too early.
2. Manual TTL mechanism with a background thread that traverses the keys and checks if they seem to be too old.

knekrasov: Update: the corresponding change is now part of this PR. @mssawant please review once you have time.

knekrasov:
@mssawant Since I added one commit to this PR, I think I have no moral right to hit "approve" button 😄 
How can we merge this PR? Who can review it?

hax/hax/motr/delivery.py
```diff
@@ -75,6 +84,7 @@ def wait_for_any(self,
         """
         condition = Condition()
         with self.lock:
+            self.groom_unsorted(promise)
```
mssawant, line 87: Okay, we are grooming the ones on which the sender does not wait on, but what about already delivered? we are still waiting for delivery but a notification will never come right? don't we need to check in recently_delivered before waiting on the condition post grooming?

knekrasov: Not sure that I understood you right.

`wait_for_*` functions block on Condition with a timeout. If the delivery will never happen, the waiting thread is unblocked by the timeout, and then [`_verify_delivered()`](https://github.com/Seagate/cortx-hare/blob/2e0ae5a413b6e1109102f900fea132ec274cf1c1/hax/hax/motr/delivery.py#L54) is invoked. Within that function `self.waiting_clients[promise]` is removed, so we don't have a leak here.

Do I miss anything?

knekrasov: And one more thing. Please note that `groom_unsorted()` moves the entries that relate to this particular promise to `recently_delivered`:
https://github.com/Seagate/cortx-hare/blob/2e0ae5a413b6e1109102f900fea132ec274cf1c1/hax/hax/motr/delivery.py#L129-L138

mssawant: I mean for the cases where the message is already delivered before a sender invokes wait_for_any or wait_for_all, in such a case we are not checking if the message is already delivered, don't we need to check the `unsorted_deliveries` before going into wait?

knekrasov: Good point. I'll need to fix that.

hax/hax/motr/delivery.py
```diff
@@ -124,3 +123,24 @@ def notify_delivered(self, message_id: MessageId):
...
+                [message_id])] = [message_id]
+
+    # This function must be invoked with the self.lock held.
+    def check_if_delivered_locked(
+            self,
+            promise: HaLinkMessagePromise) -> HaLinkMessagePromise:
+        if not self.lock.locked():
+            raise RuntimeError('DeliveryHerald.lock not acquired')
+        if promise in self.recently_delivered:
+            confirmed_msgs = self.recently_delivered.pop(promise)
```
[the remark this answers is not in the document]
knekrasov: @suvratjoshi `recently_delivered` is a dictionary:
```python
        self.recently_delivered: Dict[HaLinkMessagePromise,
                                      List[MessageId]] = {}
```
`pop()` takes a dictionary key as an argument. Key is unique by definition. So I don't think that your concern is valid.

knekrasov:
@mssawant I have moved my improvements to a separate PR #1627

Let's merge this one and get main branch to be coherent with cortx-1.0

hax/hax/motr/delivery.py
knekrasov: Fixed in #1627. Not relevant for this PR anymore.

knekrasov (approved):

PR #1623, opened 2021-05-26
EOS-20754: scripts under libexec/ are not runnable when invoked by short name

knekrasov:
All the scripts under libexec/ can be invoked by short name only in
the context of hctl script. In non-hctl context, the scripts should use
absolute path instead:

- elect-rc-leader invokes node-name and it doesn't work in context of
hctl.
- get-service-health also invokes node-name; it can be invoked directly
by Consul so node-name should be addressed by absolute path.
- utils/utils.py contains redundant invocation of node-name (by short
name). That function is removed.

### DIFF
--- a/utils/elect-rc-leader
+++ b/utils/elect-rc-leader
@@ -47,6 +47,14 @@ get_leader_node() {
         jq -r '.[] | .Node'
 }
 
+get_node_name() {
+    # hare/libexec is added to PATH env variable only by hctl 'grounding'
+    # script.
+    # In non-hctl context, the scripts should use absolute path instead
+    # when accessing scripts under hare/libexec/.
+    /opt/seagate/cortx/hare/libexec/node-name
+}
+
 cleanup() {
     # Nicely stop currently running RC handler (if any):
     pkill -9 -f 'consul watch.*prefix eq' || true
@@ -57,7 +65,7 @@ cleanup() {
 # If the session is already set - the leader is elected, so
 # there is nothing to do (except cleanup after ourself).
 get_session_id | grep -q ^- || {
-    if [[ $(get_leader_node) != $(node-name) ]]; then
+    if [[ $(get_leader_node) != $(get_node_name) ]]; then
         cleanup
     fi
     exit 0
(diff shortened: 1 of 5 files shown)


mssawant:
@knekrasov , the patch looks good, also I was thinking if some how we can suspend consul watchers post shutdown and restart them when hax service starts. I think we can do this from hax systemd unit as a pre and post step. I have seen that when Hare + Motr cluster is down, hare Consul watchers keep running and keep complaining about service critical which eats into the CPU and pacemaker commands also get delayed.

mssawant (approved):

utils/elect-rc-leader
```diff
@@ -47,6 +47,14 @@ get_leader_node() {
         jq -r '.[] | .Node'
 }
 
+get_node_name() {
+    # hare/libexec is added to PATH env variable only by hctl 'grounding'
+    # script.
+    # In non-hctl context, the scripts should use absolute path instead
+    # when accessing scripts under hare/libexec/.
+    /opt/seagate/cortx/hare/libexec/node-name
```
suvratjoshi, line 55: Instead of writing full path every time, can we have some macro name for this path, like we have
CORTX_HA_BIN
HARE_BASE_DIR
etc.

utils/elect-rc-leader
```diff
@@ -47,6 +47,14 @@ get_leader_node() {
         jq -r '.[] | .Node'
 }
 
+get_node_name() {
```
suvratjoshi, line 50: Same function name with same purpose in two files.
Can we have only one at one place?

utils/elect-rc-leader
knekrasov: @suvratjoshi where would you propose to put those macros to?

suvratjoshi: I am not sure on that how we follow this, so not sure where to put it.
Do we have a common file where we can have such macros and can be accessible globally? (Like what we do in a C language code)

suvratjoshi:
As my comments are some code formatting related and minor, and not critical. Testing will also be needed on that.
We can take up those changes later, so I am approving as this patch is urgent.

suvratjoshi (approved):

utils/elect-rc-leader
mssawant: @suvratjoshi , either its a function or a macro. I think function is good.

PR #1632, opened 2021-06-07
EOS-21072: Update s3server port to 28071

SwapnilGaonkar7:

### DIFF
--- a/utils/update-consul-conf
+++ b/utils/update-consul-conf
@@ -227,7 +227,7 @@ for id in $IOS_IDs; do
     append_ios_svc $id
 done
 
-s3port=28081
+s3port=28071
 for id in $S3_IDs; do
     append_s3_svc $id $s3port
     ((s3port++))


utils/update-consul-conf
```diff
@@ -227,7 +227,7 @@ for id in $IOS_IDs; do
     append_ios_svc $id
 done
 
-s3port=28081
+s3port=28071
```
mssawant, line 230: Going further, I was actually thinking instead of hardcoding the port here, can we take it as an argument? was thinking of having a hare configuration file (/etc/sysconfig/hare) in which we can specify all the Hare specific environmental variables, e.g. s3port. Or I think we should read the s3port somewhere from a s3server configuration file. Because everytime the port number changes we will have to make a code change and if we can achieve this without code change then it is better.
@knekrasov , what do you think?

knekrasov: Although I'm not that sure that S3 port deserves an externalized configuration parameter, in general it would be nice to have some Hare config file with various parameters such as log level, log format, number of worker threads (#1612) etc.

So it is nice to have but it deserves a separate PR.

knekrasov (approved):

mssawant:
@SwapnilGaonkar7 , kindly update the PR title summary as,
```
EOS-21072: Update s3server port to 28071
```

utils/update-consul-conf
mssawant: Yeah, separate PR sounds good. Just that we need not have to change the code every time some port changes. These things need to be configurable. Let's create a ticket to identify all the configurable Hare parameters and make them configurable.
@SwapnilGaonkar7 Can you please create a ticket for the same?

mssawant (approved):

utils/update-consul-conf
SwapnilGaonkar7: Done, https://jts.seagate.com/browse/EOS-21132

PR #1634, opened 2021-06-08
EOS-21119 Hare mini-provisioner reset phase does not shutdown the cluster

SwapnilGaonkar7:
In 'reset' phase of hare we are cleaning up the data generated. We are deleting consul KV entries generated by Hare.
But before that we need to stop Hare services as they might keep generating more data.

We are calling 'hctl shutdown' to stop services if cluster is running

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -213,6 +213,10 @@ def reset(args):
         rc = 0
         util: ConsulUtil = ConsulUtil()
 
+        if is_cluster_running():
+            logging.info('Cluster is running, shutting down')
+            shutdown_cluster()
+
         keys: List[KeyDelete] = [KeyDelete(name='epoch', recurse=False),
                                  KeyDelete(name='eq-epoch', recurse=False),
                                  KeyDelete(name='last_fidk', recurse=False),


knekrasov (approved):

mssawant:
@SwapnilGaonkar7, Summary needs to state the problem, kindly update the summary as,
```
EOS-18907: Hare mini-provisioner reset phase does not shutdown the cluster
```

provisioning/miniprov/hare_mp/main.py
```diff
@@ -213,6 +213,10 @@ def reset(args):
         rc = 0
         util: ConsulUtil = ConsulUtil()
 
+        if is_cluster_running():
+            logging.info('Cluster is running, calling shutdown')
```
mssawant, line 217: s/Cluster is running, calling shutdown/Cluster is running, shutting down

mssawant (approved):
Minor comments.

provisioning/miniprov/hare_mp/main.py

provisioning/miniprov/hare_mp/main.py
```diff
@@ -213,6 +213,10 @@ def reset(args):
         rc = 0
         util: ConsulUtil = ConsulUtil()
 
+        if is_cluster_running():
+            logging.info('Cluster is running, shutting down')
+            shutdown_cluster()
```
suvratjoshi, line 218: def shutdown_cluster(skip_consul_stop: bool):
Do we need to check the default value for `skip_consul_stop`? and set to False?

suvratjoshi:
Procedural : 
JIRA https://jts.seagate.com/browse/EOS-18907 is closed already but how this PR is being raised under that JIRA?
Sorry, as I might have missed it, is this a new requirement that we understood recently, so creating second PR?
It is not in current sprint I think. Do we need separate jira?
cc: @vaibhavparatwar

SwapnilGaonkar7:
@suvratjoshi There is separate jira EOS-21119 for this.

provisioning/miniprov/hare_mp/main.py
SwapnilGaonkar7: We always skip consul stop in mini provisioning

PR #1638, opened 2021-06-09
Problem: newer mypy fails on Hare sources

knekrasov:
Problem reason:
1. Hare doesn't specify explicit mypy version in its requirements
2. So once newer mypy version is released, it is applied automatically
if .py3venv/ folder doesn't exist (yet)
3. It turns out that mypy 0.901 stops bringing type stubs for libraries
by default. Corresponding type stubs should be installed explicitly.

Solution: add missing type-* libraries to the build requirements.

Note: these depedendencies are build-time only. Runtime dependencies are
not affected.

### DIFF
--- a/hax/requirements.txt
+++ b/hax/requirements.txt
@@ -1,4 +1,11 @@
 #:build-requirements:
+types-PyYAML
+types-chardet
+types-dataclasses
+types-requests
+types-simplejson
+types-pkg_resources
+types-toml
 flake8
 mypy
 pkgconfig


Shreya-18 (approved):

s-arya (approved):

suvratjoshi:
We dont need these types in other requirement files like cfgen or miniprovisioner, right @knekrasov ?

knekrasov:
@suvratjoshi 
> We dont need these types in other requirement files like cfgen or miniprovisioner, right @knekrasov ?

Our Makefile a kind of melds all requirements.txt altogether when preparing virtualenv:
```
$(PY_VENV_DIR): $(patsubst %,%/requirements.txt,cfgen hax provisioning/miniprov)
	@$(call _info,Initializing virtual env in $(PY_VENV_DIR))
	@$(PYTHON) -m venv --system-site-packages $@
	@$(call _info,Installing pip modules in virtual env)
	@for f in $^; do \
	     $(call _log,processing $$f); \
	     $(PIP) install --ignore-installed -r $$f; \
	 done
```
That's why it is sufficient to specify them just once.

suvratjoshi (approved):

PR #1641, opened 2021-06-09
EOS-21102: Singlenode deployment will never succeed

knekrasov:
JIRA issue: [EOS-21102](https://jts.seagate.com/browse/EOS-21102)

Problem: when hare-bootstrap script awaits for m0d processes to start,
it looks into Consul KV. The corresponding key looks as follows: `$HOSTNAME/processes/$fid`

In singlenode deployments $HOSTNAME is resolved to FQDN while in Consul
KV the corresponding node is known as 'localhost'.

As long as utils/get-process-state queries non-existent key, the output
is empty string (instead of desired 'M0_CONF_HA_PROCESS_STARTED').

Solution: use node-name script instead of $HOSTNAME variable.

### DIFF
--- a/utils/get-process-state
+++ b/utils/get-process-state
@@ -50,4 +50,8 @@ case $1 in
 esac
 
 fid=$1
-consul kv get $HOSTNAME/processes/$fid 2> /dev/null | jq -r '.state'
+get_node_name() {
+    /opt/seagate/cortx/hare/libexec/node-name
+}
+
+consul kv get $(get_node_name)/processes/$fid 2> /dev/null | jq -r '.state'


utils/get-process-state
```diff
@@ -50,4 +50,8 @@ case $1 in
 esac
 
 fid=$1
-consul kv get $HOSTNAME/processes/$fid 2> /dev/null | jq -r '.state'
+get_node_name() {
```
suvratjoshi, line 53: As now we are using this function at 2-3 places in the code, can we have this only in one file, say hare/utils/util, or any better place and import it, and remove (shell script) occurrences of the function?

knekrasov: I don't think it makes much sense given that the file `node-name` can be considered as a function as well.

suvratjoshi: Ok.
But I was just thinking from the point to reduce the code and no repetition if it is easily possible to do so.
So we dont have multiple definitions of function `get_node_name()`.

mssawant: So presently we have `get_node_name()` defined in atleast 3 files,
```
elect-rc-leader:50:get_node_name() {
get-service-health:24:get_node_name() {
utils.py:148:def get_node_name() -> str:
```
I think the problem is `get-process-state` is a shell script but still may be we can use the one from `get-service-health`?
Eventually it would help to port all the shell scripts to python scripts.
But let's not block on this, if we can use `get-service-health::get_node_name()` easily then fine else let's do all the de-duplication in a separate PR. But let's create a ticket for the same.

mssawant (approved):

suvratjoshi (approved):

PR #1647, opened 2021-06-11
EOS-17954 : Implement "hare_setup upgrade" CLI

suvratjoshi:
Implement hare_setup upgrade command.
Currently it is set as no-operation as nothing is expected to be done.
Basic API provision is done.

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -529,6 +529,11 @@ def main():
                    help_str='Validates configuration pre-requisites',
                    handler_fn=noop)
 
+    add_subcommand(subparser,
+                   'upgrade',
+                   help_str='Performs the Hare rpm upgrade tasks',
+                   handler_fn=noop)
+
     setup_logging()
 
     parsed = p.parse_args(sys.argv[1:])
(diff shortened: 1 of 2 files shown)


provisioning/miniprov/hare_mp/main.py
```diff
@@ -529,6 +529,11 @@ def main():
                    help_str='Validates configuration pre-requisites',
                    handler_fn=noop)
 
+    add_subcommand(subparser,
+                   'upgrade',
+                   help_str='Upgrades the Hare RPM (Currently unavailable)',
```
knekrasov, line 534: Why "Currently unavailable"? When will it become available?

RPM is upgradable per se. At this point there is nothing that we need to do in scope of this call (and I don't think that it will change in the near future). So this message about unavailability is misleading. Let's remove it.

SwapnilGaonkar7, line 534: `help_str='Upgrades the Hare RPM....'`
I don't think we are upgrading RPM by using this interface. Are we?

mssawant: Right, let's reword it, how about just `Performs tasks to apply changes from the upgraded Hare rpm.`?

suvratjoshi: Done!

suvratjoshi: Done!

mssawant (approved):

knekrasov (approved):

PR #1648, opened 2021-06-16
Problem: gcc is not in BuildRequires of Specfile

knekrasov:
Hare does need gcc to link with Motr C API. By some reason our Specfile
didn't have explicit requirement on gcc compiler to be installed.

Solution: add gcc to BuildRequires list.

### DIFF
--- a/hare.spec
+++ b/hare.spec
@@ -37,6 +37,7 @@ Group: System Environment/Daemons
 Source: %{name}-%{h_version}.tar.gz
 
 BuildRequires: binutils-devel
+BuildRequires: gcc >= 4.8.5
 BuildRequires: git
 BuildRequires: cortx-motr
 BuildRequires: cortx-motr-devel


suvratjoshi:
Interesting!
How this issue was caught?
Is there a JIRA linked?

Question : 
Our software is going to run on Linux only, CentOs to be specific for now. That comes with default gcc, is what I understand.
So generally this should never give an issue, but yes if we need it, we must have this listed for sure.

Similarly, can there be a possibility of a package missing from our requirements/spec files, that probably is pre-installed in the OS?

knekrasov:
@suvratjoshi There is no specific JIRA issue, the problem was reproduced when I tried to run pre-merge test on my custom (freshly created) SSC VM. I don't know why but gcc wasn't there.

suvratjoshi (approved):

vaibhavparatwar:
@mssawant @suvratjoshi do we need this PR now? if yes, we may need to rebase with latest main, if no, can we clsoe it?

suvratjoshi:
I think we need to have it.
Though we are not getting into this issue but in some rare situation it is possible.
It is a one liner change so I think we can merge this by rebasing current one with main if easily possible or else we can create a new PR and close this one,

cortx-admin:

vaibhavparatwar:
@mssawant  is this PR needed? for this we need to connect to RE team and raise PR in RE repo as this is third party package.

PR #1655, opened 2021-06-18
EOS-21355: newer mypy version fails at Hare sources (cortx-1.0)

knekrasov:
JIRA ticket: [EOS-21355](https://jts.seagate.com/browse/EOS-21355)

In Hare sources there is no specific requirement on mypy version. It
seems like the newer version 0.901 stops bringing type information (so
called type stubs) for standard libraries; those type stubs should be
installed separately.

Solution: stick mypy to version 0.812 in cortx-1.0 to avoid more
surprises in a release branch.

### DIFF
--- a/hax/requirements.txt
+++ b/hax/requirements.txt
@@ -1,6 +1,6 @@
 #:build-requirements:
-flake8
-mypy
+flake8==3.9.2
+mypy==0.812
 pkgconfig
 setuptools
 wheel
--- a/hax/setup.py
+++ b/hax/setup.py
@@ -124,7 +124,7 @@ setup(
     name='hax',
     version=get_hax_version(),
     packages=find_packages(),
-    setup_requires=['flake8', 'mypy', 'pkgconfig'],
+    setup_requires=['flake8==3.9.2', 'mypy==0.812', 'pkgconfig'],
     install_requires=[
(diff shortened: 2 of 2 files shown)


suvratjoshi:
If I am not wrong, this is the same/similar issue that we had fixed for main, recently : 
https://github.com/Seagate/cortx-hare/commit/f708ba0bd54fe622362a2a66db077bbfe9084829
Is that correct?

If this understanding is correct, then I think we need to have the same solution for main, stable and cortx-1.0.
If this solution of sticking to particular version of mypy is better than listing all that we need in requirements.txt (as in main and stable), then we need to replace the solution in main and stable with this solution?

knekrasov:
@suvratjoshi yes, the problem is the same, but the solution is different.

Per my understanding, `cortx-1.0` is a release branch. Ideally, all dependencies should be locked to specific version to guarantee that we deliver only that that has been tested.

Main and stable branches are different. There is an active development there and I don't see a reason why should freeze the versions there (if we don't lock the versions, the product will come up to release with the freshest versions that would be supported presumably longer period of time).

But I agree that the release process should be improved to cover situations like this.

cc @NeeravChoudhari, @vaibhavparatwar, @mssawant

mssawant (approved):

suvratjoshi (approved):

PR #1657, opened 2021-06-18
EOS-19917 CORTX-HARE: Implement test interface for Hare

SwapnilGaonkar7:

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -39,11 +39,20 @@ from hare_mp.validator import Validator
 from hax.util import ConsulKVBasic, ConsulUtil, repeat_if_fails
 from hax.types import KeyDelete
 from time import sleep
+from enum import Enum
 
 # Logger details
 LOG_DIR = "/var/log/seagate/hare/"
 LOG_FILE = "/var/log/seagate/hare/setup.log"
 LOG_FILE_SIZE = 5 * 1024 * 1024
+
+
+class Plan(Enum):
+    Sanity = 'sanity'
+    Regression = 'regression'
+    Full = 'full'
+    Performance = 'performance'
+    Scalability = 'scalability'
 
 
 def execute(cmd: List[str]) -> str:
@@ -230,6 +239,29 @@ def test(args):
     except Exception as error:
         logging.error('Error while checking cluster status (%s)', error)
         shutdown_cluster()
+        exit(-1)
+
+
+def test_IVT(args):
+    try:
+        rc = 0
+        path_to_cdf = args.file[0]
+
+        logging.info('Running test plan: ' + str(args.plan[0].value))
+        # TODO We need to handle plan type and execute test cases accordingly
+        if not is_cluster_running():
+            logging.error('Cluster is not running. Cluster must be running '
+                          'for executing tests')
+            exit(-1)
+        cluster_status = check_cluster_status(path_to_cdf)
+        if cluster_status:
+            logging.error('Cluster status reports failure')
+            rc = -1
+
+        logging.info('Tests executed successfully')
+        exit(rc)
+    except Exception as error:
+        logging.error('Error while running Hare tests (%s)', error)
         exit(-1)
 
 
@@ -488,6 +520,26 @@ def add_file_argument(parser):
     return parser
 
 
+def add_plan_argument(parser):
+    parser.add_argument('--plan',
+                        help='Testing plan to be executed. Supported '
+                        'values:' + str([e.value for e in Plan]),
+                        required=True,
+                        nargs=1,
+                        type=Plan,
+                        action='store')
+    return parser
+
+
+def add_param_argument(parser):
+    parser.add_argument('--param',
+                        help='Test input URL.',
+                        nargs=1,
+                        type=str,
+                        action='store')
+    return parser
+
+
 def main():
     p = argparse.ArgumentParser(description='Configure hare settings')
     subparser = p.add_subparsers()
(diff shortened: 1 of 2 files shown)


provisioning/setup.yaml
```diff
@@ -10,7 +10,7 @@ hare:
     args: --config $URL --file '/var/lib/hare/cluster.yaml'
   test:
     cmd: /opt/seagate/cortx/hare/bin/hare_setup test
-    args: --config $URL
+    args: --config $URL --plan $TYPE --param $PARAM_URL
```
nitin-seagate, line 13: We never proposed `--param $PARAM_URL`
Please provide context for this usage?

nitin-seagate (changes requested):

provisioning/setup.yaml
SwapnilGaonkar7: @nitin-seagate [This](https://seagate-systems.atlassian.net/wiki/spaces/PRIVATECOR/pages/221642873/CORTX+Factory+Manufacturing+Process) page specify following command. It is also being discussed in integration meeting.
`cortx_cli test run --plan {sanity|regression|full|performance|scalability} --param <TEST_INPUT_URL>`

knekrasov: @SwapnilGaonkar7 if Nitin has concerns, maybe it is too risky to merge this feature now? Won't this change break deployment (say, if setup.yaml parser fails..)

SwapnilGaonkar7: @knekrasov  Yes, we are not going to merge this until @nitin-seagate confirms.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -232,6 +240,28 @@ def test(args):
...
 
+def test_IVT(args):
+    try:
+        rc = 0
+        path_to_cdf = args.file[0]
+
+        logging.info('Running test plan: ' + str(args.plan[0].value))
+        # TODO We need to handle plan type and execute test cases accordingly
+        if not is_cluster_running():
+            logging.error('Cluster is not running. Cluster must be running '
```
knekrasov, line 251: This is not so obvious to me. Is it required by any specification?

Or, speaking wider, is there any requirement in which state the system will be when `test` is invoked? And in which state the system must be left when `test` stage finishes?

SwapnilGaonkar7: @knekrasov It is expected that cluster will be already running when 'test' gets called.
We should not start/stop cluster during test stage

mssawant: So now `test_IVT` command is replacing `test` command. I believe `test()` needs to be invoked for plan `Sanity`, in that case please add that call as well.
You can have a dictionary mapping the plan to a function. Have a test function per plan, e.g. rename test() -> test_sanity(). Similarly you can define test_regression, test_functionality, etc.

SwapnilGaonkar7: @mssawant Functions for specific plan needs to go in test rpm. So there will be file which will be installed as part of hare test rpm which will receive command from our hare_setup and will decide which functions will be called based on plan selected while executing IVT.
Let me know your thoughts.

mssawant: Don't we atleast need dummy functions or a check atleast in this case. Because, the command (hare_setup test) is part of the cortx-hare rpm but the implmentation is part of a different rpm. Now, use can see the command and when tries to execute it will fail for the given plan because test rpm is not installed.

knekrasov: This problem with several RPMs can be resolved as follows:
1. Main Hare RPM provides this CLI to invoke tests. It parses the CLI parameters and extracts the "plan" name
2. All the tests are pytest-driven.
3. The tests that correspond to a certain plan can be either stored in separate folders (folder name corresponds to plan name somehow) or the plan is mapped to a [mark within the test](https://docs.pytest.org/en/6.2.x/example/markers.html).
4. Test RPM does nothing except deploying some _extended_ set of test files into the location which our `test` CLI is already aware of.

So in a nutshell `test` CLI will always invoke pytest within a given folder. The folder may be empty if test RPM is not installed.

SwapnilGaonkar7: @mssawant We will add check similar to below once we add code for accessing hare rpm functionality.

> if checkRpm('cortx-hare-test') != 0:
            logging.error('\'cortx-hare-test\' is not installed')
            exit(-1)

Let me know your inputs

SwapnilGaonkar7: @mssawant @knekrasov I have also raised PR for generating test rpm.
Once test rpm is installed, following commonad can be invoked from this interface
`/opt/seagate/cortx/hare/libexec/hare_test execute --plan sanity`

provisioning/setup.yaml
nitin-seagate: @SwapnilGaonkar7 The cortx CLI command is not one to one mapping to Mini-Provisioner interfaces...
For reference please check this section:
https://seagate-systems.atlassian.net/wiki/spaces/PRIVATECOR/pages/85000402/Mini-Provisioner+ConfStore+Keys#Test-Component-ConfStore%3A
Please define which keys you need to run Test RPM for Hare

nitin-seagate (changes requested):

provisioning/setup.yaml
SwapnilGaonkar7: @nitin-seagate Currently our test section do not use keys from confstore. We will update if that changes.
Please confirm if above setup.yaml changes need to be altered.

mssawant:
@SwapnilGaonkar7, kindly squash or rebase the branch with latest main.

provisioning/setup.yaml
nitin-seagate: @SwapnilGaonkar7  Please remove `-param $PARAM_URL` for now if you are not using any test keys

SwapnilGaonkar7: Done

SwapnilGaonkar7:
@mssawant @knekrasov I have addresses Nitin's comment and also done rebase.
Please have a look

provisioning/miniprov/hare_mp/main.py
knekrasov: As discussed in that PR `hare_test` doesn't seem to make sense

knekrasov (approved):

mssawant (approved):

PR #1667, opened 2021-06-22
EOS-21308: Updating Hare RPM version to 2.0.0

vaibhavparatwar:
Signed-off-by: Vaibhav Paratwar <vaibhav.paratwar@seagate.com>
Updating Hare RPM version to 2.0.0 tobe in sync wih R2


suvratjoshi:
PR Title can be this:
EOS-21308 : Update Hare rpm version to 2.0.0
JIra number and JIRA title separated by colon.
(we are not reverting but updating)

Please write description in PR description area.
Also same for commit message as well in case commit message is not written.

vaibhavparatwar:
Done. Suvrat.

knekrasov:
@vaibhavparatwar @suvratjoshi Why do we need to revert the version?

We're told that the version of Hare component is forcibly set to 2.0.0 anyways, so it doesn't really matter what is written in our VERSION file.

vaibhavparatwar:
> @vaibhavparatwar @suvratjoshi Why do we need to revert the version?
> 
> We're told that the version of Hare component is forcibly set to 2.0.0 anyways, so it doesn't really matter what is written in our VERSION file.

Yes correct. Just to make sure that this file is at its original as it was changed for the upgrade related JIRAs. Right now it does not really matter.. but its good to keep it to its original state.

knekrasov (approved):

mssawant (approved):

PR #1673, opened 2021-06-24
EOS-21245 Improve hax testability

knekrasov:
Problem: critical parts of hax functionality are not covered by the automatic tests. Things like integration with Consul and communication to Motr land are very hard to test.

Solution:
1. Refactor hax code so that it is possible to implement integration tests that would mock or emulate Consul and Motr
2. Introduce pytest test framework
3. Implement several sample integration tests showing how Consul and Motr-related functionality can be tested.
4. Implement basic infrastructure for running such tests (e.g. integrate them into Makefile).

### DIFF
--- a/hax/test/integration/test_server.py
+++ b/hax/test/integration/test_server.py
@@ -1,0 +1,154 @@
+# Copyright (c) 2021 Seagate Technology LLC and/or its Affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+import logging
+from base64 import b64encode
+from typing import List
+
+import pytest
+import simplejson
+from hax.message import BroadcastHAStates
+from hax.motr import WorkPlanner
+from hax.motr.delivery import DeliveryHerald
+from hax.server import ServerRunner
+from hax.types import Fid, HAState, MessageId, ServiceHealth
+
+
+@pytest.fixture
+def herald(mocker):
+    return DeliveryHerald()
+
+
+@pytest.fixture
+def planner(mocker) -> WorkPlanner:
+    def fake_add(cmd):
+        if hasattr(cmd, 'reply_to') and cmd.reply_to:
+            cmd.reply_to.put([MessageId(1, 42)])
+
+    planner = WorkPlanner()
+    mocker.patch.object(planner, 'add_command', side_effect=fake_add)
+    return planner
+
+
+@pytest.fixture
+async def hax_client(mocker, aiohttp_client, herald, planner, consul_util,
+                     loop):
+    srv = ServerRunner(planner, herald, consul_util)
+    srv._configure()
+    return await aiohttp_client(srv.app)
+
+
+async def test_hello_works(hax_client):
+    resp = await hax_client.get('/')
+    assert resp.status == 200
+    text = await resp.text()
+    assert text == "I'm alive! Sincerely, HaX"
+
+
+@pytest.mark.parametrize('status,health', [('passing', ServiceHealth.OK),
+                                           ('warning', ServiceHealth.FAILED),
+                                           ('critical', ServiceHealth.FAILED)])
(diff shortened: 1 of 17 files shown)


hax/test/integration/test_server.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+async def test_hello_works(hax_client):
+    resp = await hax_client.get('/')
+    assert resp.status == 200
+    text = await resp.text()
+    assert text == "I'm alive! Sincerely, HaX"
+
+
+@pytest.mark.parametrize('status,health', [('passing', ServiceHealth.OK),
```
mssawant, line 66: This seems like a decorator and would pass all these statuses one by one to the decorated function is it? If so, its pretty cool.

hax/test/integration/test_server.py
```diff
@@ -0,0 +1,157 @@
...
+                'ServiceID': '12',
+                'ServiceName': 'ios',
+            },
+        ],
+    }]
+    resp = await hax_client.post('/', json=service_health)
+    assert resp.status == 200
+    assert planner.add_command.called
+    planner.add_command.assert_called_once_with(
+        BroadcastHAStates(
```
mssawant, line 99: How about saving the msg_ids (tags) too and waiting for them to be complete as we do it in some cases in hax?

mssawant:
@knekrasov , can you please squash the commits?

hax/test/integration/test_server.py
knekrasov: @mssawant That would be ideal. But there are several reasons why I didn't do that:
1. Waiting for a delivery will mean that there are multiple threads within the test (one is the test runner, another one - "Motr emulator"). Single thread implementation will just block forever, deliveries always happen from another thread.
2. aiohttp-aware tests use `async` and `await`s. It is a bit tricky to combine asynchronous code with threads. Especially given that helper threads will have to be started and stopped properly. In Python it is not that easy to force a thread to stop (e.g. if it hangs on a blocking syscall).

So I think that we'll need to earn some more expertise in asyncio to write the tests like you described without the fear to hang forever. So I decided to start from simpler scenarios.

hax/test/integration/test_server.py
knekrasov: Yes, it is the way to feed test function with different inputs and provide expected outputs. And thus save dozens of lines of code :-)

hax/test/integration/test_server.py
mssawant: Okay, we can take it as an improvement later. But overall the changes look pretty awesome.

mssawant:
@knekrasov, The patch looks really good. I see the VM deployment and pre-merge tests were successful but somehow the status here is still showing `Pending`, not sure why.
May be we should try an explicit deployment and some sanity tests, you can check with @pavankrishnat.

mssawant (approved):

suvratjoshi:
Hi @knekrasov ,
If I understand this well enough : 
1. I think, it would be nice if we can have a RFC/design document related to this (I am sorry, if I am missing if any documentation was already created for this). If not present already, then can we have such document using a separate JIRA? The documentation JIRA can be worked out separately and we can merge this PR now.
2. A run-log / invoking of this testing can be showed in that document, highlighting the evidences of how its getting tested. What kind of log prints means its passing and what will happen/displayed, if something fails, etc.

Thank you!

suvratjoshi (approved):

knekrasov:
@suvratjoshi 
> I think, it would be nice if we can have a RFC/design document related to this (I am sorry, if I am missing if any documentation was already created for this). If not present already, then can we have such document using a separate JIRA? The documentation JIRA can be worked out separately and we can merge this PR now.

Sorry, I don't quite understand. What exactly do you suggest to document?

I think the only thing that is worth describing is `@trace_call` and mini-API around it. But I guess the code and docstrings are quite self-descriptive.

We'll have a training anyway where we can discuss integration tests.

> A run-log / invoking of this testing can be showed in that document, highlighting the evidences of how its getting tested/can be tested. What kind of log prints means its passing and what will happen/displayed, if something fails, etc.

We invoke `pytest` to run the tests. If it shows `PASSED`, then the test is good. Just have a look at sample output at this page: https://docs.pytest.org/en/6.2.x/example/markers.html

mssawant:
@knekrasov, can you please resolve conflicts and update the PR?

knekrasov:
@mssawant Rebased. Please have a look.

mssawant:
VM deployment is failing,
```
10:44:14  test/test_delivery_herald.py::TestDeliveryHeraldAny::test_works_under_load hare_mp/main.py:38: error: Module "hax.util" has no attribute "ConsulKVBasic"
10:44:14  Found 1 error in 1 file (checked 9 source files)
10:44:14  make[4]: *** [mypy] Error 1
10:44:14  make[4]: Leaving directory `/root/rpmbuild/BUILD/cortx-hare/provisioning/miniprov'
10:44:14  make[3]: *** [check-miniprov] Error 2
```

PR #1679, opened 2021-06-25
EOS-21399 CORTX-HARE: Implement test rpm for Hare

SwapnilGaonkar7:
Modified Makefile and added new specfile to create test rpm.


hare_test.spec
```diff
@@ -0,0 +1,50 @@
...
+# limitations under the License.
+#
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+# build number
+%define h_build_num  %(test -n "$build_number" && echo "$build_number" || echo 1)
+
+Summary: Hare test
+Name: cortx-hare-test
```
knekrasov, line 22: `cortx-hare-test` or `cortx-hare-ivt`?

test/hare_test
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
+import argparse
```
knekrasov, line 2: Why do we need this script?

test/hare_test
```diff
@@ -0,0 +1,57 @@
...
+                        nargs=1,
+                        type=Plan,
+                        action='store')
+    return parser
+
+def execute(args):
+    logging.info('Executing plan (%s)', args.plan[0].value)
+
+def main():
+    p = argparse.ArgumentParser(description='Hare test suite')
```
knekrasov, line 104: All this could have happened at the level of `hare_setup test`. I am not sure that this script is actually needed.

hare_test.spec
SwapnilGaonkar7: Naming convention is like 'cortx-s3-test' or 'cortx-sspl-test'
So I think it should be 'cortx-hare-test'

test/hare_test
SwapnilGaonkar7: This script will get called as follows,
`/opt/seagate/cortx/hare/libexec/hare_test execute --plan sanity --cdf_path /var/lib/hare/cluster.yaml`

Our 'hare_setup test' will call this command in above format.

test/hare_test
SwapnilGaonkar7: We do have some vaidation like plan type etc but actul work is done here, so I think we should have arguments and supported values defined here as this is standalone rpm.

SwapnilGaonkar7:
@knekrasov I have replied to your comments, please have a look

hare_test.spec
knekrasov: Ok. Perhaps I saw `ivt` in some other pull requests today..

test/hare_test
knekrasov: @SwapnilGaonkar7 I'm afraid that "actual work" is nothing more than invoking a command like this 
`pytest <path to folders with tests> -E production -m <plan>`

And that command could have been issued right from `hare_setup` script. The only thing that `test` rpm brings is a set of test scripts that are deployed to `<path to folders with tests>`

Let's keep simple things simple.

mssawant: Agree, let's not have multiple command line interfaces. We can just have a test package with set of tests and `hare_setup test` would invoke them. We just need to make sure that a proper error message is generated in case the test cases are not available at the given path instead of ignoring or failing silently.

hare_test.spec
```diff
@@ -0,0 +1,55 @@
...
+#
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+# build number
+%define h_build_num  %(test -n "$build_number" && echo "$build_number" || echo 1)
+
+%define _python_bytecompile_errors_terminate_build 0
+
+Summary: Hare test
```
knekrasov, line 23: Suggestion: System tests bundle for Hare

hare_test.spec
```diff
@@ -0,0 +1,55 @@
...
+
+%build
+
+%install
+rm -rf %{buildroot}
+mkdir -p %{buildroot}/opt/seagate/cortx/hare/libexec/
+cp -rf test %{buildroot}/opt/seagate/cortx/hare/libexec/
+
+%post
+pip3 install pytest
```
knekrasov, line 48: This is a bad practice. Please don't invoke pip in `%post` section.

knekrasov (commented):
@SwapnilGaonkar7 Can you please list the files of the test RPM and post as a comment here?

mssawant:
@SwapnilGaonkar7, kindly elaborate the commit message.

vaibhavparatwar:
Closing this PR as this task is not valid as per new requirements.

PR #1686, opened 2021-07-01
EOS-21354: Consul always has <=1 retry-join address

knekrasov:
JIRA issue: [EOS-21354](https://jts.seagate.com/browse/EOS-21354)

In case when there are N nodes in Consul cluser, each Consul server
should have (N-1) IP addresses in retry-join list. Current
implementation adds not more than 1 retry-join address.

## Why it is a problem
When a node reboots Consul agent needs to join back to the cluster.
Servers in retry-join list are the ones that will be used to connect
back to the cluster (or form it in case when Raft quorum used to be
lost). If retry-join contains one address only, that address becomes
single point of failure: once that node is down, other Consul agents
will not be able to reconnect to the cluster.

## Solution
Modify bash scripts to allow multiple retry-join arguments.

### DIFF
--- a/utils/hare-bootstrap
+++ b/utils/hare-bootstrap
@@ -80,6 +80,11 @@ Try reinstalling Hare on the above mentioned node.";
 }
 
 get_server_nodes() {
+    # Produces the output like this:
+    #
+    # localhost 10.230.240.139
+    # localhost2 10.230.240.140
+    #
     jq -r '.servers[] | "\(.node_name) \(.ipaddr)"' \
        $conf_dir/consul-agents.json
 }
@@ -281,7 +286,14 @@ fi
 abort_if_RC_leader_election_is_impossible
 
 read _ join_ip <<< $(get_server_nodes | grep -w $(node-name))
-read _ join_peers <<< $(get_server_nodes | grep -vw $(node-name))
+
+# get_server_nodes will produce multiple lines. We'll join them with space.
+# Note: in case when get_server_nodes is empty list, consequent awk and tr
+# will exit with non-zero code.
+# In order to swallow this error, `|| true` is needed.
+
+join_peers_opt=$(get_server_nodes | grep -vw "$(node-name)" | awk '{print "--join " $2}' | \
+              tr '\n' ' ' || true)
 
 if [[ -z $join_ip ]]; then
     cat <<'EOF' >&2
@@ -314,7 +326,7 @@ if sudo systemctl --quiet is-active hare-consul-agent; then
 fi
 
 say 'Starting Consul server on this node...'
-join_peers_opt="${join_peers:+--join $join_peers}"
+
 # $join_ip is our bind_ip address
 mk-consul-env --mode server --bind $join_ip $join_peers_opt \
               --extra-options '-ui -bootstrap-expect 1'
--- a/utils/mk-consul-env
+++ b/utils/mk-consul-env
@@ -61,7 +61,7 @@ while true; do
         -h|--help)           usage; exit ;;
         -m|--mode)           mode=$2; shift 2 ;;
         -b|--bind)           bind_addr=$2; shift 2 ;;
-        -j|--join)           join_addr=$2; shift 2 ;;
+        -j|--join)           join_addr="$join_addr $2"; shift 2 ;;
         -e|--extra-options)  extra_opts=$2; shift 2 ;;
         --)                  shift; break ;;
         *)                   echo 'getopt: internal error...'; exit 1 ;;
@@ -81,8 +81,15 @@ sed -r -e "s/^(NODE=).*/\1$(node-name)/" \
     sudo tee $ENV_FILE >/dev/null
 
 if [[ -n $join_addr ]]; then
-    sudo sed -r -e "s/^(JOIN=).*/\1${join_addr:+-retry-join $join_addr}/" \
-             -i $ENV_FILE
+    # join_addr is not empty and it contains one or more ip addresses separated
+    # by space.
+    # Multiple server addresses must turn into multiple '-retry-join ipN'.
+    # For instance:
+    # '127.0.0.1 192.168.0.1' => '-retry-join 127.0.0.1 -retry-join 192.168.0.1'
+
+    retry_str="$(echo "$join_addr" | \
+                 awk '{for (i=1; i <= NF; ++i) printf " -retry-join "$i }')"
+    sudo sed -r -e "s/^(JOIN=).*/\1$retry_str/" -i $ENV_FILE
 fi
 
 if [[ -n $extra_opts ]]; then


andriytk:
If I'm not mistaken, the one retry-join address will be used only the very 1st time a Consul server agent with persistent storage is trying to connect to the cluster. Later on, after the agent is connected, it learns about all the rest nodes in the cluster and saves them to its persistent storage, so even if the retry-join node is not available any more, it still should be able to connect to the cluster.

But if the persistent storage is wiped out during the node restart, I agree - we have to specify all the rest server agents in the cluster to retry-join.

utils/hare-bootstrap
```diff
@@ -281,7 +286,12 @@ fi
 abort_if_RC_leader_election_is_impossible
 
 read _ join_ip <<< $(get_server_nodes | grep -w $(node-name))
-read _ join_peers <<< $(get_server_nodes | grep -vw $(node-name))
+
+# get_server_nodes will produce multiple lines. We'll join them with space.
+# Note: in case when get_server_nodes is empty list, consequent awk and tr will exit with non-zero code.
```
andriytk, line 291: Isn't this line too long? Please, make sure it fits 80 chars width.

utils/hare-bootstrap
```diff
@@ -281,7 +286,12 @@ fi
...
+# In order to swallow this error, `|| true` is needed.
+
+join_peers="$(get_server_nodes | grep -vw "$(node-name)" | awk '{print $2}' | tr '\n' ' ' || true)"
```
andriytk, line 294: Consider `xargs` - it might be a shorter and simple alternative, for example:
```
$ cat | xargs echo
aa
bb
cc
aa bb cc
```

utils/hare-bootstrap
```diff
@@ -314,7 +324,10 @@ if sudo systemctl --quiet is-active hare-consul-agent; then
 fi
 
 say 'Starting Consul server on this node...'
-join_peers_opt="${join_peers:+--join $join_peers}"
+
+# See https://wiki.bash-hackers.org/syntax/pe#use_an_alternate_value if you don't understand the syntax below.
```
andriytk, line 328: Again, too long line.

utils/mk-consul-env
```diff
@@ -81,8 +81,12 @@ sed -r -e "s/^(NODE=).*/\1$(node-name)/" \
     sudo tee $ENV_FILE >/dev/null
 
 if [[ -n $join_addr ]]; then
-    sudo sed -r -e "s/^(JOIN=).*/\1${join_addr:+-retry-join $join_addr}/" \
-             -i $ENV_FILE
+    # join_addr is not empty and it contains one or more ip addresses separated by space.
+    # Multiple server addresses must turn into multiple '-retry-join ipN'.
+    # For instance '127.0.0.1 192.168.0.1' => '-retry-join 127.0.0.1 -retry-join 192.168.0.1'
+
+    retry_str="$(echo "$join_addr" | tr ' ' '\n' | sed 's/\(^.*$\)/-retry-join \1/' | tr '\n' ' ')"
```
andriytk, line 88: This might be simpler (without the need for `tr`):
```
$ echo aa b c | sed 's/\(\<[[:alnum:]]\)/-retry-join \1/g'
-retry-join aa -retry-join b -retry-join c
```

andriytk (approved):
Looks good. Please, consider the suggested improvements/simplifications and fix the style issues (too long lines).

utils/hare-bootstrap
knekrasov: As far as I understand, `xargs` will replace last `tr` only. If so, I don't see much of simplification.

utils/mk-consul-env
knekrasov: Your version looks too cryptic. IMO the version with `tr` is longer but easier to understand.

@mssawant, @suvratjoshi What's your opinion?

utils/hare-bootstrap
knekrasov: Addressed

utils/hare-bootstrap
knekrasov: Addressed

utils/mk-consul-env
```diff
@@ -81,8 +81,15 @@ sed -r -e "s/^(NODE=).*/\1$(node-name)/" \
...
 if [[ -n $join_addr ]]; then
-    sudo sed -r -e "s/^(JOIN=).*/\1${join_addr:+-retry-join $join_addr}/" \
-             -i $ENV_FILE
+    # join_addr is not empty and it contains one or more ip addresses separated
+    # by space.
+    # Multiple server addresses must turn into multiple '-retry-join ipN'.
+    # For instance:
+    # '127.0.0.1 192.168.0.1' => '-retry-join 127.0.0.1 -retry-join 192.168.0.1'
+
+    retry_str="$(echo "$join_addr" | tr ' ' '\n' | \
```
mssawant, line 90: I think `join_addr` whill not have both the addresses because above we are shifting by 2 thus in case of multiple join ipaddresses but just single `--join` option will end up `join_addr` to have just 1 ipaddress, something like this,
```
[root@ssc-vm-1623:root] hctl bootstrap --mkfs /root/cluster-3-1-ios-ctrl-fail-node-fail.yaml
2021-07-01 16:34:24: Generating cluster configuration... OK
10.230.250.53 10.230.250.48
2021-07-01 16:34:27: Starting Consul server on this node...--join "10.230.250.53 10.230.250.48 "
"10.230.250.53 <<<<<< **join_addr** in mk-consul-env
........ OK
```
I think we need to mention `--join` option multiple times and append ipaddresses to `join_addr`

consider this,

```
[root@ssc-vm-1623:root] diff /home/520478/cortx-hare-fork/cortx-hare-fork-new/cortx-hare-1/utils/hare-bootstrap /opt/seagate/cortx/hare/libexec/hare-bootstrap
284c284,288
< read _ join_peers <<< $(get_server_nodes | grep -vw $(node-name))
---
> # read _ join_peers <<< $(get_server_nodes | grep -vw $(node-name))
> join_peers="$(get_server_nodes | grep -vw "$(node-name)" | awk '{print $2}' | \
>               tr '\n' ' ' || true)"
> # Remove extra trailing spaces
> join_peers=$(echo $join_peers | sed 's/ *$//g')
317c321,322
< join_peers_opt="${join_peers:+--join $join_peers}"
---
> # join_peers_opt="${join_peers:+--join $join_peers}"
> join_peers_opt="${join_peers:+$join_peers}"
319c324,326
< mk-consul-env --mode server --bind $join_ip $join_peers_opt \
---
> join_str="$(echo "$join_peers_opt" | tr ' ' '\n' | \
>                  sed 's/\(^.*$\)/--join \1/' | tr '\n' ' ')"
> mk-consul-env --mode server --bind $join_ip $join_str \
358,364c365,371
<     if (( $count > 5 )); then
<         echo 'Some agent(s) failed to start in due time:' >&2
<         diff <(get_ready_agents | sort) \
<              <(get_all_nodes | awk '{print $1}' | sort) | sed 1d >&2
<         echo 'Check connectivity and firewall (Consul ports must be opened)' >&2
<         exit 1
<     fi
---
> #     if (( $count > 5 )); then
> #         echo 'Some agent(s) failed to start in due time:' >&2
> #         diff <(get_ready_agents | sort) \
> #              <(get_all_nodes | awk '{print $1}' | sort) | sed 1d >&2
> #         echo 'Check connectivity and firewall (Consul ports must be opened)' >&2
> #         exit 1
> #     fi
367c374
<     (( count++ ))
---
> #     (( count++ ))
369a377,378
>
> exit 0
[root@ssc-vm-1623:root] diff /home/520478/cortx-hare-fork/cortx-hare-fork-new/cortx-hare-1/utils/mk-consul-env /opt/seagate/cortx/hare/libexec/mk-consul-env
64c64
<         -j|--join)           join_addr=$2; shift 2 ;;
---
>         -j|--join)           join_addr+=$2; join_addr+=' '; shift 2 ;;
82a83,84
> # Remove extra trailing spaces
> join_addr=$(echo $join_addr | sed 's/ *$//g')
84,85c86,90
<     sudo sed -r -e "s/^(JOIN=).*/\1${join_addr:+-retry-join $join_addr}/" \
<              -i $ENV_FILE
---
>     # sudo sed -r -e "s/^(JOIN=).*/\1${join_addr:+-retry-join $join_addr}/" \
>     #          -i $ENV_FILE
>     retry_str="$(echo "$join_addr" | tr ' ' '\n' | \
>                      sed 's/\(^.*$\)/-retry-join \1/' | tr '\n' ' ')"
>     sudo sed -r -e "s/^(JOIN=).*/\1$retry_str/" -i $ENV_FILE
```

suvratjoshi (approved):

utils/mk-consul-env
mssawant: I am okay with `tr` too, patch as it is did not work as expected for me. Please find my comments and suggested changes.

knekrasov: Nevermind, `[:alnum:]` didn't work for IP addresses anyway. I've re-written this with `awk` to address @mssawant's finding.

utils/mk-consul-env
knekrasov: Addressed. @mssawant Can you please have a look once again?

mssawant (approved):
Looks good.

PR #1687, opened 2021-07-01
Problem: hax_stop invoked twice

knekrasov:
notify_hax_stop() function sends stop event twice. Firstly it sends
notification using hax process fid, then using profile fid.

It waits for delivery only once by some reason.

Solution: send one notification only, use hax fid and wait for delivery.


mssawant (approved):

suvratjoshi:
@knekrasov 
PR title to include JIRA number please.

suvratjoshi (approved):

knekrasov:
@suvratjoshi We have no JIRA ticket for this. I'm not sure that the problem is observable.

PR #1694, opened 2021-07-06
EOS-21919 Custom-build-1901: 3-node deployment fails at hare

SwapnilGaonkar7:
Problem: In our current code, we have timeout of 5 sec for consul start. But we have seen that sometimes consul takes more than 5 secs to start causing timeout and deployment failure. 

Solution: Changed logic for consul start check, removed timeout and added code for service failure check so we will keep checking until either consul is started or enters in failed state.

Tested bootstrap on 1 node and 3 node.
Tested on setup on which issue was seen.

### DIFF
--- a/utils/hare-bootstrap
+++ b/utils/hare-bootstrap
@@ -204,6 +204,17 @@ is_localhost() {
     [[ -e $path && $(cat $path) == $node ]] && return 0 || return 1
 }
 
+check_consul_failure() {
+    while read node bind_ip; do
+        if ssh -n $node "sudo systemctl --quiet --state=failed is-failed hare-consul-agent"; then
+            echo 'Failed to start agent on '$node: >&2
+            true; return
+        fi
+    done < <(get_all_nodes)
+
+    false; return
+}
+
 # --------------------------------------------------------------------
 # main
 
@@ -363,22 +374,29 @@ while read node bind_ip; do
 done < <(get_client_nodes)
 wait4 ${pids[@]-}
 agents_nr=$(( ${#pids[@]} + 1 ))
+count=1
 
 # Waiting for the Consul agents to get ready...
-count=1
 while (( $(get_ready_agents_nr) != $agents_nr )); do
     if (( $count > 5 )); then
-        echo 'Some agent(s) failed to start in due time:' >&2
-        diff <(get_ready_agents | sort) \
-             <(get_all_nodes | awk '{print $1}' | sort) | sed 1d >&2
-        echo 'Check connectivity and firewall (Consul ports must be opened)' >&2
-        exit 1
+        echo 'Consul not ready on node :'
+        comm -13 <(get_ready_agents | sort) \
+                 <(get_all_nodes | awk '{print $1}' | sort) >&1
+        count=1
+
+        if check_consul_failure; then
+            echo 'Some agent(s) failed to start:' >&2
+            comm -13 <(get_ready_agents | sort) \
+                     <(get_all_nodes | awk '{print $1}' | sort) >&2
+            echo 'Check connectivity and firewall (Consul ports must be opened)' >&2
+            exit 1
+        fi
     fi
-    echo -n '.'
-    sleep 1
     (( count++ ))
+    
+    sleep 1
 done
-echo ' OK'
+echo 'Consul ready on all nodes'
 
 say 'Updating Consul configuraton from the KV store...'
 update-consul-conf &


knekrasov (commented):
@SwapnilGaonkar7 Please don't forget to state the problem you're solving by this patch. Thanks!

SwapnilGaonkar7:
@knekrasov Updated description

knekrasov:
Thank you @SwapnilGaonkar7, now it is clearer.

But why do we remove timeouts at all? Per my understanding, consul doesn't start immediately. `systemctl start` doesn't guarantee that when it returns the service is started for sure. Is it safe to have no timeout at all?

SwapnilGaonkar7:
@knekrasov Actually we are not relying on systemctl to see if consul is started, we are checking 'consul members' to see if consul is started and members are correctly added. We will be using systemctl to see if consul entered in failed state or not. In that case we will return failure.
For timeout we can have some higher timeout like 10-15 mins which will hit rare case. But according to me, we should not have shorter timeout as it might cause above mentioned failure again on slower systems
@mssawant Please provide your input also

mssawant:
@SwapnilGaonkar7 please change update the draft status of the PR.

mssawant (approved):

mssawant:
>For timeout we can have some higher timeout like 10-15 mins which will hit rare case. But according to me, we should not have shorter timeout as it might cause above mentioned failure again on slower systems
@mssawant Please provide your input also

@SwapnilGaonkar7, @knekrasov , Yeah, coming up with ideal timeout values are tricky in such cases where some environmental factors (network delay) might make it obsolete. So, I think overall it is good to keep checking for Consul until the failure. However we can log some message, e.g. `Consul not ready on node <node-name>` periodically, may be every 5 secs while waiting for Consul to start.

SwapnilGaonkar7:
@mssawant  @knekrasov  Please have a look

utils/hare-bootstrap
```diff
@@ -199,6 +199,19 @@ is_localhost() {
     [[ -e $path && $(cat $path) == $node ]] && return 0 || return 1
 }
 
+check_consul_failure() {
+
+    while read node bind_ip; do
+        if ssh -n $node "sudo systemctl --quiet --state=failed is-failed hare-consul-agent"; then
```
knekrasov, line 205: Can we use `ssh_error_info` instead?

utils/hare-bootstrap
```diff
@@ -353,20 +366,22 @@ wait4 ${pids[@]-}
 agents_nr=$(( ${#pids[@]} + 1 ))
 
 # Waiting for the Consul agents to get ready...
-count=1
 while (( $(get_ready_agents_nr) != $agents_nr )); do
-    if (( $count > 5 )); then
-        echo 'Some agent(s) failed to start in due time:' >&2
-        diff <(get_ready_agents | sort) \
-             <(get_all_nodes | awk '{print $1}' | sort) | sed 1d >&2
+    echo 'Consul not ready on node :'
```
knekrasov, line 370: Please note that this output goes to stdout while consequent `comm` will print the output to stderr.

Doesn't it make sense to redirect output of `echo` to stderr as well?

utils/hare-bootstrap
SwapnilGaonkar7: @knekrasov ssh_error_info throws error if command returns false.
In our case, command will return false in happy path. So we cant use ssh_error_info.
Please correct me if I am wrong

knekrasov: Hm, good point. Ok, please ignore then.

utils/hare-bootstrap
SwapnilGaonkar7: Done, I have redirected both to stdout as they are not really error as yet

knekrasov (approved):

utils/hare-bootstrap
```diff
@@ -199,6 +199,19 @@ is_localhost() {
     [[ -e $path && $(cat $path) == $node ]] && return 0 || return 1
 }
 
+check_consul_failure() {
+
+    while read node bind_ip; do
+        if ssh -n $node "sudo systemctl --quiet --state=failed is-failed hare-consul-agent"; then
+            echo 'Failed to start agent on '$node: >&2
+            return 0
```
mssawant, line 207: @SwapnilGaonkar7, why are we returning `0` if consul fails to start? shouldn't we be returning 1 and 0 in case of success?

utils/hare-bootstrap
mssawant, line 370: I think instead of checking in every iteration, which would generate more logs, let's keep the 5 secs timeout and check_consul_failure every 5 secs as earlier, just don't exit after 5 secs. Only if check_consul_failure returns non-zero (i.e. consul failed to start) the script must exit with an error.

mssawant (approved):

mssawant (changes requested):
Approved mistakenly, kindly address the comments.

utils/hare-bootstrap
SwapnilGaonkar7: `why are we returning 0 if consul fails to start`
@mssawant  This function checks if consul is in failed state that's why we are returning 0(success) if consul fails to start

utils/hare-bootstrap
SwapnilGaonkar7: @mssawant  Done

utils/hare-bootstrap
mssawant: @SwapnilGaonkar7 , ideally such functions return boolean, I see that you want to return True or False here. So why not do that instead? return True or False instead of 0 or 1, the code will be more readable, no?

utils/hare-bootstrap
```diff
@@ -199,6 +199,19 @@ is_localhost() {
     [[ -e $path && $(cat $path) == $node ]] && return 0 || return 1
 }
 
+check_consul_failure() {
+
```
suvratjoshi, line 203: Please remove, if there is extra blank line here.

SwapnilGaonkar7: Done

utils/hare-bootstrap
SwapnilGaonkar7: @mssawant  We can not write 'return true' in bash. Return only accept numeric value
Nearest match is 'true; return'
I have modified code accordingly

mssawant: Yeah, makes sense.

mssawant (approved):

PR #1695, opened 2021-07-07
EOS-21420: motr rconfc_stop fails during hax shutdown

mssawant:
With the addition of multiple Hax consumer threads, motr::stop()
was invoked from multiple threads which lead to multiple invocations
of rconfc_stop() during Hax shutdown. Motr rconfc might get finalised
in one of these calls. Thus a subsequent call to rconfc_stop would
fail due to already destroyed or invalid structure.

Solution:
- Allow only consumer thread 0 to invoke motr::stop() and finalise
rconfc.

### DIFF
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -57,6 +57,7 @@ class ConsumerThread(StoppableThread):
         self.consul = consul
         self.eq_publisher = EQPublisher()
         self.herald = herald
+        self.idx = idx
 
     def stop(self) -> None:
         self.is_stopped = True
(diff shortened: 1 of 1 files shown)


knekrasov (approved):
Good catch!

suvratjoshi (approved):

suvratjoshi:
@mssawant,
Any RFC/doc updating needed?

vaibhavparatwar:
Tests run:
1. Full stack deployment done with custom build. 
2. s3bench I/Os run with 10MB object size and it worked fine. No hax crash in /var/log/crash and /var/crash observed.
3. s3bench I/Os with 1GB had 88 read errors, no write errors. It is separate issue and tracked with EOS-22175. 
4. Running hare_setup init in loop. 30+ loops completed without any panic / crash or failure.

PR #1697, opened 2021-07-07
EOS-15932 No verbose mode present in hctl-shutdown

SwapnilGaonkar7:
Used argparse to handle argument and modified code to enable veebose mode when -v is passed


utils/hare-shutdown
```diff
@@ -109,23 +110,30 @@ def node_shutdown():
...
+                        help="Node shutdown",
+                        action="store_true")
+
+    args = parser.parse_args()
+    if args.verbose:
+        logging.basicConfig(level=logging.DEBUG, format='%(message)s')
+    else:
+        logging.basicConfig(level=logging.INFO, format='%(message)s')
+
+    skip_consul_stop = True if args.skip_consul_stop else False
```
knekrasov, line 131:
```suggestion
    skip_consul_stop = bool(args.skip_consul_stop)
```

SwapnilGaonkar7: Done

knekrasov (approved):

utils/hare-shutdown
```diff
@@ -109,23 +108,30 @@ def node_shutdown():
...
+    parser.add_argument("--skip-consul-stop",
+                        help="Skip consul service stop",
+                        action="store_true")
+    parser.add_argument("--node",
+                        help="Node shutdown",
+                        action="store_true")
+
+    args = parser.parse_args()
+    if args.verbose:
+        logging.basicConfig(level=logging.DEBUG, format='%(message)s')
```
mssawant, line 125: Do we want it the other way around, i.e. Info in case of verbose and Debug otherwise?

knekrasov: Verbose is usually understood as 'too wordy' or  'reporting (too) much'. 

In case DEBUG level is set, log messages of both INFO and DEBUG will be seen. If INFO level is set, only INFO messages will be seen. So per my understanding, the code makes sense.

SwapnilGaonkar7: @mssawant Agree with Konstantin. Verbose should produce more detailed logs.

mssawant: >In case DEBUG level is set, log messages of both INFO and DEBUG will be seen. If INFO level is set, only INFO messages will be seen

@knekrasov, @SwapnilGaonkar7, is this a universal assumption or just with respect to Python? I think in any case we should explicitly state this assumption in form of a comment here itself.

mssawant (approved):
Minor documentation comment.

utils/hare-shutdown
SwapnilGaonkar7: @mssawant  Done

mssawant (approved):

PR #1698, opened 2021-07-07
EOS-22227: Hare reports device failures for motr m0mkfs process

mssawant:
Motr panic: (!pver->pv_is_dirty) is seen in motr-mkfs completion window,
where the Consul KV for motr process status is updated to STARTED for motr-mkfs
process and Consul reports warning for the same, this happens when motr-mkfs
start shuting down but has not reported STOPPING yet or due to delayed Consul
warning notification. Hare thinks that process crashed because Consul reported
a warning and process status remained started but not stopping or stopped.

Solution:
This needs to be handled by storing additional information and checking for
the process type (M0MKFS or M0D) along with process status.
Thus if the process status evaluates to FAILED based on Consul service status
and present process status in Consul KV but the process type is M0MKFS then
report STOPPED instead of FAILED so that no device failures are reported for
the process. This needs to be handled for motr process type M0D in-order to
handle a cluster node shutdown, where the devices must be marked FAILED to
facilitate degraded IO. For motr process type M0D this situation is not expected
because Motr process is expected to be running if its corresponding Consul KV
status is STARTED or the current Consul service status for the same is warning
if the process is genuinely stopped or the corresponding node itself has stopped.

### DIFF
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -138,7 +138,8 @@ class ConsumerThread(StoppableThread):
                     self.consul.update_process_status(
                         ConfHaProcess(
                             chp_event=event,
-                            chp_type=m0HaProcessType.M0_CONF_HA_PROCESS_M0D,
+                            chp_type=int(
+                                m0HaProcessType.M0_CONF_HA_PROCESS_M0D),
                             chp_pid=0,
                             fid=state.fid))
                 new_ha_states.append(
--- a/hax/hax/util.py
+++ b/hax/hax/util.py
@@ -50,6 +50,14 @@ ServiceData = NamedTuple('ServiceData', [('node', str), ('fid', Fid),
 
 FidWithType = NamedTuple('FidWithType', [('fid', Fid), ('service_type', str)])
 
+MotrConsulProcStatus = NamedTuple('MotrConsulProcStatus', [(
+                                        'consul_svc_status', str),
+                                        ('consul_motr_proc_status', str)])
+
+MotrProcStatusLocalRemote = NamedTuple('MotrProcStatusLocalRemote', [(
+                                'motr_proc_status_local', ServiceHealth),
+                                ('motr_proc_status_remote', ServiceHealth)])
+
 
 def mkServiceData(service: Dict[str, Any]) -> ServiceData:
     return ServiceData(
@@ -87,6 +95,7 @@ ha_process_events = ('M0_CONF_HA_PROCESS_STARTING',
                      'M0_CONF_HA_PROCESS_STARTED',
                      'M0_CONF_HA_PROCESS_STOPPING',
                      'M0_CONF_HA_PROCESS_STOPPED')
+
 
 ha_conf_obj_states = ('M0_NC_UNKNOWN',
                       'M0_NC_ONLINE',
@@ -372,7 +381,7 @@ class ConsulUtil:
 
     def get_svc_status(self, srv_fid: Fid) -> str:
         try:
-            return self.get_process_status(srv_fid)
+            return self.get_process_status(srv_fid).consul_svc_status
         except Exception:
             return 'Unknown'
 
@@ -512,10 +521,9 @@ class ConsulUtil:
             pfid = self.get_service_process_fid(svc_fid)
         else:
             pfid = create_process_fid(fidk)
-        if (self.get_process_status(pfid) in ('M0_CONF_HA_PROCESS_STARTING',
-                                              'M0_CONF_HA_PROCESS_STARTED',
-                                              'M0_CONF_HA_PROCESS_STOPPING',
-                                              'Unknown')):
+        if self.get_process_status(pfid).consul_svc_status in (
+                'M0_CONF_HA_PROCESS_STARTING', 'M0_CONF_HA_PROCESS_STARTED',
+                'M0_CONF_HA_PROCESS_STOPPING', 'Unknown'):
             return HaNoteStruct.M0_NC_ONLINE
         else:
             return HaNoteStruct.M0_NC_FAILED
@@ -693,7 +701,8 @@ class ConsulUtil:
         assert 0 <= event.chp_event < len(ha_process_events), \
             f'Invalid event type: {event.chp_event}'
         local_node = self.get_local_nodename()
-        data = json.dumps({'state': ha_process_events[event.chp_event]})
+        data = json.dumps({'state': ha_process_events[event.chp_event],
+                           'type': m0HaProcessType(event.chp_type).name})
         # Maintain statuses for all the motr processes in the cluster
         # for every node so that in case of 1 or more node failures,
         # as every node will receive the node failure event, the failed
@@ -824,14 +833,15 @@ class ConsulUtil:
         return self.sdev_to_drive_fid(sdev_fid)
 
     # Returns status of process from its local node.
-    def get_process_status(self, fid: Fid) -> str:
+    def get_process_status(self, fid: Fid) -> MotrConsulProcStatus:
         proc_node = self.get_process_node(fid)
         key = f'{proc_node}/processes/{fid}'
         status = self.kv.kv_get(key)
         if status:
-            return str(json.loads(status['Value'])['state'])
+            val = json.loads(status['Value'])
+            return MotrConsulProcStatus(val['state'], val['type'])
         else:
-            return 'Unknown'
+            return MotrConsulProcStatus('Unknown', 'Unknown')
 
     def get_process_local_status(self, fid: Fid) -> str:
         local_node = self.get_local_nodename()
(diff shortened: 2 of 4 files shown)


hax/hax/util.py
```diff
@@ -939,23 +946,28 @@ def get_service_health(self, node: str, svc_id: int) -> ServiceHealth:
...
+                    LOG.debug('consul.status %s svc_health: %s',
+                              str(cns_status), str(svc_health))
                     local_node = self.get_local_nodename()
                     proc_node = self.get_process_node(pfid)
                     if proc_node == local_node:
                         status = svc_health[0]
                     else:
                         status = svc_health[1]
+                    if (status == ServiceHealth.FAILED and
+                            cns_status[1] == 'M0_CONF_HA_PROCESS_M0MKFS'):
```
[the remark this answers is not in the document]
mssawant: Agree, fixed.

hax/hax/util.py
```diff
@@ -1055,7 +1067,7 @@ def service_health_to_m0dstatus_update(self, proc_fid: Fid,
         self.update_process_status(ev)
 
     def is_confd_failed(self, proc_fid: Fid) -> bool:
-        status = self.get_process_status(proc_fid)
+        status = self.get_process_status(proc_fid)[0]
```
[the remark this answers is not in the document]
mssawant: Addressed.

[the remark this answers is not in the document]
mssawant: Addressed. Kindly review the updated patch.

hax/hax/util.py
```diff
@@ -939,23 +946,28 @@ def get_service_health(self, node: str, svc_id: int) -> ServiceHealth:
                     if item['Status'] == 'critical':
                         return ServiceHealth.FAILED
                     pfid = create_process_fid(svc_id)
-                    svc_consul_status = self.get_svc_status(pfid)
+                    cns_status = self.get_process_status(pfid)
                     svc_health = svc_to_motr_status_map[(item['Status'],
-                                                         svc_consul_status)]
+                                                         cns_status[0])]
+                    LOG.debug('consul.status %s svc_health: %s',
+                              str(cns_status), str(svc_health))
```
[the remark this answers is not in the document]
mssawant: Indeed, fixed.

hax/hax/handler.py
```diff
@@ -138,7 +138,8 @@ def update_process_failure(self, planner: WorkPlanner,
                     self.consul.update_process_status(
                         ConfHaProcess(
                             chp_event=event,
-                            chp_type=m0HaProcessType.M0_CONF_HA_PROCESS_M0D,
+                            chp_type=(
+                                m0HaProcessType.M0_CONF_HA_PROCESS_M0D.value),
```
knekrasov, line 142: Side note:
Probably it makes sense to make `m0HaProcessType` to extend IntEnum instead of just Enum. In that case `.value` is not needed (`int(enum)` is sufficient); more info here: https://stackoverflow.com/a/49626802.

I don't think that this comment should be addressed in this PR. But probably you could add FIXME for the future?

hax/hax/util.py
```diff
@@ -512,10 +521,11 @@ def get_conf_obj_status(self, obj_t: ObjT, fidk: int) -> int:
...
             pfid = create_process_fid(fidk)
-        if (self.get_process_status(pfid) in ('M0_CONF_HA_PROCESS_STARTING',
-                                              'M0_CONF_HA_PROCESS_STARTED',
-                                              'M0_CONF_HA_PROCESS_STOPPING',
-                                              'Unknown')):
+        if ((self.get_process_status(pfid)).consul_svc_status in (
+                                                 'M0_CONF_HA_PROCESS_STARTING',
+                                                 'M0_CONF_HA_PROCESS_STARTED',
+                                                 'M0_CONF_HA_PROCESS_STOPPING',
+                                                 'Unknown')):
```
knekrasov, line 528: [Optional]
```suggestion
        if self.get_process_status(pfid).consul_svc_status in (
                'M0_CONF_HA_PROCESS_STARTING', 'M0_CONF_HA_PROCESS_STARTED',
                'M0_CONF_HA_PROCESS_STOPPING', 'Unknown'):
```
(parentheses are redundant, as a result the code can be more compact)

hax/hax/util.py
```diff
@@ -693,7 +703,8 @@ def update_process_status(self, event: ConfHaProcess) -> None:
         assert 0 <= event.chp_event < len(ha_process_events), \
             f'Invalid event type: {event.chp_event}'
         local_node = self.get_local_nodename()
-        data = json.dumps({'state': ha_process_events[event.chp_event]})
+        data = json.dumps({'state': ha_process_events[event.chp_event],
+                           'type': str(m0HaProcessType(event.chp_type).name)})
```
knekrasov, line 707: [Optional]: `str` looks redundant

hax/hax/util.py
```diff
@@ -824,14 +835,16 @@ def node_to_drive_fid(self, node_name: str, drive: str):
...
-    def get_process_status(self, fid: Fid) -> str:
+    def get_process_status(self, fid: Fid) -> MotrConsulProcStatus:
         proc_node = self.get_process_node(fid)
         key = f'{proc_node}/processes/{fid}'
         status = self.kv.kv_get(key)
         if status:
-            return str(json.loads(status['Value'])['state'])
+            return (MotrConsulProcStatus(
+                        str(json.loads(status['Value'])['state']),
+                        str(json.loads(status['Value'])['type'])))
```
knekrasov, line 845: Value is parsed twice.

hax/hax/util.py
```diff
@@ -1055,7 +1078,7 @@ def service_health_to_m0dstatus_update(self, proc_fid: Fid,
         self.update_process_status(ev)
 
     def is_confd_failed(self, proc_fid: Fid) -> bool:
-        status = self.get_process_status(proc_fid)
+        status = (self.get_process_status(proc_fid)).consul_svc_status
```
knekrasov, line 1081: [Optional]: Parentheses are redundant.

hax/hax/util.py
```diff
@@ -372,7 +381,7 @@ def get_session_node(self, session_id: str) -> str:
 
     def get_svc_status(self, srv_fid: Fid) -> str:
         try:
-            return self.get_process_status(srv_fid)
+            return (self.get_process_status(srv_fid)).consul_svc_status
```
knekrasov, line 384: [Optional]
```suggestion
            return self.get_process_status(srv_fid).consul_svc_status
```

hax/hax/util.py
knekrasov: Here you change the type of value stored in Consul KV. Two comments:
1. Such a change should be reflected in 4/KV RFC.
2. Note that the value is not atomic. Are we sure that this is the final structure of the JSON and in the future versions we'll not want to change it again? I'm just afraid to introduce problems for upcoming SW upgrades.

hax/hax/util.py
```diff
@@ -896,35 +909,38 @@ def get_service_health(self, node: str, svc_id: int) -> ServiceHealth:
...
-            ('warning',
-             'M0_CONF_HA_PROCESS_STOPPED'): (ServiceHealth.STOPPED,
-                                             ServiceHealth.STOPPED),
-            ('warning',
-             'M0_CONF_HA_PROCESS_STARTING'): (ServiceHealth.OFFLINE,
-                                              ServiceHealth.OFFLINE),
-            ('warning',
-             'Unknown'): (ServiceHealth.UNKNOWN, ServiceHealth.UNKNOWN)}
+            MotrConsulProcStatus('passing', 'M0_CONF_HA_PROCESS_STARTING'):
+            MotrProcStatusLocalRemote(ServiceHealth.OFFLINE,
```
knekrasov, line 915: [Optional]:
These enum names are VERY long which makes the code a bit unreadable. Please note that it could be simplified a bit by using reference variables like this:
```python
        local_remote = MotrProcStatusLocalRemote
        proc_status = MotrConsulProcStatus

        svc_to_motr_status_map = {
            proc_status('passing', 'M0_CONF_HA_PROCESS_STARTING'):
            local_remote(ServiceHealth.OFFLINE, ServiceHealth.OK),
            proc_status('passing', 'M0_CONF_HA_PROCESS_STOPPING'):
            local_remote(ServiceHealth.OFFLINE, ServiceHealth.UNKNOWN),
            proc_status('passing', 'M0_CONF_HA_PROCESS_STARTED'):
            local_remote(ServiceHealth.OK, ServiceHealth.OK),
```
and so forth.
Optional, so feel free to ignore if you don't like this.

knekrasov (commented):
In general looks good, there is only one serious concern regarding KV schema.

hax/hax/util.py
mssawant: Will update 4/KV RFC.
I think its okay, as we have not released a kv version yet. We should avoid KV schema changes and try to identify and make changes as early as possible.

hax/hax/util.py
mssawant: Fixed.

hax/hax/util.py
mssawant: Removed.

hax/hax/handler.py
mssawant: Addressed, thanks.

hax/hax/util.py
mssawant: Addressed.

hax/hax/util.py
mssawant: Addressed.

SwapnilGaonkar7 (approved):

suvratjoshi:
@mssawant 
Not very sure but just a reminder in case its missed.
Does this need any changes in doc : `doc\hare-dtm-hld.md` and related diagrams?

mssawant:
@suvratjoshi , No change is required in hare-dtm-hld as we are not introducing any new events between Hare and Motr. We will be updating the hld anyway as part of hare dtm work if required.

suvratjoshi (approved):

PR #1701, opened 2021-07-08
EOS-22328 Modify formula for calculating allowed failure vector

SwapnilGaonkar7:
For 4+2+0,
`- allowed_failures:
    disk: 2
    ctrl: 2
    encl: 1
    rack: 0
    site: 0`

For 4+2+2,
`- allowed_failures:
    disk: 2
    ctrl: 2
    encl: 0
    rack: 0
    site: 0`

### DIFF
--- a/provisioning/miniprov/hare_mp/cdf.py
+++ b/provisioning/miniprov/hare_mp/cdf.py
@@ -3,24 +3,17 @@ import subprocess as S
 from dataclasses import dataclass
 from string import Template
 from typing import Any, Dict, List, Optional, Tuple
-from math import floor
+from math import floor, ceil
 import pkg_resources
 
 from hare_mp.store import ValueProvider
 from hare_mp.types import (ClusterDesc, DiskRef, DList, Maybe, NodeDesc,
                            PoolDesc, PoolType, ProfileDesc, Protocol, Text,
-                           M0ServerDesc, DisksDesc, AllowedFailures)
+                           M0ServerDesc, DisksDesc, AllowedFailures, Layout)
 
 DHALL_PATH = '/opt/seagate/cortx/hare/share/cfgen/dhall'
 DHALL_EXE = '/opt/seagate/cortx/hare/bin/dhall'
 DHALL_TO_YAML_EXE = '/opt/seagate/cortx/hare/bin/dhall-to-yaml'
-
-
-@dataclass
-class Layout:
-    data: int
-    parity: int
-    spare: int
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


knekrasov (approved):

mssawant:
@SwapnilGaonkar7 can you please add more details to the commit message? also I think it would be good add the overall algorithm as a comment in the code itself.

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -148,22 +148,19 @@ def _add_pool(self, pool: PoolHandle, out_list: List[PoolDesc]) -> None:
         cvg_per_node = int(conf.get(
             f'server_node>{machine_id}>storage>cvg_count'))
 
-        cvg_per_storage_set = cvg_per_node * node_count
-
-        if layout.data + layout.parity + layout.spare == 0:
-            raise RuntimeError('All layout parameters are 0')
+        total_unit = layout.data + layout.parity + layout.spare
```
mssawant, line 151: let's move the code to calculate allowed_failures vector to a function. And add the algorithm as a comment as a function header.

SwapnilGaonkar7: @mssawant Done

provisioning/miniprov/test/test_cdf.py
```diff
@@ -221,6 +221,149 @@ def ret_values(value: str) -> Any:
         self.assertEqual(1, len(ret[0].pools.value))
         self.assertEqual(Text('StorageSet-1__sns'), ret[0].pools.value[0])
 
+    def test_allowed_failure_generation_4_2_0(self):
```
mssawant, line 224: @SwapnilGaonkar7 , This is good but instead of 2 functions, one for each layout (4+2+0 and 4+2+2), I think we can have just one function and pass layout params to it. what do you think?

mssawant: It would be great if we can run the function and verify the allowed_failures vector for all the params listed at https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/127893680/CORTX+Durability+Configuration.

provisioning/miniprov/hare_mp/cdf.py
```diff
@@ -132,37 +132,47 @@ def _validate_pool(self, pool: PoolHandle) -> None:
...
+        enc_failure_allowed_FP = layout.parity / ceil(total_unit / node_count)
+        enc_failure_allowed = floor(enc_failure_allowed_FP)
 
-        if ratio == 0:
-            ctrl_failure_allowed = 0
+        cvg_per_storage_set = cvg_per_node * node_count
+        if cvg_per_storage_set < layout.parity:
+            ctrl_failure_allowed = cvg_per_storage_set
         else:
             ctrl_failure_allowed = layout.parity
```
mssawant, line 158: This will not work for 8 + 2 + 0, 3 nodes, 1 cvg per node and 4 devices per cvg.
encl_failures = 0 (correct)
ctrl_failures = 2 (wrong) we cannot support a ctrl failure as 4 devices will fail and per controller there will be atleast 3 parity group units. So we can ideally support just 2 drive failures. 
We need a validation function for this. There is one in the Motr, I think we need to port it to Hare for validations.

SwapnilGaonkar7: I had quick call with Madhavrao for this.
@madhavemuri Please let us know once you are done with modification of formula on confluence page.

madhavemuri: https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/127893680/CORTX+Durability+Configuration
I have updated the page, which should take care of above case as well.

suvratjoshi (approved):

provisioning/miniprov/hare_mp/cdf.py
mssawant: @madhavemuri, Okay, the change makes sense.
@SwapnilGaonkar7 can you please make the change in the formula?

SwapnilGaonkar7: @mssawant I have modified formula accordingly

mssawant: @SwapnilGaonkar7 as per the formula,
```
min (A*#NodeFailure, K)
```
you need take the minimum of cvg failures or K, I don't see that here.

SwapnilGaonkar7: temp = cvg_per_node * enc_failure_allowed
        if temp < layout.parity:
            ctrl_failure_allowed = temp
        else:
            ctrl_failure_allowed = layout.parity

This code should handle above condition where A is cvg_per_node

mssawant (approved):

PR #1707, opened 2021-07-14
EOS-22106: Devices are marked offline on process failure

Shreya-18:
### Problem:
Devices are marked failed, when the corresponding ios service
fails. However, when the service comes back online, device states
remains failed as only explicit device events can mark device online.

### Solution: 
Marking the device states to offline on process failure,
ensures that, on process coming back online, all the devices can be
updated from offline to online except failed ones.

### DIFF
--- a/hax/hax/motr/__init__.py
+++ b/hax/hax/motr/__init__.py
@@ -386,7 +386,7 @@ class Motr:
                   proc_fid, len(disk_list), int(new_state), disk_list)
         if disk_list:
             state = (ServiceHealth.OK if new_state ==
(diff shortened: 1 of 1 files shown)


mssawant:
@Shreya-18 , please elaborate the problem and the solution in the commit message.

Shreya-18:
> @Shreya-18 , please elaborate the problem and the solution in the commit message.

**Updated** the commit message. I raised the PR for custom build and testing purpose so missed it before.

mssawant (approved):

suvratjoshi (approved):

PR #1710, opened 2021-07-17
Problem: allowed enclosure faillures is not populated for dix/md pools

mssawant:
The default allowed enclosure failures for dix/md pools is not populated which
may cause inconsistencies while wrting meta data further causing IO failures
in case of enclosure or node failures.

Solution:
Calculate and populate allowed enclosure failures for dix and md pools using
number of enclosures in the cluster.


madhavemuri (approved):
Looks fine to me.

SwapnilGaonkar7 (approved):

PR #1714, opened 2021-07-22
EOS-23105 Add skeleton class for integration of Hare with HA Event Manager

SwapnilGaonkar7:
HA is adding event manager class which can be used to subscribe/register and listen for events.
Hare needs to add class to access this event manager for listen events
In this PR, we are adding skeleton class to communicate with Event Manager
Note: Event Manager class is not fully implemented yet, just skeleton class is present right now

### DIFF
--- a/utils/manage_events.py
+++ b/utils/manage_events.py
@@ -1,0 +1,139 @@
+# Copyright (c) 2021 Seagate Technology LLC and/or its Affiliates
+#
+# This program is free software: you can redistribute it and/or modify it
+# under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License,
+# or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License
+# for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing, please email
+# opensource@seagate.com or cortx-questions@seagate.com.
+
+
+import logging
+import json
+from typing import List, Optional
+from dataclasses import dataclass
+
+from cortx.utils.message_bus import MessageBus, MessageConsumer
+from ha.core.event_manager.event_manager import EventManager
+from ha.core.event_manager.subscribe_event import SubscribeEvent
+
+COMPONENT_ID = 'hare'
+
+@dataclass
+class Event:
+    version: str
+    event_type: str
+    event_id: str
+    resource_type: str
+    cluster_id: str
+    site_id: str
+    rack_id: str
+    storageset_id: str
+    node_id: str
+    resource_id: str
+    timestamp: str
+
+
+class EventListener():
+
+    """
+    This is listener for event manager
+
+    Typical use case of this class is as follows
+    subscribe(event list) - subscribe to events for listening, returns topic
+                            which can be used for listening
+    _listen(topic) - reading messages received on above topic
+    unsubscribe(event list) - unsubscribe as we dont want to receive these
+                              events anymore
+    get_message_type() - returns topic just in case user is not maintaining
+                         output of 'subscribe'
+    Note: 1. There is only 1 topic(message type, queue) for given component.
+          2. When component subcribes for event for the 1st time topic is
+             created and returned to user.
+          3. This topic will be used by user to listen for events.
+          4. After event is received in queue, user will read it, process it
+             and acknowledge it
+    """
+    event_manager = None
+    consumer = None
+    topic = None
+
+    def __init__(self, event_list: List[SubscribeEvent]):
+        '''
+
+        We will subscribe to events passes as 'event_list'.
+        We will also create consumer for events
+        '''
+        logging.debug('Inside EventListener')
+        self.event_manager = EventManager.get_instance()
+
+        topic = self._subscribe(event_list)
+        if topic is None:
+           raise RuntimeError('Failed to subscribe to events')
+
+        message_bus = MessageBus()
+        self.consumer = MessageConsumer(message_bus=message_bus,
+                                        consumer_id=COMPONENT_ID,
+                                        consumer_group=COMPONENT_ID,
+                                        message_types=[topic],
+                                        auto_ack=False,
+                                        offset='earliest')
+
+    def _subscribe(self, event_list: List[SubscribeEvent]) -> str:
+        logging.info('Subscribing to events: %s', event_list)
+        kafka_topic_name = self.event_manager.subscribe(COMPONENT_ID,
+                                                        event_list)
+        return kafka_topic_name
+
+    def unsubscribe(self, event_list: List[str]) -> bool:
+        logging.info('Unsubscribing for events: %s', event_list)
+        return self.event_manager.unsubscribe(COMPONENT_ID, event_list)
+
+    def get_next_message(self, time_out) -> Optional[Event]:
+        '''
+        Listen for events from event manager
+        Once user gets message using below command user needs to call ack() to
+        acknowledge that message is already processed
+        '''
+        logging.info('Getting next message from topic: (%s)', self.topic)
+
+        try:
+            logging.info('Listening......')
+            message = self.consumer.receive(time_out)
+            if message is not None:
+                return self._parse(message)
+            else:
+                return None
+        except Exception as e:
+            logging.error('Exception occured in get_next_message (%s)', e)
+
+    def _parse(self, message: str) -> Event:
+        data = json.loads(message.decode('utf-8'))
+        return Event(version=data['version'],
+                     event_type=data['event_type'],
+                     event_id=data['event_id'],
+                     resource_type=data['resource_type'],
+                     cluster_id=data['cluster_id'],
+                     site_id=data['site_id'],
+                     rack_id=data['rack_id'],
+                     storageset_id=data['storageset_id'],
+                     node_id=data['node_id'],
+                     resource_id=data['resource_id'],
+                     timestamp=data['timestamp'])
+
+    def ack(self):
+        '''
+        1. This method will commit last read message offset for confirming
+           which messages the consumer has already processed
+        2. Consumer will read message using 'listen' and will acknowledge
+           that message using ack method
+        '''
+        self.consumer.ack()


suvratjoshi:
@SwapnilGaonkar7 
Please add appropriate commit message and PR description.
Thanks!

SwapnilGaonkar7:
@suvratjoshi @mssawant Please check. Added description

utils/manage_events.py
```diff
@@ -0,0 +1,87 @@
...
+    get_message_type() - returns topic just in case you are not maintaining
+                         output of subscribe
+    """
+
+    event_manager = None
+
+    def __init__(self):
+        print('Inside __init__')
+        self.event_manager = EventManager.get_instance()
+
```
suvratjoshi, line 46: Would be nice to have some detailed documentation, usage : input/output as a header for all functions here.
Thank you!

suvratjoshi (approved):

utils/manage_events.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
```
knekrasov, line 1: Shebang is not needed.

utils/manage_events.py
```diff
@@ -0,0 +1,87 @@
...
+
+import logging
+from typing import List
+
+
+from cortx.utils.message_bus import MessageBus, MessageConsumer
+from ha.core.event_manager.event_manager import EventManager
+
+
+class ManageEvents():
```
knekrasov, line 29: Manage Events sounds like an action "Go and manage events!".

Suggestions:
* EventManagerAdapter
* EventManagerConnector
* EventManager
* HAEventListener
* EventListener (I would vote for this one, especially if it is placed into `hax.ha` package)

utils/manage_events.py
```diff
@@ -0,0 +1,87 @@
...
+    unsubscribe(event list) - unsubscribe as we dont want to receive these
+                              events anymore
+    get_message_type() - returns topic just in case you are not maintaining
+                         output of subscribe
+    """
+
+    event_manager = None
+
+    def __init__(self):
+        print('Inside __init__')
```
knekrasov, line 44: `logging.debug`, please. No printing - that's a rule of a thumb.

utils/manage_events.py
```diff
@@ -0,0 +1,87 @@
...
+        self.event_manager = EventManager.get_instance()
+
+    def subscribe(self, event_list: List[str]) -> str:
+        logging.info('Subscribing for events: (%s)', str(event_list)[1:-1])
+        kafka_topic_name = self.event_manager.subscribe('hare', event_list)
```
knekrasov, line 49: `'hare'` could have been refactored into a constant or a field or a method invocation. Hardcoded strings (especially when they are repeated multiple times) is not a good thing.

utils/manage_events.py
```diff
@@ -0,0 +1,87 @@
...
+        return kafka_topic_name
+
+    # Gets message type
+    def get_message_type(self) -> str:
+        print('get_message_type')
```
knekrasov, line 54: `logging` please

utils/manage_events.py
```diff
@@ -0,0 +1,87 @@
...
+        return self.event_manager.message_type('hare')
```
knekrasov, line 55: What does this method actually do?
Is `'hare'` component allowed to listen to ONE message type only? (the method uses singular word)

utils/manage_events.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # Get list of events subscribed by Hare
+    def list(self) -> List[str]:
```
knekrasov, line 58: * list is not the best name for the method (we can refrain `get_events` - it is clearer from the reader's perspective)
* What does it return? In other words, what are there in these strings? Can we parse those raw values into something meaningful?
* What is the semantics of this call?
   * I understand that this fetches some events from Kafka and return to the client. But what happens to the queue then? Are those messages marked as 'read' by the result of this code? If not, what the user of this class needs to do to make sure that the events are not read twice?

Update: after looking down into `_listen()` method, I stopped to understand what this `list()` method is intended to do :-) perhaps some explanation (like vocabulary - what is the type, what is the event, what do we listen for etc) is needed.

utils/manage_events.py
SwapnilGaonkar7: I have added some more description for this method.
Yes, hare component will listen to ONE topic only.
`Event manager creates message type (queue/topic) in the message bus based on component name when component subscribes the event for the first time, that topic name is being used by the event manager to publish event to the message name. `

utils/manage_events.py
SwapnilGaonkar7: Changes method name to 'get_events'
Will add some more info about this method/class

utils/manage_events.py
SwapnilGaonkar7: Done, also added some more info in class header

utils/manage_events.py
SwapnilGaonkar7: Done

utils/manage_events.py
SwapnilGaonkar7: Done, EventListener

utils/manage_events.py
SwapnilGaonkar7: Done

utils/manage_events.py
SwapnilGaonkar7: Done

utils/manage_events.py
SwapnilGaonkar7: Done

utils/manage_events.py
```diff
@@ -0,0 +1,87 @@
...
+        print('Get list')
+        return self.event_manager.get_events('hare')
+
+    def unsubscribe(self, event_list: List[str]) -> bool:
+        logging.info('Unsubscribing for events: (%s)', str(event_list)[1:-1])
+        return self.event_manager.unsubscribe('hare', event_list)
+
+    # Below method needs to be implemented in consumer of this class(??)
```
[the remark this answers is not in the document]
SwapnilGaonkar7: Modified the function.
Please check latest version, also added some info text

SwapnilGaonkar7:
@knekrasov @suvratjoshi I have addressed review comments.
Please have a look.

utils/manage_events.py
```diff
@@ -0,0 +1,103 @@
...
+
+    event_manager = None
+    consumer = None
+
+    def __init__(self):
+        logging.debug('Inside EventListener')
+        self.event_manager = EventManager.get_instance()
+
+    def subscribe(self, event_list: List[str]) -> str:
+        logging.info('Subscribing for events: (%s)', str(event_list)[1:-1])
```
knekrasov, line 55: Why do you make such a slice of a string?

utils/manage_events.py
```diff
@@ -0,0 +1,103 @@
...
+        kafka_topic_name = self.event_manager.subscribe(COMPONENT_ID, event_list)
+        return kafka_topic_name
+
+    # Gets message type name mapped with component name
```
knekrasov, line 59: The comment could be a docstring. Please change them so they become visible in the IDE and Python-friendly editors.

utils/manage_events.py
```diff
@@ -0,0 +1,103 @@
...
+    # Event manager creates message type (queue/topic) when component
+    # subscribes the event for the first time
+    def get_message_type(self) -> str:
```
knekrasov, line 62: Just curious - why will the developer want to get the message type? in other words, what is the use case for this value?

Can you add this piece of information to the docstring also?

utils/manage_events.py
```diff
@@ -0,0 +1,103 @@
...
+        return self.event_manager.get_events(COMPONENT_ID)
+
+    def unsubscribe(self, event_list: List[str]) -> bool:
+        logging.info('Unsubscribing for events: (%s)', str(event_list)[1:-1])
+        return self.event_manager.unsubscribe(COMPONENT_ID, event_list)
+
+    # Listen for events from event manager
+    # Once user gets message using below command user needs to call ack() to
+    # acknowledge that message is already processed
+    def listen(self, topic: str, time_out):
```
knekrasov, line 78: According to the code, the method doesn't listen but just takes at least one message. And doesn't return it BTW.

Normally, I would expect that  `listen` method is something blocking (so it will continuously fetch the data and hand it to the callback). In the way it is implemented a more descriptive name would be something like `get_next_message() -> Optional[MessageType]` (it returns not more than 1 message).

BTW what will happen if the user will invoke this methods twice in a row but will not invoke `ack()`? Will it return one and the same value?

utils/manage_events.py
```diff
@@ -0,0 +1,103 @@
...
+        logging.info('Listening on topic: (%s)', topic)
+        if self.consumer is None:
+            message_bus = MessageBus()
+            self.consumer = MessageConsumer(message_bus=message_bus,
+                                            consumer_id=COMPONENT_ID,
+                                            consumer_group=COMPONENT_ID,
+                                            message_types=[topic],
+                                            auto_ack=False,
+                                            offset='earliest')
```
knekrasov, line 87: I'm a bit confused that this consumer is created every time the user invokes this `listen` method.

utils/manage_events.py
SwapnilGaonkar7: It is to make message simpler
`Subscribing for events: ('bgdelete')`
Otherwise it will look like
`Subscribing for events: (['bgdelete'])`

utils/manage_events.py
SwapnilGaonkar7: Yes, this function will return message but we still don't know what will be type(format) of message.
We are implementing skeleton that's why I haven't added return message.
During implementation of [EOS-22224](https://jts.seagate.com/browse/EOS-22224) I am planning to add proper decode and return logic.

`BTW what will happen if the user will invoke this methods twice in a row but will not invoke ack()? Will it return one and the same value?
`
That depends on event manager implementation and I was planning to add more details about it while implementing above task. 
Sounds good?

utils/manage_events.py
SwapnilGaonkar7: Done

utils/manage_events.py
SwapnilGaonkar7: No, it will be created only once.
Below check will make sure that same consumer will be used 2nd time onwards.
`if self.consumer is None:`

utils/manage_events.py
knekrasov: I doubt that it makes sense. The readability is rather bad while the result is hardly beneficial.

Suggestion: 
```suggestion
        logging.info('Subscribing to events: %s', event_list)
```

Just FYI (I don't force you to apply this in this patch; this code will probably be changed in the future anyway):

The fact is that we use patterns in logging (e.g. `logging.debug('My value: %s', some_function())` to make the evaluations lazy. So in this example `some_function()` will be evaluated if and only if the log lever requires this log message to be rendered. If the log level is INFO or WARN, this unnecessary evaluation will even not happen.

Another point is that internally `%s` causes `str()` to happen just automatically. And that `str()` will also be lazy.
It means that `logging.info('Subscribing for events: (%s)', event_list)` will caluse `str(event_list)` lazy evaluation, but `logging.info('Subscribing for events: (%s)', str(event_list))` will force `str()` even if the log message is not required.

So it is just a good habit - keep logging lazy when possible.

utils/manage_events.py
knekrasov: Sorry, you're right.

utils/manage_events.py
knekrasov: > Yes, this function will return message but we still don't know what will be type(format) of message.

I see, but then it's a good reason to force MessageBus creators (or cortx-ha team?) to provide this information 😉 

Suggestion:
1. Add a stub class that would store the parsed message - e.g. `BaseEvent`. Nowadays it can be empty but it already can have FIXME comment. That class can have some fields in the future and it also can be inherited (e.g. different subclasses can represent different message types that our application can work with).
2. Implement a stub parsing method that takes a raw message string and returns `BaseEvent`. Nowadays its implementation can be no-op or something dummy. There should be just FIXME or TODO comment.

Why this is worth doing:
1. This approach defines the contracts (because we'll use types instead of raw strings; other parts of application will depend on those types and can freely access the right keys without any fear to get None or KeyError or something). When (by any reason) the format of the messages changes, that change will be reflected at the level of type. Other parts of our application will be validated with mypy and thus if we access some non-existent fields (the message structure changed), then we'll get build error. So the mistake will not be hidden.
2. Our parsing method, stub type and TODO/FIXME comments define our intention - so the way how this code should evolve will be defined. So although there are many things to be clarified at MessageBus or EventManager sides, we know how this code should look like in the future. Even if the code will be polished by another developer.

cc @vaibhavparatwar, @mssawant, @suvratjoshi

knekrasov (changes requested):
I think that overall interface (i.e. list of functions that `EventListener` exposes for external use) should be way shorter and easier to use. Currently it mostly repeats the methods that `EventManager` has. `EventManager`'s API is definitely not perfect but we still need to find a way to simplify the developer's life (e.g. make the contracts clearer - see my comment on message structure and types). The less methods we have - the less possibility to use them in a right way.

As far as I understand, the use cases for `EventListener` is as follows:
1. I know the name of the message type and I want to start receiving such messages
2. I want to know if a new message of the given type was published
(@SwapnilGaonkar7, @mssawant if there are more use cases - please share)

Having this list in mind, I would anticipate the API like this:

```python
# variant 1: API with explicit acks
listener = EventRetriever('my-message-type-name')
if listener.has_new_events():
    msg = listener.retrieve()
    process_message(msg)  # this is not part of the API, just some external function
    listener.ack()

# variant 2: API with callback. The convention is that if callback finishes without an error, then the message is ack'ed. 
# Otherwise no ack happens. Exception also causes internal listener loop to finish (in other words, uncaught exception will cause `listen` method to return.

def handler(msg: BaseEvent) -> None:
    if _not_happy(msg):
        raise RuntimeError('Bad mood, I can't process message ' + msg)
    process_message(msg)  # this is not part of the API, just some external function

listener = EventListener('my-message-type')
listener.listen(handler) # blocking call
```
At this point I can't tell whether variant 1 is better than variant 2. Probably the first variant is a bit more flexible (so that we can listen to multiple event types simultaneously and/or build generator functions on the top of this approach). But any of the approaches is shorter than proposed in this PR.

utils/manage_events.py
SwapnilGaonkar7: Done

utils/manage_events.py
SwapnilGaonkar7: @knekrasov  Changes name to 'get_next_message()', this function will return 1 message at a time.

utils/manage_events.py
knekrasov: I doubt that we really need this method

utils/manage_events.py
```diff
@@ -0,0 +1,114 @@
...
+        '''Get list of events subscribed by Hare'''
+
+        logging.debug('Get list')
+        return self.event_manager.get_events(COMPONENT_ID)
+
+    def unsubscribe(self, event_list: List[str]) -> bool:
+        logging.info('Unsubscribing for events: %s', event_list)
+        return self.event_manager.unsubscribe(COMPONENT_ID, event_list)
+
+    def get_next_message(self, topic: str, time_out) -> Optional[Dict]:
```
knekrasov, line 81: Let's use more specific type instead of `Dict`. 

`Dict` has no commitments on which keys it has and what type do they have. Let's create a custom type (dataclass) that will hold the parsed message. And please add `_parse()` method that translates the raw string into that well-typed class.

utils/manage_events.py
```diff
@@ -0,0 +1,114 @@
...
+
+        logging.info('Getting next message from topic: (%s)', topic)
+        if self.consumer is None:
+            message_bus = MessageBus()
+            self.consumer = MessageConsumer(message_bus=message_bus,
+                                            consumer_id=COMPONENT_ID,
+                                            consumer_group=COMPONENT_ID,
+                                            message_types=[topic],
+                                            auto_ack=False,
+                                            offset='earliest')
```
knekrasov, line 96: This could go to `__init__` method.

utils/manage_events.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+
+    event_manager = None
+    consumer = None
+
+    def __init__(self):
+        logging.debug('Inside EventListener')
+        self.event_manager = EventManager.get_instance()
+
+    def subscribe(self, event_list: List[SubscribeEvent]) -> str:
```
knekrasov, line 56: I really doubt that `subscribe`/`unsubscribe` methods are really needed.

Internally, your EventListener can potentially use them, but I see no point to expose these two methods to the client. The client just wants to start receiving the events sent to hare.

Can this listener class check somewhere at the very beginning if hare is registered to receive such events and if not, automatically subscribe on them?

utils/manage_events.py
```diff
@@ -0,0 +1,114 @@
...
+        '''Gets message type name mapped with component name
+        Event manager creates message type (queue/topic) when component
+        subscribes the event for the first time. User can use this message type
+        or topic for listening events. Note it is same topic which user will
+        receive after calling subscribe method'''
+
+        logging.debug('get_message_type')
+        return self.event_manager.message_type(COMPONENT_ID)
+
+    def get_events(self) -> List[str]:
```
knekrasov, line 71: Do we really need this method?

utils/manage_events.py
```diff
@@ -0,0 +1,114 @@
...
+                                            consumer_group=COMPONENT_ID,
+                                            message_types=[topic],
+                                            auto_ack=False,
+                                            offset='earliest')
+
+        try:
+            logging.info('Listening......')
+            message = self.consumer.receive(time_out)
+            if message is not None:
+                return json.loads(message.decode('utf-8'))
```
knekrasov, line 102: This could go to `_parse` method.

utils/manage_events.py
SwapnilGaonkar7: @knekrasov  Client needs to provide list of events which it wants to receive(subscribe).
How client should communicate them if we remove subscribe/unsubscribe?

knekrasov: The client can provide this list when creating EventListener (i.e. as `__init__` parameter).

And one more point: as far as I understand, subscribe/unsubscribe is cluster-wise action. In other words, they either subscribe ALL hare instances in the cluster to an event or the unsubscribe ALL hare instances. Is it correct?

If this is correct, then unsubscribing must be done with care: listening is done at every particular node and listen is node-specific. If one of the nodes stops listening, it should not unsubscribe because it is possible that there are other hare instances that are still willing to receive notifications.

utils/manage_events.py
SwapnilGaonkar7: Done

utils/manage_events.py
SwapnilGaonkar7: Done

utils/manage_events.py
SwapnilGaonkar7: Done

utils/manage_events.py
SwapnilGaonkar7: Done

utils/manage_events.py
SwapnilGaonkar7: Done

knekrasov (approved):
Please resolve Codacy complaints. Thanks

PR #1716, opened 2021-07-23
EOS-23123: hax transitions motr drives to M0_NC_FAILED instead of M0_NC_OFFLINE

mssawant:
On node failure, hax notifies M0_NC_FAILED for all the motr processes on that node to all
the peer motr processes. Along with this the corresponding drives to the ioservices are also
notified as M0_NC_FAILED. But motr does not allow transition to M0_NC_ONLINE from M0_NC_FAILED.
A FAILED motr device, that is notified M0_NC_FAILED in case of permanent failure must go
through M0_NC_FAILED -> M0_NC_REPAIRING -> M0_NC_REPAIRED -> M0_NC_REBALANCING ->
M0_NC_ONLINE to become online. Thus a direct transition from M0_NC_FAILED -> M0_NC_ONLINE is
not allowed. Thus, hax needs to notify M0_NC_OFFLINE instead of M0_NC_FAILED for the ioservice
devices in case of node/process failure, this is also interpreted as transient failure for
devices.

Solution:
Notify M0_NC_OFFLINE and not M0_NC_FAILED for devices corresponding to the ioservice failure.

### DIFF
--- a/hax/hax/motr/__init__.py
+++ b/hax/hax/motr/__init__.py
@@ -380,8 +380,6 @@ class Motr:
         proc_fid = Fid.from_struct(note.no_id)
         for svc in services:
             disk_list += cns.get_disks_by_parent_process(proc_fid, svc.fid)
-        LOG.debug('proc fid=%s encloses %d disks with state %d as follows: %s',
-                  proc_fid, len(disk_list), int(new_state), disk_list)
         if disk_list:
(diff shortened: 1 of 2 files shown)


SwapnilGaonkar7 (approved):
Looks good

vaibhavparatwar (approved):

suvratjoshi (approved):

PR #1725, opened 2021-07-28
EOS-22484: RC can't be re-elected after hctl shutdown

knekrasov:
Jira issue: [EOS-22484](https://jts.seagate.com/browse/EOS-22484)

As long as Consul is not stopped anymore during hctl shutdown, Consul
Session may be kept alive even if Hare cluster is down. The Consul
Session mechanism is used for RC leader election. In the described
situation in newly started Hare cluster there is no Hax process which
can consider itself as running at RC leader. In particular, this means
that there is no Hax process that would keep updating FS statistics in Consul KV.

When Consul sessions are destructed?

1. When Consul cluster is restarted. It was the case before --skip-consul-stop flag was added.
2. When any the checks registered as a part of contract of this Session start failing.
In case of RC leader session the contract is the health state of confd and hax processes.
NOTE: only "critical" state is a reason for session destruction

Consul monitors the health state of Hare component processes by means of
check-service script. It is written in a way so that if a systemd service
was stopped gracefully, its status will be "warning" (but not "critical").

## Solution:
When hax starts, it should look for a stale RC leader session and
destroy it.

Signed-off-by: Konstantin Nekrasov <konstantin.nekrasov@seagate.com>

### DIFF
--- a/hax/hax/hax.py
+++ b/hax/hax/hax.py
@@ -19,6 +19,7 @@
 #
 
 import logging
+import re
 from typing import List, NamedTuple
 
 from hax.filestats import FsStatsUpdater
@@ -57,6 +58,29 @@ def _run_stats_updater_thread(motr: Motr,
 
 
 @repeat_if_fails()
+def _remove_stale_session(util: ConsulUtil) -> None:
+    '''Destroys a stale RC leader session if it exists or does nothing otherwise.
+
+    An RC leader session may survive 'hctl shutdown'. In such a case 'leader'
+    key will contain a garbage value 'elect2805' but the session will be alive
+    and thus RC leader will not be re-elected.
+    '''
+    if not util.kv.kv_get_raw('leader'):
+        # No leader key means that there can be no stale RC leader session for
+        # sure. We're starting for the first time against a fresh Consul KV.
+        return
+
+    sess = util.get_leader_session_no_wait()
+    node = util.get_leader_node()
+    if re.match(r'^elect[\d]+$', node):
+        LOG.debug(
+            'Stale leader session found: RC leader session %s is '
+            'found while the leader node seems to be stub: %s', sess, node)
+        util.destroy_session(sess)
+        LOG.debug('Stale session %s destroyed to enable RC re-election', sess)
+
+
+@repeat_if_fails()
 def _get_motr_fids(util: ConsulUtil) -> HL_Fids:
     hax_ep: str = util.get_hax_endpoint()
     hax_fid: Fid = util.get_hax_fid()
@@ -91,6 +115,7 @@ def main():
     planner = WorkPlanner()
 
     util: ConsulUtil = ConsulUtil()
+    _remove_stale_session(util)
     cfg: HL_Fids = _get_motr_fids(util)
 
     LOG.info('Welcome to HaX')
--- a/hax/hax/util.py
+++ b/hax/hax/util.py
@@ -390,6 +390,17 @@ class ConsulUtil:
             raise HAConsistencyException(
                 'Could not get the leader from Consul')
 
+    def destroy_session(self, session: str) -> None:
+        """
+        Destroys the given Consul Session by name.
+        The method doesn't raise any exception if the session doesn't exist.
+        """
+        try:
+            self.cns.session.destroy(session)
+        except (ConsulException, HTTPError, RequestException) as e:
+            raise HAConsistencyException('Failed to communicate to'
+                                         ' Consul Agent: ' + str(e))
+
     def get_session_node(self, session_id: str) -> str:
         try:
             session = self.cns.session.info(session_id)[1]
--- a/stubs/consul/__init__.pyi
+++ b/stubs/consul/__init__.pyi
@@ -111,6 +111,10 @@ class Session:
         consistency: str = None,
         dc: str = None,
     ) -> Tuple[int, Any]: ...
+    def destroy(
+        self,
+        session_id: str
+    ) -> bool: ...
 
 class ConsulException(Exception): ...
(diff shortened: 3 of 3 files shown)


hax/hax/hax.py
```diff
@@ -56,6 +57,30 @@ def _run_stats_updater_thread(motr: Motr,
...
+    and thus RC leader will not be re-elected.
+    """
+    if not util.kv.kv_get_raw('leader'):
+        # No leader key means that there can be no stale RC leader session for
+        # sure. We're starting for the first time against a fresh Consul KV.
+        return
+
+    sess = util.get_leader_session_no_wait()
+    node = util.get_leader_node()
+    if re.match(r'^elect[\d]+$', node):
```
andriytk, line 76: If some node hostname will be `elect123` or something - it would match this regexp and would cause session destructions on each `hax` startup on this node. Why not check explicitly on the magic `elect2805` value?

knekrasov: Because any `^elect[\d]+$` is considered to be a stub. 

https://github.com/Seagate/cortx-hare/blob/ce3411e975b4cb93f2e456e1c5f3832191535844/utils/hare-bootstrap#L112-L123

In other words, whole bootstrapping and RC re-election will probably be broken if some of the hostnames match that regexp :-) And this doesn't seem to be the problem of this PR.

andriytk: No, it won't be broken in case of bootstrapping via hctl, because we are checking for the session, not the hostname value there.

knekrasov: Anyway we have other parts of the code that do have assumption that `elect[\d]+` means "No leader elected". If we don't have that assumption anymore, then other parts of the code should be adjusted.

mssawant:
@knekrasov , kindly address the codacy issue regarding docstring.

mssawant (approved):
Looks good, kindly address the codacy issue.

suvratjoshi (approved):

knekrasov:
> kindly address the codacy issue regarding docstring.

@mssawant Sorry, it seems like it is impossible. Codacy configuration seems to be inconsistent. When I fix D212 warning, D213 is triggered and vice versa. 

Rules D212 and D213 are mutually exclusive, but both of them seem to be enabled. This is similar to https://github.com/PyCQA/pydocstyle/issues/475

Just please have a look at these screenshots:

image

image

hax/hax/hax.py
andriytk: Probably. I'm not aware of the other parts, but I don't think that it's right to use it here as a pattern instead of a concrete value.

knekrasov: `elect2805` gets written from hare-shutdown by your suggestion (we need to put "fake" leader node there to enable RC re-election to happen when the hare cluster is restarted again). `elect2805` is chosen as a particular value that corresponds to the code below:
https://github.com/Seagate/cortx-hare/blob/ce3411e975b4cb93f2e456e1c5f3832191535844/utils/hare-bootstrap#L112-L123
(I guess you are the author)

Regarding other parts. This is another example where this pattern is already used:
https://github.com/Seagate/cortx-hare/blob/ce3411e975b4cb93f2e456e1c5f3832191535844/utils/utils.py#L167-L168

BTW this function appeared quite long ago: https://github.com/Seagate/cortx-hare/pull/728

Long story short, 
1. there is (perhaps implicit) assumption that sometimes `leader` contains the text which is NOT a hostname. It obviously means that not all hostnames are suitable for Hare cluster, but that decision seems to have been made at the very beginning of Hare history.
2. In this case the value is described by `^elect[\d]+$` regexp.
3. Because of \[1] some values will always be considered as "stub" instead of "real hostname", whatever regexp to describe it we use.

That's why I suggest to move this discussion out of this PR. If you think there is a way to consider that any value in `leader` always means hostname, let's create a ticket and address it separately.

vaibhavparatwar:
Deployment on 3N VM passed, verified the JIRA and updated JIRA with detailed testing logs @mssawant @knekrasov

vaibhavparatwar:
@mssawant @knekrasov - JIRA logged against RE team - https://jts.seagate.com/browse/EOS-23252 for codacy issue.

hax/hax/hax.py
andriytk: Let me try to explain it again: the code at `hare-bootstrap` does not assume "that sometimes leader contains the text which is NOT a hostname". This code just writes some value to leader in order to trigger the RC leader election mechanism, but it does not check the value. So again - it does not assume anything about it.

The problem is with the code, which indeed makes such assumptions - I have never written it and was not aware about it. In any case, we better don't extend such assumptions in our code - it's wrong and dangerous.

mssawant: I am okay with having a specific magic value to represent an invalid session. That sounds reasonable.
I think `consul kv put leader elect$RANDOM`  triggered the pattern matching which is also done in some other places.
Now, to trigger RC leader during `hctl bootstrap` we need to change the leader value and having 1 magic value will not work as it will not trigger the change in the leader KV. I think we can keep `consul kv put leader elect$RANDOM` as its the bootstrap path, in case of failure to bootstrap the cluster won't be usable anyway.
The problem mainly lies when the services are stopped and started outside `hctl bootstrap`, mainly through conrtx-ha (via pacemaker). Presently as part of hctl shutdown we are resetting the leader KV, here we can use a defined magic for resetting.
@knekrasov, I wonder if we have really solved the problem as we are destroying the session but will it really re-trigger RC leader election in case of hax stop - start sequences because we are resetting the leader KV in `hare-shutdown` but not during hax shutdown.

PR #1726, opened 2021-07-28
EOS-22149: Hare system tests - Add Bootstrap-shutdown test case to IVT framework

pavankrishnat:
Added Bootstrap-Shutdown test case to IVT framework.

CMD to run Bootstrap-Shutdown test in loop for 10 times.
`/opt/seagate/cortx/hare/bin/hare_setup test --config 'json:///opt/seagate/cortx_configs/provisioner_cluster.json' --file /var/lib/hare/cluster.yaml --plan sanity --dev 10`

Test will run only if `--dev` argument is used.

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -259,132 +259,6 @@ def test(args):
         exit(-1)
 
 
-def test_hare_prereq():
-    """Test suite for setting the test environment before start of test."""
-    pcs_status = ['pcs', 'status']
-    cluster_stop = ['cortx', 'cluster', 'stop']
-
-    logging.info('Test started on Host: {}'.format(execute(['hostname'])))
-    logging.info('Check that all services are up in PCS.')
-
-    resp = execute(pcs_status)
-    logging.info('PCS status: %s', resp)
-
-    if 'cluster is not currently' in resp:
-        return
-
-    logging.info('Make Node ready for testing, by stopping the cluster.')
-    resp = execute(cluster_stop)
-    logging.info('Cluster Stopped: %s', resp)
-    if 'Cluster stop is in progress' not in resp:
-        logging.error('Cluster is in progress.')
-
-
-def test_hare_postreq(cdf_file: str, timeinfo: str, logfile: str):
-    """Test suite is for restoring the setting after test."""
-    pcs_status = ['pcs', 'status']
-    cluster_start = ['cortx', 'cluster', 'start']
-
-    logging.info('Start the Cluster')
-    resp = execute(cluster_start)
-    logging.info('cluster status: %s', resp)
-    if 'Cluster start operation performed' not in resp:
-        logging.error('Cluster not yet started.')
-
-    logging.info('PCS: Check all services are up.')
-    resp = execute(pcs_status)
-    logging.info('PCS status: %s', resp)
-    if 'stopped' not in resp:
-        logging.error('Some services are not up.')
-
-    logging.info('hctl: Check that all the services are started.')
-    cluster_sts = check_cluster_status(cdf_file)
-    if cluster_sts:
-        resp = execute(['journalctl', '--since', timeinfo, '>', logfile])
-        logging.info('created journal log: %s', logfile)
-        raise Exception('hctl status reports failure.')
-
-    logging.info('Successfully performed cleanup after testing')
-
-
-def _hctlstatus():
-    """know the hctl status."""
-    hctl_status = ['hctl', 'status', '-d']
-    hctl_shutdown = ['hctl', 'shutdown', '--skip-consul-stop']
-
-    process = subprocess.Popen(hctl_status,
-                               stdin=subprocess.PIPE,
-                               stdout=subprocess.PIPE,
-                               stderr=subprocess.PIPE,
-                               encoding='utf8')
-    logging.debug('Issuing CLI command: %s', hctl_status)
-    out, err = process.communicate()
-    exit_code = process.returncode
-    logging.debug('Finished. Exit code: %d', exit_code)
-    if exit_code:
-        if 'Cluster is not running' not in out:
-            resp = execute(hctl_shutdown)
-            logging.info('hctl shutdown: %s', resp)
-            if is_cluster_running():
-                raise Exception('After shutdown, hctl is running.')
-        else:
-            raise Exception(
-                f'Command hctl status exited with error code {exit_code}'
-                f'Command output: {err}')
-    else:
-        raise Exception('Fail to control cluster, exit the test.')
-
-
-def test_hare_bootstrap_shutdown(args):
-    """Test suite for single node hare init in loop."""
-    loop_count = int(args.dev[0])
-    hctl_shutdown = ['hctl', 'shutdown', '--skip-consul-stop']
-
-    test_hare_prereq()
-
-    if is_cluster_running():
-        _hctlstatus()
-
-    logging.info('-------Starting BOOTSTRAP-SHUTDOWN in LOOP-------')
-    for count in range(loop_count):
-        logging.info('Loop count# {}'.format(count + 1))
-        now = datetime.now()     # current date and time
-        date_time = now.strftime("%Y-%m-%d %H:%M:%S")
-        jlog = '~/journal_ctrl_' + now.strftime('%Y_%m_%d_%H%M%S') + '.log'
-
-        logging.info('Start hctl Bootstrap')
-        resp = bootstrap_cluster(str(args.file[0]), True)
-        if resp:
-            logging.error('Failed to bootstrap')
-            resp = execute(['journalctl', '--since', date_time, '>', jlog])
-            logging.info('created journal log: %s', jlog)
-            exit(-1)
-
-        logging.info('Check that all the services are up in hctl.')
-        if is_cluster_running():
-            logging.info('hctl is running.')
-            cluster_sts = check_cluster_status(str(args.file[0]))
-            if cluster_sts:
-                resp = execute(
-                    ['journalctl', '--since', date_time, '>', jlog])
-                logging.info('created journal log: %s', jlog)
-                raise Exception('hctl status reports failure.')
-        else:
-            resp = execute(['journalctl', '--since', date_time, '>', jlog])
-            logging.info('created journal log: %s', jlog)
-            raise Exception('After Bootstrap, hctl is not running.')
-
-        logging.info('Shutdown the cluster.')
-        resp = execute(hctl_shutdown)
-        logging.info('hctl shutdown: %s', resp)
-        if is_cluster_running():
-            resp = execute(['journalctl', '--since', date_time, '>', jlog])
-            logging.info('created journal log: %s', jlog)
-            raise Exception('After shutdown, hctl is running.')
-
-    test_hare_postreq(str(args.file[0]), date_time, jlog)
-
-
 def test_IVT(args):
     try:
         rc = 0
@@ -399,7 +273,8 @@ def test_IVT(args):
                               'running for executing tests')
                 exit(-1)
         else:
-            test_hare_bootstrap_shutdown(args)
+            subprocess.Popen(['pytest', '/root/devTest/eos22149/st/test/test_h'
+                                        'are_bootstrap_shutdown.py', 'args'])
 
         cluster_status = check_cluster_status(path_to_cdf)
         if cluster_status:
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -31,6 +31,7 @@ from enum import Enum
 from sys import exit
 from time import sleep
 from typing import Any, Callable, Dict, List
+from datetime import datetime
 
 import yaml
 from cortx.utils.product_features import unsupported_features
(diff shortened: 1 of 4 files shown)


provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
         exit(-1)
 
 
+def executecmds(cmd: List[str]) -> list:
```
knekrasov, line 258: Let it be just `execute`. Cmds is misleading (it executes just one command only).

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
...
+    process = subprocess.Popen(cmd,
+                               stdin=subprocess.PIPE,
+                               stdout=subprocess.PIPE,
+                               stderr=subprocess.PIPE,
+                               encoding='utf8')
+    resp = list(process.communicate())
+    resp.append(str(process.returncode))
+
+    return resp
```
knekrasov, line 267: The whole method could be looking like this:

```python
    def _execute(cmd: List[str]) -> str:
        process = Popen(cmd,
                        stdin=PIPE,
                        stdout=PIPE,
                        stderr=PIPE,
                        encoding='utf8')
        logging.debug('Issuing CLI command: %s', cmd)
        out, err = process.communicate()
        exit_code = process.returncode
        logging.debug('Finished. Exit code: %d', exit_code)
        if exit_code:
            raise CliException(out, err, exit_code)
        return out
```
So if everything is fine, it will just return the `stdout`. In case of non-zero exit code an exception is thrown (with stderr and exact exit code) - the exception type is custom.

Why do you want to return a list of stdout, stderr and exit code? This result value is not comfortable to handle.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
...
+
+    resp = executecmds(pcs_status)
+    logging.info('PCS status: %s', resp[0])
+
+    if 'cluster is not currently' in resp[0]:
+        return
+
+    logging.info('Make Node ready for testing, by stopping the cluster')
+    resp = executecmds(cluster_stop)
+    if int(resp[2]):
```
knekrasov, line 286: `resp[2]` looks too magical. Please refer my comment with `_execute` example.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
...
+        logging.info('Cluster status : %s', resp[0])
+        logging.error('Cluster failed to stop %s', resp[1])
+        raise Exception(
+            f'Command {cluster_stop} exited with error code '
+            f'{int(resp[2])}. Command output: {resp[1]}')
+    else:
+        logging.info('Cluster Stopped : %s', resp[0])
+        for line in resp[0]:
```
knekrasov, line 294: This deserves a separate function. And please think how to avoid `assert` instruction.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
...
+
+    cluster_sts = check_cluster_status(cdf_file)
+    if cluster_sts:
+        logging.error('Cluster status reports failure')
+        resp = executecmds(['journalctl', '--since', timeinfo, '>', logfile])
+        logging.info('created journal log.%s', logfile)
+        exit(-1)
+
+    logging.info('PCS: Check all services are up.')
+    sleep(20)
```
knekrasov, line 330: Sleep-driven development is error-prone.

FYI: in the past I implemented a tool called `pcswrap` that can be helpful for things like this (e.g. start all services and wait until they are started for sure - with timeout). Currently its source code is in cortx-ha repository: https://github.com/Seagate/cortx-ha/tree/main/ha/pcswrap

We could leverage either the whole pcswrap tool (TBD) or reuse some parts of its codebase. For instance, [Waiter](https://github.com/Seagate/cortx-ha/blob/19c6cad22b4a7f2036bc7e292b40383bae4d7fdb/ha/pcswrap/pcswrap/internal/waiter.py#L21) class can be beneficial:

https://github.com/Seagate/cortx-ha/blob/19c6cad22b4a7f2036bc7e292b40383bae4d7fdb/ha/pcswrap/pcswrap/client.py#L97-L103


@mssawant What's your opinion? Can we take back (or copy) pcswrap back to our repository?

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
...
+            f'Command {hctl_status} exited with error code {int(resp[2])}.'
+            f'Command output: {resp[1]}')
+    else:
+        for line in resp[0]:
+            assert 'stopped' not in line, 'Some services are not up.' \
+                if 'Some services are not up.' else line
+    logging.info('Successfully performed cleanup after testing')
+
+
+# @pytest.mark.sanity
```
knekrasov, line 358: Do you need this decorator? If not, please remove the line. Otherwise please uncomment it,

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
...
+            jlog = '~/journal_ctrl_' + now.strftime('%Y_%m_%d_%H%M%S') + '.log'
+
+            logging.info('Start hctl Bootstrap')
+            resp = bootstrap_cluster(str(args.file[0]), True)
+            if resp:
+                logging.error('Failed to bootstrap')
+                resp = executecmds(['journalctl', '--since', date_time, '>',
+                                    jlog])
+                logging.info('created journal log %s', jlog)
+                exit(-1)
```
knekrasov, line 393: Please don't use `exit()`. Especially if this function will become a test runnable by pytest.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
...
+            date_time = now.strftime("%Y-%m-%d %H:%M:%S")
+            jlog = '~/journal_ctrl_' + now.strftime('%Y_%m_%d_%H%M%S') + '.log'
+
+            logging.info('Start hctl Bootstrap')
+            resp = bootstrap_cluster(str(args.file[0]), True)
+            if resp:
+                logging.error('Failed to bootstrap')
+                resp = executecmds(['journalctl', '--since', date_time, '>',
+                                    jlog])
+                logging.info('created journal log %s', jlog)
```
knekrasov, line 392:
```suggestion
                logging.info('created journald log: %s', jlog)
```

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
...
+                exit(-1)
+
+            logging.info('Check that all the services are up in hctl.')
+            if is_cluster_running():
+                logging.info('hctl is running.')
+            else:
+                logging.error('Still hctl is not running.')
+                resp = executecmds(['journalctl', '--since', date_time, '>',
+                                    jlog])
+                logging.info('created journal log %s', jlog)
```
knekrasov, line 402:
```suggestion
                logging.info('created journald log: %s', jlog)
```

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
...
+                exit(-1)
```
knekrasov, line 403: Please don't use `exit()`

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
...
+
+            sleep(10)
```
knekrasov, line 405: Why do you need `sleep`?

provisioning/miniprov/hare_mp/main.py
```diff
@@ -565,6 +747,16 @@ def add_param_argument(parser):
     return parser
 
 
+def add_dev_argument(parser):
+    parser.add_argument('--dev',
+                        help='Test Development purpose. Supported '
+                        'values: any non zero value',
```
knekrasov, line 753: What does the number value mean? Is it just a workaround? If so, please consider looking for `store_true` in the [official tutorial](https://docs.python.org/3/howto/argparse.html).

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
...
+                logging.error('Cluster status reports failure')
+                rc = -1
+        else:
+            logging.info('Running test plan: ' + str(args.plan[0].value))
+            test_hare_bootstrap_shutdown(args)
+
+            cluster_status = check_cluster_status(path_to_cdf)
+            if cluster_status:
+                logging.error('Cluster status reports failure')
+                rc = -1
```
knekrasov, line 453: Here you could have just raised an exception.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
...
-            logging.error('Cluster status reports failure')
-            rc = -1
+            if not is_cluster_running():
+                logging.error('Cluster is not running. Cluster must be '
+                              'running for executing tests')
+                exit(-1)
+            cluster_status = check_cluster_status(path_to_cdf)
+            if cluster_status:
+                logging.error('Cluster status reports failure')
+                rc = -1
```
knekrasov, line 445: Here you could have just raised an exception.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
...
-                          'for executing tests')
-            exit(-1)
-        cluster_status = check_cluster_status(path_to_cdf)
-        if cluster_status:
-            logging.error('Cluster status reports failure')
-            rc = -1
+            if not is_cluster_running():
+                logging.error('Cluster is not running. Cluster must be '
+                              'running for executing tests')
+                exit(-1)
```
knekrasov, line 441: Here you could have just raised an exception.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,21 +255,202 @@ def test(args):
...
+
 def test_IVT(args):
     try:
         rc = 0
         path_to_cdf = args.file[0]
+        is_dev_opt_enbl = int(args.dev[0])
 
-        logging.info('Running test plan: ' + str(args.plan[0].value))
+        if not is_dev_opt_enbl:
+            logging.info('Running test plan: ' + str(args.plan[0].value))
```
knekrasov, line 436: Log message is repeated both in `if` and in `else` branches. Let's move it out of `if` statment.

provisioning/miniprov/hare_mp/main.py
knekrasov, line 445: Both if and else branches deserve separate functions. E.g.
```python

if dev_mode:
   ensure_cluster_running()
else:
    run_bootstrap_shutdown_test()
logging.info('Tests executed successfully')
```

provisioning/miniprov/hare_mp/main.py
```diff
@@ -593,13 +785,14 @@ def main():
...
                        handler_fn=init))
 
-    add_param_argument(
-        add_plan_argument(
-            add_file_argument(
-                add_subcommand(subparser,
-                               'test',
-                               help_str='Tests Hare component',
-                               handler_fn=test_IVT))))
+    add_dev_argument(
```
knekrasov, line 788: Do you like this stairway of functions?
If not, we could make use of `functools` to improve readability. Example: https://mathieularose.com/function-composition-in-python

knekrasov (changes requested):
Please also squash the commits. Thank you

provisioning/miniprov/hare_mp/main.py
pavankrishnat: used function which is already in use

provisioning/miniprov/hare_mp/main.py
pavankrishnat: For now I will remove the decorator, once PR#1679 merged will move completed code accordingly.

provisioning/miniprov/hare_mp/main.py
pavankrishnat: Sure

provisioning/miniprov/hare_mp/main.py
pavankrishnat: Dev argument is used to separate, my test interface from actual main branch. and using this Dev value as loop count.

provisioning/miniprov/hare_mp/main.py
pavankrishnat: I have not created it, I used it, to add my dev argument in IVT framework.

knekrasov: But wouldn't you improve it given that you touched this piece of code anyway?

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,17 +259,148 @@ def test(args):
...
+    cluster_sts = check_cluster_status(cdf_file)
+    if cluster_sts:
+        resp = execute(['journalctl', '--since', timeinfo, '>', logfile])
+        logging.info('created journal log: %s', logfile)
+        raise Exception('hctl status reports failure.')
+
+    logging.info('Successfully performed cleanup after testing')
+
+
+def _hctlstatus():
```
knekrasov, line 310: the function name tells nothing about what it does and what behavior is expected.

Suggestion: `get_hctl_status()`

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,17 +259,148 @@ def test(args):
...
+    process = subprocess.Popen(hctl_status,
+                               stdin=subprocess.PIPE,
+                               stdout=subprocess.PIPE,
+                               stderr=subprocess.PIPE,
+                               encoding='utf8')
+    logging.debug('Issuing CLI command: %s', hctl_status)
+    out, err = process.communicate()
+    exit_code = process.returncode
+    logging.debug('Finished. Exit code: %d', exit_code)
+    if exit_code:
```
knekrasov, line 324: This is a code duplication. Please refactor `execute()` function respectfully instead.

provisioning/miniprov/hare_mp/main.py
pavankrishnat: Sure will take it and do this change

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,17 +259,148 @@ def test(args):
...
+                raise Exception('After shutdown, hctl is running.')
+        else:
+            raise Exception(
+                f'Command hctl status exited with error code {exit_code}'
+                f'Command output: {err}')
+    else:
+        raise Exception('Fail to control cluster, exit the test.')
+
+
+def test_hare_bootstrap_shutdown(args):
```
knekrasov, line 338: Why do we have tests hardcoded in this main.py file?

Can we have IVT tests be written as valid pytest tests? 
- In this case they would be located in a special directory and/or marked with special [marker](url),
- The only thing  that the code in this mini provisioner CLI would have is to invoke pytest with proper command line arguments

Pros:
1. List of tests become visible and easily extensible
2. There is a standard way to run the tests. As a consequence, such test can be integrated with test coverage tools and with CI (because pytest support standard formats to report the test run results) in the future.


@vaibhavparatwar, @mssawant, @suvratjoshi Any thoughts?

provisioning/miniprov/hare_mp/main.py
```diff
@@ -254,17 +259,148 @@ def test(args):
...
+    """Test suite for single node hare init in loop."""
+    loop_count = int(args.dev[0])
+    hctl_shutdown = ['hctl', 'shutdown', '--skip-consul-stop']
+
+    test_hare_prereq()
+
+    if is_cluster_running():
+        _hctlstatus()
```
knekrasov, line 346: As far as I understand, this function will just invoke `hctl status` and return the stdout as the return value. That value is ignored. Then what is the point to invoke this function here?

knekrasov (changes requested):

provisioning/miniprov/hare_mp/main.py
pavankrishnat: Note: once [PR#1679](https://github.com/Seagate/cortx-hare/pull/1679) is merged I will move my changes accordingly.

provisioning/miniprov/hare_mp/main.py
pavankrishnat: hctl status returns non zero error code when cluster is not running, In this case I don't want to exit, so I added few checks.

knekrasov: I don't see a problem here.
1. `execute()` raises an exception if the exit code is non-zero
2. you can catch that exception here in your code (and potentially - analyze the exception).
3. ..and keep processing after that. No exit will be triggered.

But all that unpleasant things like `PIPE`, `returncode` etc would be kept in one place only - in `execute()` function.

provisioning/miniprov/hare_mp/main.py
knekrasov: I believe you are not blocked by #1679. In other words, you can start putting ivt tests into some folder in the source code even without that patch. Do I miss anything|?

PR #1727, opened 2021-07-29
EOS-21138: Support '--pre-factory' option in Hare factory reset

supriyachavan4398:
Solution:
By default, cleanup will take you to post-factory(i.e. after post-install) and Pre-factory will take you to before-factory (i.e. undo everything that was done in post-install).
So we need to support this option in cleanup state.
In hare code post-install is no-op. So `--pre-factory` option is also no-op for us. 
We can use this command to cleanup with --pre-factory `/opt/seagate/cortx/hare/bin/hare_setup cleanup --config 'json:///root/hare.cleanup.conf.tmpl.1-node' --pre-factory`

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -43,8 +43,8 @@ from hare_mp.systemd import HaxUnitTransformer
 from hare_mp.validator import Validator
 
 # Logger details
-LOG_DIR = "/var/log/seagate/hare/"
-LOG_FILE = "/var/log/seagate/hare/setup.log"
+LOG_DIR = "/var/log/seagate/hare/hare_deployment/"
+LOG_FILE = "/var/log/seagate/hare/hare_deployment/setup.log"
 LOG_FILE_SIZE = 5 * 1024 * 1024
 
 
@@ -310,14 +310,20 @@ def kv_cleanup():
         raise RuntimeError('Error during key delete in transaction')
 
 
+def pre_factory():
+    logging.info('Executing pre-factory cleanup command...')
+    deployment_logs_cleanup()
+
+
 def cleanup(args):
     try:
-        rc = -1
+        kv_cleanup()
+        logs_cleanup()
+        config_cleanup()
+        if args.pre_factory:
+            pre_factory()
 
-        if kv_cleanup() == 0 and config_cleanup() == 0 and logs_cleanup() == 0:
-            rc = 0
-
-        exit(rc)
+        exit(0)
     except Exception as error:
         logging.error('Error during cleanup (%s)', error)
         exit(-1)
@@ -325,13 +331,11 @@ def cleanup(args):
 
 def logs_cleanup():
     try:
-        logging.info('Cleaning up hare log directory(/var/log/seagate/hare)')
-        os.system('rm -rf /var/log/seagate/hare')
+        logging.info('Cleaning up hare log directory(/var/log/seagate/hare/*)')
+        os.system('rm -f /var/log/seagate/hare/*')
 
-        return 0
     except Exception as error:
-        logging.error('Error during logs cleanup (%s)', error)
-        return -1
+        raise RuntimeError(f'Error during logs cleanup : key={error}')
 
 
 def config_cleanup():
@@ -339,10 +343,18 @@ def config_cleanup():
         logging.info('Cleaning up hare config directory(/var/lib/hare)')
         os.system('rm -rf /var/lib/hare/*')
 
-        return 0
     except Exception as error:
-        logging.error('Error during config cleanup (%s)', error)
-        return -1
+        raise RuntimeError(f'Error during config cleanup : key={error}')
+
+
+def deployment_logs_cleanup():
+    try:
+        logging.info('''Cleaning up hare deployment log directory
+            (/var/log/seagate/hare/hare_deployment)''')
+        os.system('rm -rf /var/log/seagate/hare/hare_deployment')
+
+    except Exception as error:
+        raise RuntimeError(f'Error during deployment log cleanup: key={error}')
 
 
 def generate_support_bundle(args):
@@ -572,6 +584,16 @@ def add_param_argument(parser):
     return parser
 
 
+def add_factory_argument(parser):
+    parser.add_argument('--pre-factory',
+                        help="""Deletes contents of /var/log/seagate/hare,
+                        /var/lib/hare and /var/motr. Undoes everything that
+                        is done in post-install stage of hare provisioner""",
+                        required=False,
+                        action='store_true')
+    return parser
+
+
 def main():
     p = argparse.ArgumentParser(description='Configure hare settings')
     subparser = p.add_subparsers()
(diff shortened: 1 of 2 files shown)


cortx-admin:

knekrasov:
@supriyachavan4398 `--pre-factory` option is now supported by the CLI but it is effectively no-op. Is it expected?

This makes me think that the problem is not stated properly in the PR description. What problem do we solve?

mssawant:
@supriyachavan4398 , kindly restrict the commit message content to 80 columns.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -565,6 +565,14 @@ def add_param_argument(parser):
     return parser
 
 
+def add_factory_argument(parser):
+    parser.add_argument('--pre-factory',
```
mssawant, line 569: from the document https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/221642873/CORTX+Factory+Manufacturing+Process, I see some confusion,
```
Perform Factory Reset
Testing may have generated a log Login using CORTX cluster CLI. All the data and configuration needs to be cleared. After you are done with all the tests you can perform the factory reset using the following commands. 


$ cortx_cli cluster reset [--type {data|all}] [--pre-factory]
Note: 

**This internally calls <component>_setup reset --config <URL> and <component>_setup cleanup--config <URL> for all components.**

Normally factory reset will bring the node up to the factory configured stage. There will be an additional option provided to undo “post_install” activities as well. For example,  Motr LVM configuration will be removed when “--factory” is passed. By default deployment logs are not cleaned up but they will be cleaned up when --factory is passed to it. 

Node will marked with appropriate stage.
```
But Hare HLD for the same mentions,
```
Note here that '--pre-factory' option here will be an additional option provided to undo “post_install” activities. For more info refer to above doc

Hare’s factory reset(reset + cleanup) currently do not handle this option.

We need to support this option in future(
```
So, few questions,
1. will `--pre-factory` option be passed to the component's setup utility? or is it for provisioner and provisioner will internally invoke reset + cleanup?
2. is the option `--pre-factory` or --factory` as mentioned in the factory manufacturing process document?
3. If hare needs to support `--pre-factory` then, do we need to invoke `reset + cleanup` as part of `--pre-factory` implementation? cc @SwapnilGaonkar7 , @vaibhavparatwar .

supriyachavan4398:
> @supriyachavan4398 `--pre-factory` option is now supported by the CLI but it is effectively no-op. Is it expected?
> 
> This makes me think that the problem is not stated properly in the PR description. What problem do we solve?

In Hare code, post-install supports to no-op. So for this task, we can't revert changes to before post-install. That's why `--pre-factory` option is now only supported by the CLI.

provisioning/miniprov/hare_mp/main.py
supriyachavan4398: `<component>_setup cleanup --config <URL> --pre-factory` this command needs to be supported for pre-factory.
pre-factory is an additional option to cleanup interface to undo post-install phase.
We don't need to call to reset and cleanup as pre-factory is an additional option for the cleanup phase.
@mssawant Please let me know if it answers your questions. Ref. https://seagate-systems.atlassian.net/wiki/spaces/PRIVATECOR/pages/559743237/F-97+Reset+Cleanup+Pre-Factory

mssawant: Yes it does. Thank you.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -565,6 +565,14 @@ def add_param_argument(parser):
...
+                        help='Pre-factory will take you to before-factory',
```
mssawant, line 570: The help message is not very clear, `pre-factory` is synonymous with `before-factory`. Consider,
```
Deletes contents of /var/log/seagate/hare, /var/lib/hare and /var/motr. Undoes everything that is done in post-install stage of hare mini-provisioner.
```

mssawant: Let's define a stub for `pre-factory` option and invoke it from cleanup() if `--pre-factory` argument is passed to cleanup command.

mssawant: Make sure that the pre-factory stub is invoked if `--pre-factory` argument is passed to cleanup command.

supriyachavan4398: Added new stub for `pre-factory` and invoked it from cleanup() if `--pre-factory` argument is passed to cleanup command.

supriyachavan4398: > The help message is not very clear, `pre-factory` is synonymous with `before-factory`. Consider,
> 
> ```
> Deletes contents of /var/log/seagate/hare, /var/lib/hare and /var/motr. Undoes everything that is done in post-install stage of hare mini-provisioner.
> ```

Added this content in the help message.

supriyachavan4398:
Deployment job: http://eos-jenkins.colo.seagate.com/job/Cortx-Deployment/job/VM-Deployment-R2-1Node/1796/

provisioning/miniprov/hare_mp/main.py
```diff
@@ -310,39 +310,51 @@ def kv_cleanup():
         raise RuntimeError('Error during key delete in transaction')
 
 
+def pre_factory():
+    logging.info('Executing pre-factory cleanup command...')
+    deployment_logs_cleanup()
```
mssawant, line 315: We need to cleanup /var/motr/ as well.
Also I think Let's delete `/etc/sysconfig/m0d-0x7200000000000001*` and `/etc/sysconfig/s3server-0x7200000000000001*`

SwapnilGaonkar7: @mssawant https://seagate-systems.atlassian.net/wiki/spaces/~527368708/pages/edit-v2/563872735?draftShareId=d5838841-79a4-4353-9d31-2e149c19c640
According to above doc, these changes will be taken care by Motr.
Is my understanding correct?

mssawant: @SwapnilGaonkar7, As per the document, motr is cleaning `/var/motr/m0d-*`, so we need to cleanup up `/var/motr/hax` and sysconfig file as mentioned above.

supriyachavan4398: We are cleaning  `/var/motr/m0d-*` and `/var/motr/hax`  in  [EOS-23169](https://jts.seagate.com/browse/EOS-23169)  this story. So, I think we can do it in the next task.  Is this okay?
cc. @mssawant @SwapnilGaonkar7 @vaibhavparatwar

mssawant: May be we can do this in a separate ticket. cc @vaibhavparatwar

provisioning/miniprov/hare_mp/main.py
```diff
@@ -310,39 +310,51 @@ def kv_cleanup():
...
-        logging.error('Error during config cleanup (%s)', error)
-        return -1
+        raise RuntimeError(f'Error during config cleanup : key={error}')
+
+
+def deployment_logs_cleanup():
+    try:
+        logging.info('''Cleaning up hare deployment log directory
+            (/var/log/seagate/hare/hare_deployment)''')
+        os.system('rm -rf /var/log/seagate/hare/hare_deployment')
```
mssawant, line 354: Another improvement that I see is using `os.remove(), os.unlink() or shutil.rmtree()` instead of a direct call to unix shell command. I think using python functions instead would increase the portability factor of the code.
What is your opinion @knekrasov ? cc @vaibhavparatwar

mssawant (approved):

vaibhavparatwar (approved):

knekrasov (approved):

PR #1729, opened 2021-07-30
EOS-23006 Map a motr process to a controller and not a node

SwapnilGaonkar7:
**Problem**:
Node is now mapped to a motr enclosure and not a controller.
Presently while notifying controller failure, hax maps entire node to
a controller and not the failed motr process (ioservice).

**Solution**:
In case of IO service failure we are just marking controller
  as failed and not marking whole node as failed(but if all the Io services
  are failed then we will mark node as failed).
If we mark node as failed(because all IO services are failed) and then
  if IO service comes back then we will mark node as online.
In case of node failure we are mapping all the controllers to node instead of just 1.

### DIFF
--- a/hax/hax/motr/__init__.py
+++ b/hax/hax/motr/__init__.py
@@ -19,7 +19,7 @@
 import ctypes as c
 import logging
 from errno import EAGAIN
-from typing import Any, List
+from typing import Any, List, Optional
 
 from hax.exception import ConfdQuorumException, RepairRebalanceException
 from hax.message import (EntrypointRequest, FirstEntrypointRequest,
@@ -278,13 +278,36 @@ class Motr:
                 continue
             note = HaNoteStruct(st.fid.to_c(), ha_obj_state(st))
             notes.append(note)
+
+            if st.fid.container == ObjT.PROCESS.value:
+                self.consul_util.set_process_state(st.fid, st.status)
+
             notes += self._generate_sub_services(note, self.consul_util,
                                                  notify_devices)
             # For process failure, we report failure for the corresponding
-            # node (enclosure) and CVGs.
+            # node (enclosure) and CVGs if all Io services are failed.
             if (st.fid.container == ObjT.PROCESS.value
                     and st.status in (ServiceHealth.FAILED, ServiceHealth.OK)):
-                notes += self.notify_node_status(note)
+                # Check if we need to mark node as failed,
+                # otherwise just mark controller as failed/OK
+                # If we receive process failure then we will check if all IO
+                # services are failed, if True then we will mark node as failed
+                # If we receive process 'OK' then we will check if node is
+                # not in failed state then we will mark node as OK
+                # If both the above conditions are not true then we will just
+                # mark controller status
+                is_node_failed = self.is_node_failed(note)
+                if (st.status == ServiceHealth.FAILED
+                        and is_node_failed):
+                    notes += self.notify_node_status(note)
+                elif (st.status == ServiceHealth.OK
+                        and not is_node_failed):
+                    notes += self.notify_node_status(note)
+                else:
+                    ctrl_note = self.get_ctrl_status(note)
+                    if ctrl_note is not None:
+                        notes.append(ctrl_note)
+
             if st.fid.container == ObjT.DRIVE.value:
                 self.consul_util.update_drive_state([st.fid], st.status)
         if not notes:
@@ -345,8 +368,10 @@ class Motr:
                   len(event.nvec))
         notes: List[HaNoteStruct] = []
         for n in event.nvec:
-            n.note.no_state = self.consul_util.get_conf_obj_status(
-                ObjT[n.obj_t], n.note.no_id.f_key)
+            n.note.no_state = HaNoteStruct.M0_NC_ONLINE
+            if (n.obj_t in (ObjT.PROCESS.name, ObjT.SERVICE.name)):
+                n.note.no_state = self.consul_util.get_conf_obj_status(
+                    ObjT[n.obj_t], n.note.no_id.f_key)
             notes.append(n.note)
 
         LOG.debug('Replying ha nvec of length ' + str(len(event.nvec)))
@@ -406,16 +431,43 @@ class Motr:
 
         node_fid = self.consul_util.get_node_fid(node)
         encl_fid = self.consul_util.get_node_encl_fid(node)
-        ctrl_fid = self.consul_util.get_node_ctrl_fid(node)
-        LOG.debug('node_fid: %s encl_fid: %s ctrl_fid: %s with state: %s',
-                  node_fid, encl_fid, ctrl_fid, new_state)
+        ctrl_fids = self.consul_util.get_node_ctrl_fids(node)
+        LOG.debug('node_fid: %s encl_fid: %s ctrl_fids: %s with state: %s',
+                  node_fid, encl_fid, ctrl_fids, new_state)
 
         notes = []
-        if node_fid and encl_fid and ctrl_fid:
+        if node_fid and encl_fid:
             notes = [HaNoteStruct(no_id=x.to_c(), no_state=new_state)
-                     for x in [node_fid, encl_fid, ctrl_fid]]
+                     for x in [node_fid, encl_fid]]
+
+        if ctrl_fids:
+            for x in ctrl_fids:
+                notes.append(HaNoteStruct(no_id=x.to_c(), no_state=new_state))
 
         return notes
+
+    def get_ctrl_status(self,
+                        proc_note: HaNoteStruct) -> Optional[HaNoteStruct]:
+        new_state = proc_note.no_state
+        proc_fid = Fid.from_struct(proc_note.no_id)
+        assert ObjT.PROCESS.value == proc_fid.container
+        LOG.debug('Notifying ctrl status for process_fid=%s state=%s',
+                  proc_fid, new_state)
+
+        ctrl_fid = self.consul_util.get_ioservice_ctrl_fid(proc_fid)
+
+        if ctrl_fid:
+            return HaNoteStruct(no_id=ctrl_fid.to_c(), no_state=new_state)
+        return None
+
+    # Check if all the IO services of node are failed
+    def is_node_failed(self, proc_note: HaNoteStruct):
+        proc_fid = Fid.from_struct(proc_note.no_id)
+        assert ObjT.PROCESS.value == proc_fid.container
+
+        node = self.consul_util.get_process_node(proc_fid)
+
+        return self.consul_util.all_io_services_failed(node)
 
     def notify_hax_stop(self):
         LOG.debug('Notifying hax stop')
--- a/hax/hax/util.py
+++ b/hax/hax/util.py
@@ -658,33 +658,139 @@ class ConsulUtil:
         return None
 
     @repeat_if_fails()
-    def get_node_ctrl_fid(self, node: str) -> Optional[Fid]:
+    def get_node_ctrl_fids(self, node: str) -> Optional[List[Fid]]:
         """
-        Returns the fid of the controller for the given node.
-
         Parameters:
             node : hostname of the node.
         """
         # Example,
-        # {
+        # [
         #    "key": "m0conf/sites/0x5300000000000001:0x1/
         #            racks/0x6100000000000001:0x2/encls/
         #            0x6500000000000001:0x4/ctrls/0x6300000000000001:0x5",
-        # },
+        # ]
         encl_fid = self.get_node_encl_fid(node)
         if not encl_fid:
             return None
-        ctrl_items = self.kv.kv_get('m0conf/sites', recurse=True)
+        ctrl_items = self.get_all_sites()
         regex = re.compile(
             f'^m0conf\\/.*\\/racks\\/.*\\/encls\\/{encl_fid}\\/ctrls\\/'
             '([^/]+)$')
+        list_fids: List[Fid] = []
         for ctrl in ctrl_items:
             match_result = re.match(regex, ctrl['Key'])
             if not match_result:
                 continue
             ctrl_fid: str = match_result.group(1)
-            return Fid.parse(ctrl_fid)
+            list_fids.append(Fid.parse(ctrl_fid))
+        return list_fids
+
+    @repeat_if_fails()
+    def get_io_service_devices(self,
+                               ioservice_fid: Fid) -> Optional[List[str]]:
+        if not ioservice_fid:
+            return None
+        # Example key m0conf/nodes/0x6e00000000000001:0x3/processes/
+        #   0x7200000000000001:0x15/services/0x7300000000000001:0x17/sdevs/
+        #   0x6400000000000001:0x18:{"path": "/dev/sdc", "state": "offline"}
+        sdev_items = self.kv.kv_get('m0conf/nodes', recurse=True)
+        regex = re.compile(
+            f'^m0conf\\/.*\\/processes\\/{ioservice_fid}\\/.*\\/sdevs\\/.*$')
+        sdev_fids = []
+        for sdev in sdev_items:
+            match_result = re.match(regex, sdev['Key'])
+            if not match_result:
+                continue
+            sdev_key: str = match_result.group(0)
+            key = sdev_key.split('/')
+            sdev_fids.append(key[8])
+        return sdev_fids
+
+    @repeat_if_fails()
+    def get_all_sites(self):
+        return self.kv.kv_get('m0conf/sites', recurse=True)
+
+    @repeat_if_fails()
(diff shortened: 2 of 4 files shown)


mssawant:
@SwapnilGaonkar7 , please describe the Problem and the Solution with additional details in the commit message.

hax/hax/motr/__init__.py
```diff
@@ -405,17 +429,48 @@ def notify_node_status(self,
...
+
+        ctrl_fid = self.consul_util.get_ioservice_ctrl_fid(proc_fid)
+
+        if ctrl_fid is not None:
+            return HaNoteStruct(no_id=ctrl_fid.to_c(), no_state=new_state)
+        else:
+            return None
+
+    # Check if all the IO services of node are failed
+    def check_for_node_failure(self, proc_note: HaNoteStruct) -> bool:
```
vaibhavparatwar, line 463: @mssawant comments:
1. can you change name of his function to "is_node_failed"
2. can we pass process fids instead of HaNoteStruct?

hax/hax/motr/__init__.py
```diff
@@ -405,17 +429,48 @@ def notify_node_status(self,
...
+        proc_fid = Fid.from_struct(proc_note.no_id)
+        assert ObjT.PROCESS.value == proc_fid.container
+
+        node = self.consul_util.get_process_node(proc_fid)
+
+        if self.consul_util.is_all_IOservices_failed(node):
```
vaibhavparatwar, line 469: rename - is_all_IOservices_failed to are_all_IOservices_failed

hax/hax/util.py
```diff
@@ -639,33 +639,119 @@ def get_node_fid(self, node: str) -> Optional[Fid]:
...
         for ctrl in ctrl_items:
             match_result = re.match(regex, ctrl['Key'])
             if not match_result:
                 continue
             ctrl_fid: str = match_result.group(1)
+            list_fids.append(Fid.parse(ctrl_fid))
+        return list_fids
+
+    @repeat_if_fails()
+    def get_ioservice_ctrl_fid(self, ioservice_fid: Fid,) -> Optional[Fid]:
```
vaibhavparatwar, line 673: @mssawant  comment: 
add TODO:
Instead of comparing devices, add a mapping from controller <-> I/O service

hax/hax/util.py
```diff
@@ -639,33 +639,119 @@ def get_node_fid(self, node: str) -> Optional[Fid]:
...
+        if not ioservice_fid:
+            return None
+        # Example key m0conf/nodes/0x6e00000000000001:0x3/processes/
+        #   0x7200000000000001:0x15/services/0x7300000000000001:0x17/sdevs/
+        #   0x6400000000000001:0x18:{"path": "/dev/sdc", "state": "offline"}
+        sdev_items = self.kv.kv_get('m0conf/nodes', recurse=True)
+        regex = re.compile(
+            f'^m0conf\\/.*\\/processes\\/{ioservice_fid}\\/.*\\/sdevs\\/.*$')
+        sdev_fid = None
+        for sdev in sdev_items:
```
vaibhavparatwar, line 689: @mssawant comment:

Can we create a function for this snippet (get_ioservice_devices, get_controller_devices) and common out sdev related code. Pass regex as an argument? - one function to list all sdevs and pass the regex- one for ioservice and one for controller.

hax/hax/motr/__init__.py
```diff
@@ -281,13 +281,35 @@ def ha_obj_state(st):
...
+                # Check if we need to mark node as failed,
+                # otherwise just mark controller as failed/OK
+                # If we receive process failure then we will check if all IO
+                # services are failed, if True then we ill mark node as failed
+                # If we receive process 'OK' then wwe will check if node is
+                # not in failed state then we ill mark node as OK
+                # If both the above conditions are not true then we will just
+                # mark controller status
+                if (st.status == ServiceHealth.FAILED
+                        and self.check_for_node_failure(note) is True):
```
vaibhavparatwar, line 303: - check_for_node_failure can be invoked once before loop.
- Some typos in comment section

knekrasov, line 303:
```suggestion
                        and self.check_for_node_failure(note)):
```
Which means that `check_for_node_failure` is not the best name. It seems like the functionality behind corresponds to "Is node failed?" predicate.

knekrasov (commented):
@SwapnilGaonkar7 Is it possible to implement a couple of auto tests that prove that your patch works properly?

Please note that we already have the corresponding infrastructure: see https://github.com/Seagate/cortx-hare/blob/main/hax/test/integration/test_motr.py

cc @vaibhavparatwar

hax/hax/motr/__init__.py
```diff
@@ -405,17 +429,48 @@ def notify_node_status(self,
...
-                     for x in [node_fid, encl_fid, ctrl_fid]]
+                     for x in [node_fid, encl_fid]]
+
+        if int(len(ctrl_fids)) != 0:
+            for x in ctrl_fids:
+                notes.append(HaNoteStruct(no_id=x.to_c(), no_state=new_state))
 
         return notes
 
+    def notify_ctrl_status(self,
```
knekrasov, line 447: The word "notify" is misleading. It notifies nobody, but just creates the HaNoteStruct if there is something to report about. Please think of a better name.

hax/hax/motr/__init__.py
```diff
@@ -405,17 +429,48 @@ def notify_node_status(self,
...
+        assert ObjT.PROCESS.value == proc_fid.container
+        LOG.debug('Notifying ctrl status for process_fid=%s state=%s',
+                  proc_fid, new_state)
+
+        ctrl_fid = self.consul_util.get_ioservice_ctrl_fid(proc_fid)
+
+        if ctrl_fid is not None:
+            return HaNoteStruct(no_id=ctrl_fid.to_c(), no_state=new_state)
+        else:
+            return None
```
knekrasov, line 460:
```suggestion
        if ctrl_fid:
            return HaNoteStruct(no_id=ctrl_fid.to_c(), no_state=new_state)
        return None
```

hax/hax/motr/__init__.py
knekrasov: `is` is a well-known prefix for predicate functions so when the developer sees it, (s)he can guess what to expect from the function. But I've never seen functions with `are_` prefix 😁 

On the other hand, `is` doesn't correspond to the plural form.
Suggestion: rename to `all_io_services_failed`.

hax/hax/motr/__init__.py
```diff
@@ -405,17 +429,48 @@ def notify_node_status(self,
...
+    def check_for_node_failure(self, proc_note: HaNoteStruct) -> bool:
+        proc_fid = Fid.from_struct(proc_note.no_id)
+        assert ObjT.PROCESS.value == proc_fid.container
+
+        node = self.consul_util.get_process_node(proc_fid)
+
+        if self.consul_util.is_all_IOservices_failed(node):
+            return True
+        else:
+            return False
```
knekrasov, line 472: Keep it simple.

```suggestion
        return self.consul_util.is_all_IOservices_failed(node)
```

hax/hax/motr/__init__.py
SwapnilGaonkar7: `can you change name of his function to "is_node_failed"`
Done
'can we pass process fids instead of HaNoteStruct?'
We dont have process fids here, that info is present in consul which can be accesses by utils only

hax/hax/motr/__init__.py
SwapnilGaonkar7: Done

hax/hax/util.py
SwapnilGaonkar7: Done

hax/hax/motr/__init__.py
SwapnilGaonkar7: Done

SwapnilGaonkar7: Done

hax/hax/motr/__init__.py
SwapnilGaonkar7: Done

hax/hax/motr/__init__.py
SwapnilGaonkar7: Done

hax/hax/util.py
SwapnilGaonkar7: Done

hax/hax/util.py
```diff
@@ -639,33 +639,129 @@ def get_node_fid(self, node: str) -> Optional[Fid]:
...
+        if not sdev_fids:
+            return None
+
+        sdev_fid = sdev_fids[0]
+
+        # Example key m0conf/sites/0x5300000000000001:0x1/racks/
+        #    0x6100000000000001:0x2/encls/0x6500000000000001:0x4/ctrls/
+        #    0x6300000000000001:0x5/drives/0x6b00000000000001:0x38:
+        #    {"sdev": "0x6400000000000001:0x37", "state": "M0_NC_UNKNOWN"}
+        sites_items = self.kv.kv_get('m0conf/sites', recurse=True)
```
mssawant, line 714: We are doing too many recursive calls to consul kv, mainly for `m0conf/nodes` and now for `m0conf/sites`. In PR https://github.com/Seagate/cortx-hare/pull/1737 I have added caching for such recursive calls. I think let's have a function get_all_sites() that will execute recursive call to consul kv for `m0conf/sites`. Please use that where `m0conf/sites` is referred in your PR. This will help identify the redundant calls. Later we can cache the result.

mssawant (approved):
Please address posted comments. Rest looks okay. We need to identify the redundant functions and consolidate them later on.

hax/hax/motr/__init__.py
SwapnilGaonkar7: Done

hax/hax/util.py
SwapnilGaonkar7: Done

hax/hax/motr/__init__.py
```diff
@@ -281,13 +281,36 @@ def ha_obj_state(st):
...
+                # If we receive process 'OK' then we will check if node is
+                # not in failed state then we will mark node as OK
+                # If both the above conditions are not true then we will just
+                # mark controller status
+                is_node_failed = self.is_node_failed(note)
+                if (st.status == ServiceHealth.FAILED
+                        and is_node_failed):
+                    notes += self.notify_node_status(note)
+                elif (st.status == ServiceHealth.OK
+                        and is_node_failed is False):
```
knekrasov, line 307:
```suggestion
                        and not is_node_failed):
```

hax/test/integration/test_motr.py
```diff
@@ -71,6 +72,77 @@ def fn(trace: Invocation) -> bool:
...
+    '''
+    def fn(trace: Invocation) -> bool:
+        if len(trace.args) < 2:
+            return False
+        note = trace.args[1]
+        if (not isinstance(note, ctypes.Array) or
+                len(note) != 5):
+            return False
+
+        fids = [FidStruct(0x7200000000000001, 0x15),
```
knekrasov, line 93: The matcher code implicitly requires the order in which Fids appear in the HA notes. Do we really want to make the order that hard?

hax/test/integration/test_motr.py
```diff
@@ -177,3 +242,208 @@ def my_services(name):
...
+                ('m0conf/nodes/0x6e00000000000001:0x3/processes'
+                 '/0x7200000000000001:0xa/services/0x7300000000000001:0xc',
+                 '{\"name\": \"ios\", \"state\": \"failed\"}')]
+            ]
+        elif (key == 'm0conf/nodes/0x6e00000000000001:0x3/processes'
+              '/0x7200000000000001:0x15' and recurse):
+            return [new_kv(k, v) for k, v in [(
+                'm0conf/nodes/0x6e00000000000001:0x3/processes'
+                '/0x7200000000000001:0x15',
+                '{\"name\": \"m0_server\", \"state\": \"failed\"}'),
```
knekrasov, line 305: [Optional]: Instead of playing with escaping characters, one could use something like `json.dumps({'name': 'm0_server', 'state': 'failed'}). It is a bit more readable and such code can be formatted automatically by the formatter tool (while long strings are not that easy to format).

hax/test/integration/test_motr.py
```diff
@@ -177,3 +242,208 @@ def my_services(name):
...
+                        return_value=Fid(0x6e00000000000001, 0x3))
+    mocker.patch.object(consul_util,
+                        'get_node_encl_fid',
+                        return_value=Fid(0x6500000000000001, 0x4))
+
+    motr.broadcast_ha_states([HAState(fid=Fid(0x7200000000000001, 0x15),
+                                      status=ServiceHealth.FAILED)])
+
+    traces = motr._ffi.traces
+    assert AssertionPlan(tr_and(tr_method('ha_broadcast'),
```
knekrasov, line 448: I would also like to see that in such a case NO node failure notification happened. In other words, there is no node fid in the broadcast (is it a correct understanding of your feature, BTW?)

knekrasov (commented):
Nice work on the tests BTW!

hax/test/integration/test_motr.py
SwapnilGaonkar7: Done

hax/test/integration/test_motr.py
SwapnilGaonkar7: Done

hax/test/integration/test_motr.py
SwapnilGaonkar7: Done, modified code to remove dependency on order

hax/hax/motr/__init__.py
SwapnilGaonkar7: Done

hax/test/integration/test_motr.py
```diff
@@ -71,6 +73,109 @@ def fn(trace: Invocation) -> bool:
...
+
+            if not bMatch:
+                return False
+
+        return True
+
+    return fn
+
+
+def node_not_failed() -> TraceMatcher:
```
knekrasov, line 122: Not sure that it works.

You see, TraceMatcher is a function applied to one particular invocation trace. TraceMatcher "doesn't see" all the traces at the same time, it "sees" just one trace at a time. TraceMatcher can say whether this particular trace matches the criteria. But it can't actually say whether ALL traces match or don't match.

So I would go another way.
1. Implement a matcher that matches a node in the traces ("a trace contains FID with node FID")
2. And then implement `not exist` operation (presumably at the level of AssertionPlan class).

So effectively you could build a predicate "the exist no invocation trace that has a node fid"

SwapnilGaonkar7: @knekrasov I think we are dealing with single trace only(if I am not wrong).
We are checking if Fid is present in that trace args or not.
Let me know your inputs on this

knekrasov: There is no proof that there is only one trace :-)

In other words, if by any chance or because of a new piece of functionality there will be 2 traces, your test should still be valid. If the second or the third trace contains a node, then the test should fail

hax/test/integration/test_motr.py
```diff
@@ -71,6 +73,109 @@ def fn(trace: Invocation) -> bool:
...
+                len(notes) != 5):
+            return False
+
+        fids = [FidStruct(0x7200000000000001, 0x15),
+                FidStruct(0x6e00000000000001, 0x3),
+                FidStruct(0x6500000000000001, 0x4),
+                FidStruct(0x6300000000000001, 0x5),
+                FidStruct(0x6300000000000001, 0x6)]
+
+        for _, note in enumerate(notes):
```
knekrasov, line 100: If the call was invoked with one and the same fid duplicated multiple times, this matcher will still return true.

In other words, if the HaNoteStruct erroneously contains `[FidStruct(0x6300000000000001, 0x6), FidStruct(0x6300000000000001, 0x6), FidStruct(0x6300000000000001, 0x6)]`, this matcher will return True.

SwapnilGaonkar7: Done

hax/test/integration/test_motr.py
```diff
@@ -71,6 +73,110 @@ def fn(trace: Invocation) -> bool:
...
+                if (fid.f_container == out_fid.f_container and
+                        fid.f_key == out_fid.f_key):
+                    bMatch = True
+                    del fids[i]
+                    break
+
+            if not bMatch:
+                return False
+
+        return True
```
knekrasov, line 118:
```suggestion
        # fids is empty -> all required fids were found
        return not fids
```

knekrasov (approved):
Good enough. Thanks!

hax/test/integration/test_motr.py
SwapnilGaonkar7: Done

PR #1730, opened 2021-07-31
Problem: Invalid allowed_failures vector values

mssawant:
If allowed_failures vector is not populated in CDF, cfgen calculates invalid
default allowed_failures vector with incorrect values for enclosure and
controllers.

Solution:
Check for empty allowed_failures vector from CDF and calculate the allowed_failures
vector based on the cluster configuration.

### DIFF
--- a/cfgen/cfgen
+++ b/cfgen/cfgen
@@ -17,6 +17,7 @@ import subprocess
 import sys
 from typing import (Any, Callable, Dict, Iterable, Iterator, List, NamedTuple,
                     Optional, Set, Tuple)
+from math import floor, ceil
 import yaml
 import logging
 
@@ -24,6 +25,11 @@ import logging
 __version__ = '0.14'
 
 LOG = logging.getLogger('cfgen')
+
+faultToleranceInfo = NamedTuple('faultToleranceInfo', [
+                            ('max_drives_per_node', int),
+                            ('max_drives_per_ctrl', int),
+                            ('min_drives_per_node', int)])
 
 
 def repeat_on_cmd_timeout(max_retries=-1):
@@ -1230,6 +1236,37 @@ def cluster_encls(m0conf: Dict[Oid, Any]) -> List[Oid]:
     return encls
 
 
+def cluster_ctrls(m0conf: Dict[Oid, Any]) -> List[Oid]:
+    ctrls: List[Oid] = []
+    for item in m0conf:
+        if item.type is ObjT.controller:
+            ctrls.append(item)
+    return ctrls
+
+
+def drives_per_node_and_ctrl(m0conf: Dict[Oid, Any]) -> faultToleranceInfo:
+    max_drives_per_node = 0
+    max_drives_per_ctrl = 0
+    min_drives_per_node = 0
+    drives_per_node = 0
+    for encl_id in m0conf:
+        if encl_id.type is ObjT.enclosure:
+            for ctrl_id in m0conf[encl_id].ctrls:
+                drives_per_ctrl = len(m0conf[ctrl_id].drives)
+                drives_per_node += drives_per_ctrl
+                if max_drives_per_ctrl < drives_per_ctrl:
+                    max_drives_per_ctrl = drives_per_ctrl
+            if max_drives_per_node < drives_per_node:
+                max_drives_per_node = drives_per_node
+            if (min_drives_per_node == 0 or
+                    min_drives_per_node > drives_per_node):
+                min_drives_per_node = drives_per_node
+            drives_per_node = 0
+    return faultToleranceInfo(max_drives_per_node=max_drives_per_node,
+                              max_drives_per_ctrl=max_drives_per_ctrl,
+                              min_drives_per_node=min_drives_per_node)
+
+
 Failures = NamedTuple('Failures', [('site', int),
                                    ('rack', int),
                                    ('enclosure', int),
@@ -1276,6 +1313,46 @@ class ConfPool(ToDhall):
         return '{ %s }' % args
 
     @classmethod
+    def gen_sns_tolerance(cls, m0conf: Dict[Oid, Any],
+                          pool_desc: Dict[str, Any]) -> Failures:
+        encls_nr = len(cluster_encls(m0conf))
+        ctrls_per_encl = floor(len(cluster_ctrls(m0conf)) / encls_nr)
+        data_units = pool_desc['data_units']
+        parity_units = pool_desc['parity_units']
+        spare_units = pool_desc.get('spare_units', 0)
+        total_units = data_units + parity_units + spare_units
+        fault_tol_info = drives_per_node_and_ctrl(m0conf)
+        min_pool_width = fault_tol_info.min_drives_per_node * encls_nr
+        actual_pool_width = len(pool_drives(m0conf, pool_desc))
+        if ((min_pool_width < actual_pool_width) and
+                (min_pool_width / total_units) > 0):
+            # Pool is asymmetric, all the nodes don't have same number of
+            # drives.
+            # if (min_pool_width / total_units) > 0 means that one of the
+            # nodes can probably get more than K units of a parity group.
+            # Thus we cannot allow node failures.
+            enc_failure_allowed = 0
+            if fault_tol_info.max_drives_per_ctrl > parity_units:
+                ctrl_failure_allowed = 0
+        else:
+            enc_failure_allowed_FP = parity_units / ceil(total_units /
+                                                         encls_nr)
+            enc_failure_allowed = floor(enc_failure_allowed_FP)
+            temp = ctrls_per_encl * enc_failure_allowed
+            ctrl_failure_allowed = min(temp, parity_units)
+        return Failures(0, 0, enc_failure_allowed, ctrl_failure_allowed,
+                        parity_units)
+
+    @classmethod
+    def gen_md_tolerance(cls, m0conf: Dict[Oid, Any],
+                         pool_desc: Dict[str, Any]) -> Failures:
+        encls_nr = len(cluster_encls(m0conf))
+        ctrls_per_encl = floor(len(cluster_ctrls(m0conf)) / encls_nr)
+        encl_failures = encls_nr - 1
+        ctrl_failures = ctrls_per_encl * encl_failures
+        return Failures(0, 0, encl_failures, ctrl_failures, encl_failures)
+
+    @classmethod
     def build(cls, m0conf: Dict[Oid, Any], parent: Oid,
               pool_desc: Dict[str, Any]) -> Oid:
         assert parent.type is ObjT.root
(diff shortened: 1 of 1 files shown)


cfgen/cfgen
```diff
@@ -1261,6 +1270,31 @@ class ConfPool(ToDhall):
...
         return '{ %s }' % args
 
+    @classmethod
+    def gen_sns_tolerance(cls, m0conf: Dict[Oid, Any],
+                          pool_desc: Dict[str, Any]) -> Failures:
+        encls_nr = len(cluster_encls(m0conf))
+        ctrls_per_encl = floor(len(cluster_ctrls(m0conf)) / encls_nr)
+        data_units = pool_desc['data_units']
+        parity_units = pool_desc['parity_units']
+        spare_units = pool_desc.get('spare_units')
```
knekrasov, line 1280: Potentially, spare_units can be None. On the next line you're summing numbers which is dangerous:

```
Python 3.6.8 (default, Nov 16 2020, 16:55:22)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-44)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> t = None
>>> 2 + t
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
>>> t + 2
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```

```suggestion
        spare_units = pool_desc.get('spare_units', 0)
```

cfgen/cfgen
```diff
@@ -1261,6 +1270,31 @@ class ConfPool(ToDhall):
...
+        return Failures(0, 0, enc_failure_allowed, ctrl_failure_allowed,
+                        parity_units)
+
+    @classmethod
+    def gen_md_tolerance(cls, m0conf: Dict[Oid, Any],
+                         pool_desc: Dict[str, Any]) -> Failures:
+        encls = len(cluster_encls(m0conf))
+        ctrls = len(cluster_ctrls(m0conf))
+        encl_failures = encls - 1
+        ctrl_failures = floor((ctrls/encls))
```
knekrasov, line 1295:
```suggestion
        ctrl_failures = floor(ctrls/encls)
```

knekrasov (changes requested):

cfgen/cfgen
mssawant: Right, thanks.

cfgen/cfgen
mssawant: Fixed.

cfgen/cfgen
```diff
@@ -1261,6 +1270,31 @@ class ConfPool(ToDhall):
...
+        parity_units = pool_desc['parity_units']
+        spare_units = pool_desc.get('spare_units', 0)
+        total_units = data_units + parity_units + spare_units
+        enc_failure_allowed_FP = parity_units / ceil(total_units / encls_nr)
+        enc_failure_allowed = floor(enc_failure_allowed_FP)
+        temp = ctrls_per_encl * enc_failure_allowed
+        ctrl_failure_allowed = min(temp, parity_units)
+        return Failures(0, 0, enc_failure_allowed, ctrl_failure_allowed,
+                        parity_units)
+
```
suvratjoshi, line 1288: For SNS pool :
In my 6 node cluster example : 
I had 2 disks on 5 nodes and 8 disks on the 6th node. In CDF the tolerance values were not mentioned so that it will auto generate values.
It generated tolerance = [0, 0, 1, 1, 4]
Here disks = 4 and CVG =2, it means either we allow 1 CVG to fail or 4 disks to fail.
Lets assume that 2, 2 disks from first 2 nodes fail (where these nodes only had 2 nodes each).
It satisfies 4 disks allowance but its crossing the limit of 1 CVG (as what is generated automatically).

So the formula needs to be changed?

cfgen/cfgen
```diff
@@ -1296,20 +1330,23 @@ class ConfPool(ToDhall):
...
+                                     spare_units=0)
         d = pool_desc.get('allowed_failures')
-        if t in (PoolT.sns, PoolT.dix) and d:
-            tolerance = Failures(d['site'], d['rack'], d['encl'], d['ctrl'],
-                                 d['disk'])
+        if t in (PoolT.sns, PoolT.dix):
+            if d:
+                tolerance = Failures(d['site'], d['rack'], d['encl'],
+                                     d['ctrl'], d['disk'])
+            else:
```
suvratjoshi, line 1343: I am wondering if the else condition at line 1358 which uses "gen_allowances()" is still applicable?

cfgen/cfgen
```diff
@@ -1296,20 +1330,23 @@ class ConfPool(ToDhall):
...
+                if t == PoolT.dix:
+                    tolerance = ConfPool.gen_md_tolerance(m0conf, pool_desc)
```
suvratjoshi, line 1345: Is it a good idea to save new values in 'allowed_failures'?

cfgen/cfgen
```diff
@@ -1261,6 +1270,31 @@ class ConfPool(ToDhall):
...
+        total_units = data_units + parity_units + spare_units
+        enc_failure_allowed_FP = parity_units / ceil(total_units / encls_nr)
+        enc_failure_allowed = floor(enc_failure_allowed_FP)
+        temp = ctrls_per_encl * enc_failure_allowed
+        ctrl_failure_allowed = min(temp, parity_units)
+        return Failures(0, 0, enc_failure_allowed, ctrl_failure_allowed,
+                        parity_units)
+
+    @classmethod
+    def gen_md_tolerance(cls, m0conf: Dict[Oid, Any],
```
suvratjoshi, line 1290: gen_md_tolerance should be renamed to gen_dix_tolerance?

cfgen/cfgen
mssawant: Looks like this is for md or dix pool. Need to check this, I have some idea, probably need to get the minimum from all the nodes.

suvratjoshi: Got input from Madhav for considering this  in case of asymmetric pools :
https://jts.seagate.com/browse/EOS-23178?focusedCommentId=2164896&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2164896

cfgen/cfgen
```diff
@@ -1261,6 +1270,31 @@ class ConfPool(ToDhall):
...
 
+    @classmethod
+    def gen_sns_tolerance(cls, m0conf: Dict[Oid, Any],
+                          pool_desc: Dict[str, Any]) -> Failures:
+        encls_nr = len(cluster_encls(m0conf))
+        ctrls_per_encl = floor(len(cluster_ctrls(m0conf)) / encls_nr)
+        data_units = pool_desc['data_units']
+        parity_units = pool_desc['parity_units']
+        spare_units = pool_desc.get('spare_units', 0)
+        total_units = data_units + parity_units + spare_units
```
suvratjoshi, line 1281: While calculating the pool width, we need to consider this formula here? : 
https://jts.seagate.com/browse/EOS-23178?focusedCommentId=2164896&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2164896

cfgen/cfgen
mssawant: Dix is also meta data thus md.

cfgen/cfgen
mssawant: @suvratjoshi what were the layout parameters (N, K, S) that you used?

suvratjoshi: N+K+S = 14+4+0

mssawant: hmm... so we are tolerating 4 failures. Now in your case with 14 + 4 + 0 and 6 nodes, 5 nodes will host 10 units of a parity group and 1 node will host remaining 4. So even if your 6th node with 8 drives fail, we will only lose maximum 4 units of a parity group, which is tolerable, same with ctrl and with 4 drives. So allowed_failures vector is fine I think.

cfgen/cfgen
mssawant: Yeah, so for now we discard asymmetric configurations which do not satisfy the symmetric pool width >= N + K + S.

mssawant: But this is a separate pool/pver validation issue.

cfgen/cfgen
mssawant: Yeah, can you the generated tolerance vector directly. Fixed.

knekrasov (approved):

cfgen/cfgen
suvratjoshi:
```
Lets assume that 2, 2 disks from first 2 nodes fail (where these nodes only had 2 disks each).
It satisfies 4 disks allowance but its crossing the limit of 1 CVG (as what is generated automatically).
```
I think tolerance disk count should be 2 in this case, so that even if 2 disks fail from a node then it will suffice all these values of allowance vector [0,0,1,1,2]

Now if we consider the pool width formula in case of asymmetric pool then it will be :
C * min(2,2,2,2,2,8) = 6 * 2 = 12

Similarly, tolerance value for disks can be taken as min(2,2,2,2,2,8), which will be 2.

suvratjoshi (approved):

cfgen/cfgen
mssawant: Sorry in the above reply I missed the parity units distribution, yes we need to handle asymettric pools. Have updated the PR with the related code changes. Kindly review.

suvratjoshi:
Hi @mssawant 
I have marked conversations a resolved.
I think we are good to merge this patch.

cfgen/cfgen
```diff
@@ -1275,6 +1312,46 @@ class ConfPool(ToDhall):
                           f'pvers = {oids_to_dhall(self.pvers)}'])
         return '{ %s }' % args
 
+    @classmethod
```
suvratjoshi, line 1315: Need any changes under
Failures.__doc__ = """
?

mssawant: I don't think so, documentation is pretty much generic.

mssawant:
Merging to feature branch of `fault_tolerance` as there are multiple fixes related to the feature that can be tested together.

suvratjoshi:
@mssawant 
DCO is failing
As this is for fault tolerance branch?

mssawant:
@suvratjoshi, Not sure, looks like some problem with Swapnil's username in the commit.
```
Commit sha: 393c336, Author: SwapnilGaonkar7, Committer: GitHub; Expected "SwapnilGaonkar7 68695643+SwapnilGaonkar7@users.noreply.github.com", but got "Swapnil Gaonkar swapnil.gaonkar@seagate.com".
```
But the expected user name looks weird.

mssawant:
DCO check failure is dur to commit,
```
commit 393c336af70f6055047ad556118a459915cf7114
Author: SwapnilGaonkar7 <68695643+SwapnilGaonkar7@users.noreply.github.com>
Date:   Tue Aug 3 15:25:35 2021 +0530

    EOS-23105 Add skeleton class for integration of Hare with HA Event Manager (#1714)

    HA is adding event manager class which can be used to subscribe/register and listen for events.
    Hare needs to add class to access this event manager for listen events

    In this PR, we are adding skeleton class to communicate with Event Manager
    Note: Event Manager class is not fully implemented yet, just skeleton class is present right now

```
```
Commit sha: 393c336, Author: SwapnilGaonkar7, Committer: GitHub; Expected "SwapnilGaonkar7 68695643+SwapnilGaonkar7@users.noreply.github.com", but got "Swapnil Gaonkar swapnil.gaonkar@seagate.com".
```

PR #1743, opened 2021-08-04
EOS-21367 : Facter rpm doesn't exist

vaibhavparatwar:
Solution:
We replace facter with puppet agent

Signed-off-by: vaibhavparatwar <vaibhav.paratwar@seagate.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,11 @@ and health-checking mechanisms.
   sudo yum -y install python3 python3-devel
   ```
 
+* Install puppet-agent (&ge; 6.13.0)
+  ```sh
+  sudo yum localinstall -y https://yum.puppetlabs.com/puppet/el/7/x86_64/puppet-agent-7.0.0-1.el7.x86_64.rpm
+  ```
+
 * Install Consul.
   ```sh
   sudo yum -y install yum-utils
--- a/hare.spec
+++ b/hare.spec
@@ -47,7 +47,7 @@ BuildRequires: python36-pip
 BuildRequires: python36-setuptools
 
 Requires: consul >= 1.7.0, consul < 1.10.0
-Requires: facter >= 3.14.8
+Requires: puppet-agent >= 6.13.0
 Requires: jq
 Requires: cortx-motr = %{h_motr_version}
 Requires: cortx-py-utils


vaibhavparatwar:
@andriytk @daniarherikurniawan  this PR will address issues faced by community for puppet-agent installation and requirements. cc @mssawant @knekrasov

johnbent (approved):

daniarherikurniawan:
Thank you @vaibhavparatwar ! I will test out the patch.

mssawant (approved):

suvratjoshi (approved):

daniarherikurniawan:
Hi @vaibhavparatwar,
I tested this patch on two different cloud providers and it works, but haven't fully solved the problem.
First of all, I want to confirm that the puppet agent is installed flawlessly with your method; and the `facter` at `/opt/puppetlabs/bin/facter` has version > 3.14 as required.

However, the servers (tested at Chameleon and Emulab) were not able to find the `puppet-agent`'s facter because the `hare.spec` failed to update the symlink. Currently, in the`hare.spec` there is this condition:
``` sh 
    if [[ ! -e /usr/bin/facter && -e /opt/puppetlabs/bin/facter ]] ; then
        ln -vsf /opt/puppetlabs/bin/facter /usr/bin/facter
```
In my servers, that condition is `false` because the first condition is not satisfied, thus the `facter` still points to the original facter (version 2.4.1)  instead of the `puppet-agent`'s facter (version 4.0.46).
Therefore, I am suggesting that we remove the ` ! -e /usr/bin/facter` checking and just have these instead:
``` sh 
    if [[ -e /opt/puppetlabs/bin/facter ]] ; then
        ln -vsf /opt/puppetlabs/bin/facter /usr/bin/facter
```

By doing so, it will always use puppet-agent's facter no matter what. This solution solved my problems on Emulab and Chameleon cloud. 
Note: the `facter` version 2.4.1 is installed automatically without me running any `facter` specific installation, I don't really investigate precisely when this facter got installed.

Please let me know if you have any questions. Thanks! 
cc: @mssawant @suvratjoshi

vaibhavparatwar:
> Hi @vaibhavparatwar,
> I tested this patch on two different cloud providers and it works, but haven't fully solved the problem.
> First of all, I want to confirm that the puppet agent is installed flawlessly with your method; and the `facter` at `/opt/puppetlabs/bin/facter` has version > 3.14 as required.
> 
> However, the servers (tested at Chameleon and Emulab) were not able to find the `puppet-agent`'s facter because the `hare.spec` failed to update the symlink. Currently, in the`hare.spec` there is this condition:
> 
> ```shell
>     if [[ ! -e /usr/bin/facter && -e /opt/puppetlabs/bin/facter ]] ; then
>         ln -vsf /opt/puppetlabs/bin/facter /usr/bin/facter
> ```
> 
> In my servers, that condition is `false` because the first condition is not satisfied, thus the `facter` still points to the original facter (version 2.4.1) instead of the `puppet-agent`'s facter (version 4.0.46).
> Therefore, I am suggesting that we remove the ` ! -e /usr/bin/facter` checking and just have these instead:
> 
> ```shell
>     if [[ -e /opt/puppetlabs/bin/facter ]] ; then
>         ln -vsf /opt/puppetlabs/bin/facter /usr/bin/facter
> ```
> 
> By doing so, it will always use puppet-agent's facter no matter what. This solution solved my problems on Emulab and Chameleon cloud.
> Note: the `facter` version 2.4.1 is installed automatically without me running any `facter` specific installation, I don't really investigate precisely when this facter got installed.
> 
> Please let me know if you have any questions. Thanks!
> cc: @mssawant @suvratjoshi

@daniarherikurniawan  Thank you for your inputs. We will handle it in separate PR and this PR we will merge with current changes. cc @mssawant

PR #1745, opened 2021-08-05
EOS-15788: Node failure events handling

knekrasov:
JIRA ticket: [EOS-15788](https://jts.seagate.com/browse/EOS-15788)

This PR adds integration with HA EventManager. By the result of this integration node state events ("offline", "online") will be received by hax and translated into Motr-specific broadcast messages.

Partial problems resolved by this PR:
1. No dev environment to debug against HA component. As a solution, the HA simulator is implemented (see `ha-simulator/` folder for more details). Hare developers don't need to deploy Kafka, Zookeeper, configure MessageBus etc to trigger HA events to happen.
2. The task itself leads to coupling with cortx-ha component (Reason: in order to receive HA events, Hare side needs to import cortx-ha types like EventManager, SubscribeEvent etc). The proposed solution identifies such issues and handles them gracefully (so Hare will be built and ran even in the environments without cortx-ha support).

### DIFF
--- a/hax/hax/ha/handler/node.py
+++ b/hax/hax/ha/handler/node.py
@@ -1,0 +1,64 @@
+# Copyright (c) 2021 Seagate Technology LLC and/or its Affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+import logging
+
+from hax.ha.events import Event
+from hax.ha.handler import EventHandler
+from hax.message import BroadcastHAStates
+from hax.motr.planner import WorkPlanner
+from hax.types import HAState, ServiceHealth
+from hax.util import ConsulUtil
+
+LOG = logging.getLogger('hax')
+
+
+class NodeEventHandler(EventHandler):
+    """
+    Handles HA events with resource_type == 'node'.
+
+    As a result of such HA event the corresponding BroadcastHAStates is
+    offered to WorkPlanner.
+    """
+    def __init__(self, consul: ConsulUtil, planner: WorkPlanner):
+        """Constructor."""
+        self.cns = consul
+        self.planner = planner
+
+    def handle(self, msg: Event) -> None:
+        node_fid = self.cns.get_node_fid(msg.node_id)
+        if not node_fid:
+            LOG.warn('Unknown [node_id=%s] provided. HA event is ignored',
+                     msg.node_id)
+            return
+
+        get_health = self._get_status_by_text
+
+        self.planner.add_command(
+            BroadcastHAStates(states=[
+                HAState(fid=node_fid, status=get_health(msg.event_type))
+            ],
+                              reply_to=None))
+
+    def _get_status_by_text(self, status: str):
+        if status == 'online':
+            return ServiceHealth.OK
+        elif status == 'offline':
+            return ServiceHealth.FAILED
+        else:
+            return ServiceHealth.UNKNOWN
--- a/hax/hax/ha/thread.py
+++ b/hax/hax/ha/thread.py
@@ -1,0 +1,116 @@
+# Copyright (c) 2021 Seagate Technology LLC and/or its Affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+#
+
+import logging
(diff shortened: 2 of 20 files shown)


knekrasov:
Note to myself: Seems like this PR has to have another target (fault_tolerance branch instead of main).

SwapnilGaonkar7:
@knekrasov Can you add some details about how you tested event handling(sender, receiver)

knekrasov:
@SwapnilGaonkar7 et al: I've added README.md file. Please [have a look](https://github.com/Seagate/cortx-hare/blob/7400a9cd6c3a97508fb7648f260836ce06256781/ha-simulator/README.md).

vaibhavparatwar:
> @SwapnilGaonkar7 et al: I've added README.md file. Please [have a look](https://github.com/Seagate/cortx-hare/blob/7400a9cd6c3a97508fb7648f260836ce06256781/ha-simulator/README.md).

Thanks @knekrasov . @pavankrishnat this may be useful in case we want to start testing without HA code availability.

hax/hax/ha/handler/node.py
```diff
@@ -0,0 +1,64 @@
...
+        node_fid = self.cns.get_node_fid(msg.node_id)
+        if not node_fid:
+            LOG.warn('Unknown [node_id=%s] provided. HA event is ignored',
+                     msg.node_id)
+            return
+
+        get_health = self._get_status_by_text
+
+        self.planner.add_command(
+            BroadcastHAStates(states=[
```
SwapnilGaonkar7, line 53: This will cause 'broadcast_ha_states' to execute right?
If yes then in that function we need to check if node failure is getting broadcasted(st.fid.container == ObjT.NODE.value) and we should call 'notify_node_status'.
Please correct me if I am wrong.

Shreya-18:
@mssawant I have a question,
When we receive a node failure event from HA, do we get process failure events for the processes running on that node along with it? 
Or Are the process statuses in consul kv updated?

vaibhavparatwar:
@knekrasov please address codacy issues.

hax/hax/ha/handler/node.py
knekrasov: @SwapnilGaonkar7 according to what I see in the code, it doesn't make much sense.

```python
    def notify_node_status(self,
                           proc_note: HaNoteStruct) -> List[HaNoteStruct]:
        new_state = proc_note.no_state
        proc_fid = Fid.from_struct(proc_note.no_id)
        assert ObjT.PROCESS.value == proc_fid.container
        LOG.debug('Notifying node status for process_fid=%s state=%s',
                  proc_fid, new_state)

        node = self.consul_util.get_process_node(proc_fid)

        node_fid = self.consul_util.get_node_fid(node)
        encl_fid = self.consul_util.get_node_encl_fid(node)
        ctrl_fids = self.consul_util.get_node_ctrl_fids(node)
        LOG.debug('node_fid: %s encl_fid: %s ctrl_fids: %s with state: %s',
                  node_fid, encl_fid, ctrl_fids, new_state)

        notes = []
        if node_fid and encl_fid:
            notes = [HaNoteStruct(no_id=x.to_c(), no_state=new_state)
                     for x in [node_fid, encl_fid]]

        if ctrl_fids:
            for x in ctrl_fids:
                notes.append(HaNoteStruct(no_id=x.to_c(), no_state=new_state))

        return notes
```

Facts:
1. This function takes a PROCESS fid, learns the node where this process runs and adds the node state to the list of `HaNoteSruct` to report to Motr layer.
2. The function doesn't update node status in Consul KV anyway.

In my case I don't have a process. Instead, I have the node FID already so this FID is included into the HaNoteStructs that will be broadcast. Do you think I should do anything else?

SwapnilGaonkar7: notify_node_status reports encl, ctrls status as failed(We are calling it when all IO services are down)
I think it should also be called when we receive node failure event.

Shreya-18: > This function takes a PROCESS fid, learns the node where this process runs and adds the node state to the list of HaNoteSruct to report to Motr layer.

Correct, and in case of the node failure event, we won't receive any process failure events. So which process fid should we pass here? Maybe we need to update the function params to receive the node and state.

knekrasov: Yes, I'll need to refactor `notify_node_status` function then. Working on that..

Shreya-18: > The notify_node_status() doesn't update node status in Consul KV anyway.
as part of EOS-23005, I am adding update statuses for node, enclosure, and controllers in consul KV, in this function.

SwapnilGaonkar7: Looks good

SwapnilGaonkar7 (approved):

hax/hax/ha/handler/node.py
knekrasov: Addressed

hax/hax/ha/handler/node.py
```diff
@@ -0,0 +1,64 @@
...
+                HAState(fid=node_fid, status=get_health(msg.event_type))
+            ],
+                              reply_to=None))
+
+    def _get_status_by_text(self, status: str):
```
mssawant, line 58: return type hint? ServiceHealth?

hax/hax/ha/thread.py
```diff
@@ -0,0 +1,122 @@
...
+from hax.motr.planner import WorkPlanner
+from hax.types import StoppableThread
+from hax.util import ConsulUtil, InterruptedException, wait_for_event
+
+from ha.core.event_manager.subscribe_event import SubscribeEvent
+
+LOG = logging.getLogger('hax')
+
+
+class EventPollingThread(StoppableThread):
```
mssawant, line 35: Instead of polling, did we check if there's a callback mechanism with message bus where a callback function is invoked on whenever event is posted to the subscribed topic?

mssawant (approved):
Looks good. Few minor comments. It would be good if you post some locally tested results to the PR.

hax/hax/ha/thread.py
knekrasov: [python-kafka](https://kafka-python.readthedocs.io/en/master/apidoc/KafkaConsumer.html) doesn't seem to provide this kind of interface. But probably @SwapnilGaonkar7 knows better as he is the author of EventListener

mssawant:
@knekrasov , Merging this PR for now, let's note the comments and fix them before landing to main.

PR #1746, opened 2021-08-06
EOS-23252: Disable D212 pattern for pep257 tool for Codacy

mukul-seagate11:
Problem that was bumped into: https://github.com/Seagate/cortx-hare/pull/1725#issuecomment-889066153
where both D212 and D213 warnings are enabled in Codacy (https://github.com/PyCQA/pydocstyle/issues/475) but they are mutually exclusive which makes the situation more curious is that this is about position of the first line in a multiline docstring as,

image

### DIFF
--- a/prospector.yml
+++ b/prospector.yml
@@ -1,0 +1,18 @@
+output-format: json
+
+strictness: medium
+test-warnings: true
+doc-warnings: false
+member-warnings: false
+inherits:
+  - default
+ignore-paths:
+  - docs
+ignore-patterns:
+  - (^|/)skip(this)?(/|$)
+autodetect: true
+max-line-length: 88
+
+pep257:
+  disable:
+    - D212


mssawant:
@mukul-seagate11 , it would be good if you post a link for the reference of D212 and D213 pattern for pep257 or few lines describing what it means by D212 and D213 pattern from the actual documentation in the commit and PR description.
It would help to understand what the problem is and how we are trying to solve it.

Please refer to https://github.com/torvalds/subsurface-for-dirk/blob/a48494d2fbed58c751e9b7e8fbff88582f9b2d02/README#L88, we try to follow the same.

mukul-seagate11:
> @mukul-seagate11 , it would be good if you post a link for the reference of D212 and D213 pattern for pep257 or few lines describing what it means by D212 and D213 pattern from the actual documentation in the commit and PR description.
> It would help to understand what the problem is and how we are trying to solve it.
> 
> Please refer to https://github.com/torvalds/subsurface-for-dirk/blob/a48494d2fbed58c751e9b7e8fbff88582f9b2d02/README#L88, we try to follow the same.

you can refer https://github.com/Seagate/cortx-hare/pull/1725 for more details

suvratjoshi:
> you can refer #1725 for more details
Hi @mukul-seagate11 
Sorry but I am not able to figure out what kind of codacy false positive we are trying to resolve here.
I even referred this mentioned PR but I could not figure this out.
The description can have some information related to patterns and pep257 tool as well.

May be something like this example :
Pattern xyz : Variable declared but not used

It would be nice to have all details present in the commit message and PR description so that later in the git log history, we clearly come to know the problem and the solution.

Thank you!

mssawant:
>you can refer #1725 for more details

@mukul-seagate11, every PR must be self explanatory. You can of course add a reference to some other occurrence to an event or documentation as an example, but in general as a reader I should get an idea about the problem and the solution looking at the commit message and the PR description.

mukul-seagate11:
@knekrasov @vaibhavparatwar, please approve & merge the PR as the issue was reported by you

mukul-seagate11:
@mssawant, description updated as problem reported by Hare

PR #1767, opened 2021-08-25
EOS-23639: Generate C coverage data in xml form for coverage tool.

ajaykumarptl:
Need `C` Code coverage data in form of xml to give input to Codacy code coverage tool.

Solution:
Implemented `C` coverage class into python module for coverage data creation in xml
form.
NOTE: For `C` coverage `hax` binary should be executed. As `hax` binary is not executed
with unit tests this script will not be able to generate coverage data.

Signed-off-by: ajaykumarptl <ajaykumar.patel@seagate.com>


cortx-admin:

Makefile
```diff
@@ -488,7 +489,15 @@ uninstall:
...
 .PHONY: check
-check: check-cfgen check-hax check-miniprov flake8 mypy
+check: check-cfgen check-hax check-miniprov flake8 mypy gcov
+
+.PHONY: gcov
+gcov:
+	@$(call _info, Checking gcov package.)
+	@if [ -z $(COV_BIN) ]; then \
+	$(call _log, installing gcov); \
+		yum install -y lcov; \
```
knekrasov, line 499: I don't think it is a good practice. Makefile CAN suggest which packages are needed but it should not install them. The user may install gcov in some other ways than running `yum` (`make install`? snap? flatpak? nix?)

hax/setup.py
```diff
@@ -174,7 +174,9 @@ def is_runtime_comment(line: str) -> bool:
             define_macros=[('M0_INTERNAL', ''), ('M0_EXTERN', 'extern')],
             library_dirs=[get_motr_libs_dir()],
             runtime_library_dirs=[get_motr_libs_dir()],
-            libraries=['motr'],
-            extra_compile_args=[x for x in get_motr_cflags() + ['-fPIC']])
+            libraries=['motr', 'gcov'],
```
knekrasov, line 177: What will be the result of these `gcov` and `-ftest-coverage`?

Will it instrument the c code during the build? If so, is it a good idea to instrument the code always (even in production)? Maybe it is a better idea to enable C coverage by some external flag or environment variable?

utils/hare_cov/README.md
```diff
@@ -13,13 +13,19 @@
 5.  Run the `hare_coverage.py`. It will generate the `xml` report inside 
     `hare_coverage_report/coverage.xml`.
 
+NOTE: For `C` coverage `hax` binary should be executed. As `hax` binary is not executed
```
knekrasov, line 16: Not sure that it works like it is stated. `hax` binary is not the point.

The fact is that we have `hax.c` file. Python code interacts with `hax.c` and with `libmotr` via FFI layer. In order to make C coverage work, the execution path should go into FFI layer. That layer is intentionally mocked in all automated tests. In other words, we don't have any test that would depend on real `libmotr`. Perhaps that can be done when we implement system tests (currently we have no such tests).

Makefile
ajaykumarptl: Added log message. Removed installation part.

hax/setup.py
ajaykumarptl: Yes, it instrument the c code during the build. Added environment variable.

utils/hare_cov/README.md
ajaykumarptl: I am referring systemd hare-hax service which will call `/opt/seagate/cortx/hare/bin/hax`.  While doing bootstrap hare-hax service started and thus `hax` and that will create `gcno` for hax.c which will be used for the coverage calculation. I don't have much knowledge about but this is my observation.

Yes, It will work for system tests. In future when we implement system tests then we just need to run that for actual coverage calculation till that this script will create 0 coverage output.

vaibhavparatwar:
Closing this PR .. currently this task is not prioritized

PR #1772, opened 2021-08-27
Problem: get_service_health is hard to read

knekrasov:
The decision-making dict in this function is VERY wordy, the identifiers
and function names are very long, as a result the structure of the
mapping is rather hard to read. Example of just 1 mapping:

```
            cur_consul_status('passing', 'M0_CONF_HA_PROCESS_STARTING'):
            local_remote_health_ret(ServiceHealth.OFFLINE,
                                    ServiceHealth.OK),
```

Solution: shorten the type classifiers (ServiceHealth.OFFLINE ->
S.OFFLINE), shorten the utility function names so that every mapping
fits into 1 line.

### DIFF
--- a/hax/hax/util.py
+++ b/hax/hax/util.py
@@ -1098,80 +1098,64 @@ class ConsulUtil:
         # {(consul_svc_status, motr_process_status):(local_node_ha_status,
         #                                            remote_node_ha_status)}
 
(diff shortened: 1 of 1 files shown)


cortx-admin:

vaibhavparatwar:
@mssawant  do we need this PR now?

PR #1651, opened 2021-06-17
EOS-16272: cfgen to generate aux pool layouts considering node failures

suvratjoshi:
This feature is developed considering the node (enclosure) failure
case. If a node in the cluster fails, it means that all the disks
in that node become inaccessible. As this feature considers only
node failures, it keeps the spare pools ready considering the
other alive nodes from the cluster.
The feature works using the mathematical combinations formula
C(n, r) = (n! / (r!(n-r)!))
It basically creates combinations of all the nodes in case of
failures of nodes from 1 to the allowed number of nodes to be
failed, by allowance vector.
For each and every such generated combinations of (alive) nodes,
it creates the sets of the disks and assigns a pool to each of
them. Such spare pools created considering allowed node failures
are called as aux pools.

This PR also covers https://github.com/Seagate/cortx-hare/issues/1742

Signed-off-by: Suvrat Joshi <suvrat.joshi@seagate.com>

PR #1773, opened 2021-08-27
Problem: EOS-24186 S3 Mini Provisioner job is failing on main during bootstrap cluster

SwapnilGaonkar7:
'aux_pool_list' function tries to access 'disk_refs' without checking if
it is present or not. cfgen do work without specifying 'disk_refs' in pool.
In above case we are considering all disks as part of pool.

Solution: Modified above mentioned function to list all the devices mentioned in CDF
file and use them if 'disk_refs' is not specified

### DIFF
--- a/cfgen/cfgen
+++ b/cfgen/cfgen
@@ -21,6 +21,17 @@ from math import floor, ceil
 import yaml
 import logging
 
+from dataclasses import dataclass
+
+
+# This class will hold info for disks mentioned in CDF file
+# One use case is to generate list of all disks present in CDF
+# file
+@dataclass
+class DiskRef():
+    node: str
+    path: str
+
 
 __version__ = '0.14'
 
@@ -1929,6 +1940,12 @@ def validate_m0conf(m0conf: Dict[Oid, Any]) -> None:
                for _rel, children in rel_children.items())
 
 
+# This function will return list of disks present in provided pool
+# It will return None if 'disk_refs' is not present in pool
+def get_pool_diskrefs(pool_desc: Dict[str, Any]):
+    return pool_desc.get('disk_refs')
+
+
 # AUX Pools :
 # This feature is developed considering the node (enclosure) failure
 # case. If a node in the cluster fails, it means that all the disks
@@ -1953,7 +1970,9 @@ def validate_m0conf(m0conf: Dict[Oid, Any]) -> None:
 # Outputs : None
 # Saves generated aux pools list in aux_pools[]
 
-def aux_pool_list(pool_desc: Dict[str, Any], m0conf: Dict[Oid, Any]) -> None:
+def aux_pool_list(pool_desc: Dict[str, Any],
+                  disks: List[DiskRef],
+                  m0conf: Dict[Oid, Any]) -> None:
     pool = pool_desc
 
     # Example of encl_ref_combination :
@@ -2004,9 +2023,16 @@ def aux_pool_list(pool_desc: Dict[str, Any], m0conf: Dict[Oid, Any]) -> None:
     # Generate the node combinations considering node failures as
     # allowed by node tolerance values
     temp_pool_nodes = []
-    for nodes in pool['disk_refs']:
-        temp_pool_nodes += [
-            nodes['node']]
+
+    disk_refs = get_pool_diskrefs(pool)
+    if disk_refs is None:
+        for disk in disks:
+            temp_pool_nodes += [
+                disk.node]
+    else:
+        for nodes in disk_refs:
+            temp_pool_nodes += [
+                nodes['node']]
     pool_nodes = list(set(temp_pool_nodes))
     for encl_failures in range(tolerance.enclosure, 0, -1):
         encl_ref_combination += [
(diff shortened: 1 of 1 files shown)


cfgen/cfgen
```diff
@@ -1953,7 +1953,9 @@ def validate_m0conf(m0conf: Dict[Oid, Any]) -> None:
 # Outputs : None
 # Saves generated aux pools list in aux_pools[]
 
-def aux_pool_list(pool_desc: Dict[str, Any], m0conf: Dict[Oid, Any]) -> None:
+def aux_pool_list(pool_desc: Dict[str, Any],
+                  disks: List,
```
knekrasov, line 1957: `List[type]` is required.

cfgen/cfgen
```diff
@@ -2082,7 +2087,18 @@ def build_cluster(cluster_desc: Dict[str, Any]) -> Cluster:
         pool_t = pool_type(pool)
         if pool_t == PoolT.sns:
             # For SNS pools create the aux pool list
-            aux_pool_list(pool, conf)
+            disks: List = []
```
knekrasov, line 2090: Type parameter for `List` is missing.

knekrasov:
@suvratjoshi Isn't it the same problem you addressed last week in your PR #1651?

cfgen/cfgen
SwapnilGaonkar7: Done

cfgen/cfgen
SwapnilGaonkar7: Done

knekrasov (approved):

cfgen/cfgen
```diff
@@ -2004,9 +2014,16 @@ def aux_pool_list(pool_desc: Dict[str, Any], m0conf: Dict[Oid, Any]) -> None:
     # Generate the node combinations considering node failures as
     # allowed by node tolerance values
     temp_pool_nodes = []
-    for nodes in pool['disk_refs']:
-        temp_pool_nodes += [
-            nodes['node']]
+
+    disk_refs = pool.get('disk_refs')
```
mssawant, line 2018: ` aux_pool_list()` has grown big and we are compiling `disk_refs` at-least twice in this patch. Please create a function to get disk refs from pool_desc. Accept pool_desc and a parameter.

SwapnilGaonkar7: Done

mssawant (approved):

cfgen/cfgen
```diff
@@ -2082,7 +2110,18 @@ def build_cluster(cluster_desc: Dict[str, Any]) -> Cluster:
         pool_t = pool_type(pool)
         if pool_t == PoolT.sns:
             # For SNS pools create the aux pool list
-            aux_pool_list(pool, conf)
+            disks: List[DiskRef] = []
+            disk_refs = get_pool_diskrefs(pool)
```
knekrasov, line 2114: disk_refs is used in `if disk_refs is None` only. Is that correct?

So if `disk_refs` is not None, then the value is ignored.

SwapnilGaonkar7: Yes, correct. We need to generate disk_refs only if it is not present in pool description

cfgen/cfgen
```diff
@@ -2004,9 +2018,16 @@ def aux_pool_list(pool_desc: Dict[str, Any], m0conf: Dict[Oid, Any]) -> None:
     # Generate the node combinations considering node failures as
     # allowed by node tolerance values
     temp_pool_nodes = []
-    for nodes in pool['disk_refs']:
-        temp_pool_nodes += [
-            nodes['node']]
+
+    disk_refs = get_pool_diskrefs(pool)
+    if disk_refs is None:
```
knekrasov, line 2023:
```suggestion
    # We need to generate disk_refs only if it is not present in pool description
    if not get_pool_diskrefs(pool):
```

SwapnilGaonkar7: @knekrasov  Here, disk_refs  is used after checking so this changes is not needed.
I have applied your suggestion at previous comment

knekrasov: Ah, correct, I put the suggestion to the wrong line. Apologies :-)

cfgen/cfgen
```diff
@@ -1929,6 +1937,10 @@ def validate_m0conf(m0conf: Dict[Oid, Any]) -> None:
                for _rel, children in rel_children.items())
```
suvratjoshi, line 1939: Function header

cfgen/cfgen
```diff
@@ -21,6 +21,14 @@ from math import floor, ceil
 import yaml
 import logging
 
+from dataclasses import dataclass
+
+
```
suvratjoshi, line 26: Documentation comment can be added here

suvratjoshi (approved):

cfgen/cfgen
SwapnilGaonkar7: Done

cfgen/cfgen
SwapnilGaonkar7: Done

SwapnilGaonkar7:
3 node deployment passed http://eos-jenkins.mero.colo.seagate.com/job/Cortx-Deployment/job/VM-Deployment-R2-3Node/832/console

PR #1778, opened 2021-09-01
EOS-24266: Drive states are still broadcast when MKFS is down

knekrasov:
The fix for EOS-24124 was incomplete: broadcast to Motr layer still
happened even in case of MKFS stopped state (the condition was around
Consul KV update only).

Solution:
1. Exclude drive status broadcast when MKFS is not online
2. Implement an automatic test that validates this scenario

Signed-off-by: Konstantin Nekrasov <konstantin.nekrasov@seagate.com>

### DIFF
--- a/hax/hax/motr/__init__.py
+++ b/hax/hax/motr/__init__.py
@@ -402,16 +402,20 @@ class Motr:
         new_state = note.no_state
         proc_fid = Fid.from_struct(note.no_id)
 
-        for svc in services:
-            disk_list += cns.get_disks_by_parent_process(proc_fid, svc.fid)
+        state = (ServiceHealth.OK if new_state ==
+                 HaNoteStruct.M0_NC_ONLINE else ServiceHealth.OFFLINE)
+        is_mkfs = self._is_mkfs(proc_fid)
+
+        mkfs_down = is_mkfs and state != ServiceHealth.OK
+
+        if not mkfs_down:
+            for svc in services:
+                disk_list += cns.get_disks_by_parent_process(proc_fid, svc.fid)
         if disk_list:
-            state = (ServiceHealth.OK if new_state ==
-                     HaNoteStruct.M0_NC_ONLINE else ServiceHealth.OFFLINE)
             # XXX: Need to check the current state of the device, transition
             # to ONLINE only in case of an explicit request or iff the prior
             # state of the device is UNKNOWN/OFFLINE.
-            is_mkfs = self._is_mkfs
-            if not (state == ServiceHealth.STOPPED and is_mkfs(proc_fid)):
+            if not mkfs_down:
                 # We don't mark the devices as failed if the process is MKFS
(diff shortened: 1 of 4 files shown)


hax/test/integration/test_motr.py
```diff
@@ -388,6 +417,134 @@ def my_get(key: str, recurse: bool = False):
...
+    # import pudb.remote
+    # pudb.remote.set_trace(term_size=(130, 50), port=9998)
+    motr.broadcast_ha_states([HAState(fid=Fid(0x7200000000000001, 0x15),
+                                      status=ServiceHealth.FAILED)])
+
+    assert not consul_util.update_drive_state.called, \
+        'The drive state should not be updated when MKFS stops'
+
+    traces = motr._ffi.traces
+    assert AssertionPlan(
```
mssawant, line 542: @knekrasov, This whole mechanism is amazing! using wrapper function to capture the function args and using them for verification. We have a similar mechanism in Motr which is used to trace function calls. I think we can further extend this tracing mechanism to trace entry and exit of a function. I think we should take it out of the `test` dir into main hax sources.
Such mechanism will help us trace the performance issues like client slow start. What do you think?

mssawant (approved):
Looks great! just few minor style comments.

hax/hax/motr/__init__.py
```diff
@@ -402,16 +402,20 @@ def _generate_sub_disks(self, note: HaNoteStruct,
         new_state = note.no_state
         proc_fid = Fid.from_struct(note.no_id)
 
-        for svc in services:
-            disk_list += cns.get_disks_by_parent_process(proc_fid, svc.fid)
+        state = (ServiceHealth.OK if new_state ==
+                 HaNoteStruct.M0_NC_ONLINE else ServiceHealth.OFFLINE)
+        is_mkfs = self._is_mkfs
+
+        mkfs_down = is_mkfs and state != ServiceHealth.OK
```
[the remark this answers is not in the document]
knekrasov: Yes, mkfs process is down.

hax/test/integration/test_motr.py
knekrasov: @mssawant There are two aspects:

1. For 'normal' Pythonic code things like standard `Mock` will work better (one can check whether the function was called, the returned value can either be mocked or a real function can be invoked, but the invocation can still [be spied on](https://wesmckinney.com/blog/spying-with-python-mocks/)). The solution is fine given it is something that works out of the box.
2. I had to implement this `@trace_call` and `AssertionPlan` mini framework because HaxFFI is very hard to mock using standard methods (reason: it is stuffed with ctypes-based functions). And one more item - I wanted to be able to 'query' for HaxFFI calls and see the sequence of the invocations (given it is not a unit test, the amount of calls that can happen is rather large and we are interested in the order); this functionality is not provided by standard Mock.

So long story short, I'm glad you liked the approach and will be happy to improve it. But I'm not sure that we'll want to use it everywhere 😁

Issue #1682, opened 2021-06-25
Client application startup is very slow


andriytk:
On the SAGE cluster with 23 nodes (7 server and 16 client nodes with 16 m0_client processes configured each) the startup of any client application (c0cp or mcp) takes about 20-30 seconds. And the more m0_client processes are configured - the more time it takes to start up. (That's why the number of configured m0_client processes is limited to 16 atm.)

See attached [m0trace.19259.txt.zip](https://github.com/Seagate/cortx-motr/files/5835742/m0trace.19259.txt.zip) from the client run.

andriytk:
@mssawant, @knekrasov any idea why `Preparing the reply for HaNvecGetEvent` takes so much time?

knekrasov:
@andriytk 
> any idea why Preparing the reply for HaNvecGetEvent takes so much time?

I think the reason is that this preparation caused more than 3000 separate read requests to Consul KV.

Here are some details:

This is the handler method:
```python
   @log_exception
    def ha_nvec_get_reply(self, event: HaNvecGetEvent) -> None:
        LOG.debug('Preparing the reply for HaNvecGetEvent (nvec size = %s)',
                  len(event.nvec))
        notes: List[HaNoteStruct] = []
        for n in event.nvec:
            n.note.no_state = self.consul_util.get_conf_obj_status(
                ObjT[n.obj_t], n.note.no_id.f_key)
            notes.append(n.note)

        LOG.debug('Replying ha nvec of length ' + str(len(event.nvec)))
        self._ffi.ha_nvec_reply(event.hax_msg, make_array(HaNoteStruct, notes),
                                len(notes))
```
According to the log you provided, the nvec length is 1024. So the loop above had 1024 iterations; `self.consul_util.get_conf_obj_status()` was invoked that many times.

If we go down to `get_conf_obj_status()`, we'll see 2 explicit KV reads there: 
```python
@repeat_if_fails()
    def get_conf_obj_status(self, obj_t: ObjT, fidk: int) -> int:
        obj_state: int = HaNoteStruct.M0_NC_ONLINE
        if obj_t.name in (ObjT.PROCESS.name, ObjT.SERVICE.name):
            # 'node/<node_name>/process/<process_fidk>/service/type'
            node_items = self.kv.kv_get('m0conf/nodes', recurse=True)
            # TODO [KN] This code is too cryptic. To be refactored.
            keys = getattr(self,
                           'get_{}_keys'.format(
[...]

PR #1737, opened 2021-08-03
Problem: hax makes multiple Consul kv recurse calls

mssawant:
In context of handling various Hare and Motr events, e.g. BroadcastHAStates,
Motr ProcessEvent, Motr nvec requests, Motr entrypoint requests, etc. Hax
makes multiple recurse calls to Consul KV not just one per event but multiple
calls per event as well. This creates multiple such http requests and also
affects the overall performance.
In a large cluster comprising of multiple clients and server this issue can
be seen easily where a motr client takes several miliseconds to start.
(see https://github.com/Seagate/cortx-hare/issues/1682)

Solution:
Cache result of such ConsulKV requests in thread local storage and refer to
the same instead of making multiple Consul KV calls over http.
Invalidate the cache in case of the relevant KV update.

Addresses: https://github.com/Seagate/cortx-hare/issues/1682
Addresses: https://jts.seagate.com/browse/EOS-22148

Signed-off-by: Mandar Sawant <mandar.sawant@seagate.com>

PR #1780, opened 2021-09-06
Basic mechanism for Consul KV caching

knekrasov:
Problem: a business operation triggers too may KV reads

Solution: Introduce basic cache mechanism

Basic ideas of the cache:
1. The notion of business operation is introduced.
2. The cache itself is not shared between different threads.
3. Lifetime of the cache may be shorter than a single command handling.
4. When active, cache accumulates the mapping between the input arguments and the result values of the methods decorated by `@uses_consul_cache` decorator.
5. Business operation starts at a frame that invokes `@uses_consul_cache` function and when it DOES NOT provide kv_cache. Such invocation creates the cache instance that is used while evaluating the given business operation. The reference to this cache instance is put (implicitly) at the frame that INVOKED such a business operation.
6. If another business operation is invoked from the same frame (e.g. when the operation is invoked in a for-loop), that implicit cache value will be used.
7. When frame is destroyed, the implcit cache is destroyed also (just because the cache is stored as a local variable).
8. Any function which is decorated with `@invalidates_consul_cache`, clears the current (implicit) cache if it exists.

Addresses to #1682


vaibhavparatwar (approved):

vaibhavparatwar:
@knekrasov  please resolve codacy issues if possible

andriytk:
Test result with 200 m0_client processes configured in the CDF:

```
    m0_clients:
      s3: 0         # number of S3 servers to start
      other: 200      # max quantity of other Motr clients this host may have
```

The startup time is about 7.5 secs (vs [20 secs](https://github.com/Seagate/cortx-hare/pull/1737#issuecomment-900260191) without the patch):

```
# time ./mcp -ep 192.168.180.176@tcp:12345:4:1 -prof 0x7000000000000001:0x2a1 -proc 0x7200000000000001:0x29 -hax 192.168.180.176@tcp:12345:1:1 -v /tmp/1m 0x12345:0x12304
2021/09/15 15:10:15 mio.go:542: W: off=0 len=1048576 bs=1048576 gs=1048576 speed=99 (Mbytes/sec)

real	0m7.540s
user	0m0.193s
sys	0m0.180s
```

So the patch helps, similar to another patch at https://github.com/Seagate/cortx-hare/pull/1737. But it does not solve the problem. (To solve it completely, the startup time should be less than 1 sec and not depend on the cluster configuration.)

pavankrishnat:
**component QA Testing done :**

Custom build created successfully :
http://eos-jenkins.colo.seagate.com/job/GitHub-custom-ci-builds/job/centos-7.9/job/custom-ci/2084/

RPMs at location:
http://cortx-storage.colo.seagate.com/releases/cortx/github/integration-custom-ci/centos-7.9.2009/custom-build-2084/

**3-Node** deployment successful :
http://eos-jenkins.colo.seagate.com/job/Cortx-Deployment/job/VM-Deployment-3Node-CentOS-7.9/111/

Tested with default CDF and customize CDF (**m0_clients other: 240 and s3: 1**)

With both CDFs, did Bootstrap-shutdown, hctl status, m0cp write, m0cat read and dgmode Read (by creating Node Fail)

cmd output with customize CDF:
```
[root@ssc-vm-g2-rhev4-1052 ~]# time dd if=/dev/urandom of=/root/write_file_other_240 bs=1048576 count=48
48+0 records in
48+0 records out
50331648 bytes (50 MB) copied, 0.36622 s, 137 MB/s

real    0m0.368s
user    0m0.000s
sys     0m0.368s
[root@ssc-vm-g2-rhev4-1052 ~]# time m0cp -l 192.168.93.26@tcp:12345:4:22 -H 192.168.93.26@tcp:12345:1:1 -p 0x7000000000000001:0x98f -P 0x7200000000000001:0x65 -L 9 -s 4096 -c 48 -o 0x12345:0x12304 /root/write_file_other_240

real    0m47.601s
user    0m0.801s
sys     0m0.453s
[root@ssc-vm-g2-rhev4-1052 ~]# time m0cat -l 192.168.93.26@tcp:12345:4:22 -H 192.168.93.26@tcp:12345:1:1 -p 0x7000000000000001:0x98f -P 0x7200000000000001:0x65 -L 9 -s 4096 -c 48 -o 0x12345:0x12304 /root/read_Before_file_other_240
pver is : <0:0>

real    0m48.859s
user    0m0.832s
sys     0m0.425s

[root@ssc-vm-g2-rhev4-1052 ~]# time m0cat -l 192.168.93.26@tcp:12345:4:22 -H 192.168.93.26@tcp:12345:1:1 -p 0x7000000000000001:0x98f -P 0x7200000000000001:0x65 -L 9 -s 4096 -c 48 -o 0x12345:0x12304 /root/read_After_file_other_240
pver is : <0:0>
motr[67631]:  ba70  ERROR  [cob.c:904:cob_ios_rio_replied]  rpc item error = -110
motr[67631]:  e9f0  ERROR  [io_req_fop.c:287:io_bottom_half]  [0x2962bc0] rpc item 0x7fa13c053b50 rc=-110
motr[67631]:  e9f0  ERROR  [io_req_fop.c:349:io_bottom_half]  ioo=0x2962bc0 from=192.168.93.137@tcp:12345:2:2 rc=-110 ti_rc=-110 @<4300000300012345:12304>
motr[67631]:  e950   WARN  [io_req.c:1605:ioreq_dgmode_read]  Process failed parity groups in dgmode/read ioo=0x2962bc0 dgmap_nr=1

real    2m49.164s
user    0m1.272s
sys     0m0.467s
[root@ssc-vm-g2-rhev4-1052 ~]# md5sum /root/read_*
0fe57a405017f9f543ccea241e6ca2c2  /root/read_After_file_other_240
0fe57a405017f9f543ccea241e6ca2c2  /root/read_Before_file_other_240
[root@ssc-vm-g2-rhev4-1052 ~]# diff /root/read_After_file_other_240 /root/read_Before_file_other_240
[root@ssc-vm-g2-rhev4-1052 ~]# 
```

All above step worked fine.

mssawant:
hmm.. 47 and 48 secs for m0cp and m0cat is more than expected. @pavankrishnat, to confirm m0cp and m0cat timings are before node failure right?

pavankrishnat:
> hmm.. 47 and 48 secs for m0cp and m0cat is more than expected. @pavankrishnat, to confirm m0cp and m0cat timings are before node failure right?

yes @mssawant , those values are before node failure.

andriytk:
And what are the times without the patch?

andriytk:
Test result on the Sage cluster:

```
18:28 tkachuk1@sage-tier1-1:mcp$ time ./mcp -ep 172.18.1.1@o2ib:12345:4:1 -prof 0x7000000000000001:0x531 -proc 0x7200000000000001:0x19 -hax 172.18.1.1@o2ib:12345:1:1 -pool 0x6f00000000000001:0x4bf -osz 1024 -v /tmp/32k 0x123456023:0x87002603
2021/09/22 18:29:14 mio.go:542: W: off=0 len=32768 bs=1048576 gs=1048576 speed=710 (Kbytes/sec)

real	0m25.858s
user	0m1.389s
sys	0m1.752s
18:29 tkachuk1@sage-tier1-1:mcp$ rpm -q cortx-hare
cortx-hare-2.0.0-1_gitc991d62.el7.x86_64
```

25 secs is, of course, better than 40 secs, but definitely not good enough.

andriytk:
We see this regression also on the Sage cluster:

```
/c0cp 173772834631651062 17427322223284166419 './file1' 64 -p
bw = 000.0000 MB/s                                                                                             [ 00/01 ]
                                                                                                               
bw = 000.1966 MB/s                                                                                             [ 01/01 ]
                                                                                                               [  100% ]    init[ cput:     1.5700 s     0.0000 MB/s ][ wclk:    33.2723 s     0.0000 MB/s ]
  create[ cput:     0.0100 s     0.0000 MB/s ][ wclk:     0.0131 s     0.0000 MB/s ]
Motr I/O[ OSFS:     0.0001 s  1929.1749 MB/s ][ MOTR:     0.0362 s     5.4304 MB/s ]
    copy[ cput:     0.0200 s     9.8304 MB/s ][ wclk:     0.0420 s     4.6781 MB/s ]
    free[ cput:     0.1500 s     0.0000 MB/s ][ wclk:     0.3298 s     0.0000 MB/s ]
./file1 166152
c0cp success
#####
-rw-r--r-- 1 umanesan1 jusers 305472 Sep 23 13:48 ./file1
./c0cat 173772834631651062 17427322223284166419 './file2' 64 305472 -p
motr[02754]:  6860   WARN  [ha/link.c:1285:ha_link_outgoing_fop_replied]  rc=-67 nr=1 hl=0x1ac37b0 ep=172.18.1.24@o2ib:12345:1:1 lq_tags=(confirmed=2 delivered=2 next=2 assign=4)
motr[02754]:  6860   WARN  [ha/link.c:1289:ha_link_outgoing_fop_replied]  old_rc=0 old_nr=0 hl=0x1ac37b0 ep=172.18.1.24@o2ib:12345:1:1
motr[02754]:  6910  FATAL  [lib/assert.c:50:m0_panic]  panic: !rep->hae_disconnected_previously at ha_entrypoint_state_cb() (ha/ha.c:310)  [git: 2.0.0-307-11-gcacd425] 
Motr panic: !rep->hae_disconnected_previously at ha_entrypoint_state_cb() ha/ha.c:310 (errno: 0) (last failed: none) [git: 2.0.0-307-11-gcacd425] pid: 2754  
Motr panic reason: HA has already decided that this process has failed. There is no good reason to continue doing something, and there is no code yet to handle graceful shutdown in this case. Let's just terminate the process and let HA do it's job.
```

The test scenario is that c0cp utility finishes to write an object and c0cat immediately starts reading it back, and fails.

It's not quite clear whether it's caused by this patch, though. Maybe the issue was always there, and it just didn't show up because of long (40 secs) startup delays? And now with faster startups, we start seeing it again?

andriytk:
The time of `Preparing the reply for HaNvecGetEvent` is still very big even with the patch (about 13 secs with 8 m0_client processes configured per client node). See the snip from the system log - [sage-t11-delay-sys.log.gz](https://github.com/Seagate/cortx-hare/files/7246574/sage-t11-delay-sys.log.gz).

andriytk:
> The approach you're describing sounds as if we need to inline all operations into a single function (so that any result you fetch you can reuse via local variables) and get rid of some repeating.

No, this is not exactly what I'm proposing. We can store the fetched result from KV Store in a dictionary as a member of `ConsulUtil` class. We can add two more functions (methods) to this class: 1) to fetch the result from the KV Store (along with all the services' health) with one or two call to Consul and parse the result into the dictionary (one time!), let's call it `fetch_objs_status()`, and 2) to flush the result - `flush_objs_status()`.

Then we can call `ConsulUtil::get_conf_obj_status()` as usual in the loop for all objects from `Motr::ha_nvec_get_reply()`. And it will build the reply from the dictionary (instead of going to Consul making the recursive calls for the whole KV Store content and parsing its result on each iteration by every single object).

mssawant:
Tested degraded read with m0client running and node failure, dgread completed successfully. I think PR is good to land. Just few minor comments. cc @knekrasov, @vaibhavparatwar .

PR #1786, opened 2021-09-09
Problem: hare mini-provisioner doesn not generate motr confd.xc

mssawant:
Hare mini provisioners config stage generates all the required configuration
files for Motr and Hare but does not convert confd.dhall to confd.xc (Motr
configuration file in Motr xcode grammar.

Solution:
- Convert generated confd.dhall in text using `dhall text` command.
- Pass the output of `dhall text` to m0confgen to generate confd.xc.

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -523,6 +523,11 @@ def generate_config(url: str, path_to_cdf: str) -> None:
     if path:
         path += os.pathsep + '/opt/seagate/cortx/hare/bin/'
     python_path = os.pathsep.join(sys.path)
+
+    conf = ConfStoreProvider(url)
+    hostname = conf.get_hostname()
+    save(f'{conf_dir}/node-name', hostname)
+
     cmd = ['configure', '-c', conf_dir, path_to_cdf]
     execute(cmd, env={'PYTHONPATH': python_path, 'PATH': path,
                       'LC_ALL': "en_US.utf-8", 'LANG': "en_US.utf-8"})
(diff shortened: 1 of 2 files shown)


provisioning/miniprov/hare_mp/main.py
```diff
@@ -523,16 +523,31 @@ def generate_config(url: str, path_to_cdf: str) -> None:
     if path:
         path += os.pathsep + '/opt/seagate/cortx/hare/bin/'
     python_path = os.pathsep.join(sys.path)
+
+    conf = ConfStoreProvider(url)
```
knekrasov, line 527: @mssawant the node-name file will be written by `configure` script (it repeats the logic of hare-bootstrap, so node-name file is populated with the name evaluated from consul-agents.json file).

But the point is that if you write to node-name here, its content will be overwritten below by `configure` script, so this code has no effect.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -523,16 +523,31 @@ def generate_config(url: str, path_to_cdf: str) -> None:
...
     utils.import_kv(conf_dir)
     cmd = ['update-consul-conf']
     execute(cmd, env={'PYTHONPATH': python_path, 'PATH': path,
                       'LC_ALL': "en_US.utf-8", 'LANG': "en_US.utf-8"})
-    conf = ConfStoreProvider(url)
-    hostname = conf.get_hostname()
-    save(f'{conf_dir}/node-name', hostname)
+
+    # Create motr confd.xc
+    with open(f'{conf_dir}/confd.dhall', 'r') as confdhall:
```
knekrasov, line 540: @mssawant Do you really need this thing?

I believe this is already done by `configure` script.:
https://github.com/Seagate/cortx-hare/blob/220d3202f74bb32be6479956d3cbce4b3c28a52e/hax/helper/configure.py#L70-L81

provisioning/miniprov/hare_mp/utils.py
```diff
@@ -114,17 +114,18 @@ def import_kv(self, conf_dir_path: str):
                                str(item_data['value']))
 
 
-def execute(cmd: List[str], env=None) -> str:
+def execute(cmd: List[str], encoding=None, env=None, stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE, inp=None) -> str:
     process = subprocess.Popen(cmd,
-                               stdin=subprocess.PIPE,
-                               stdout=subprocess.PIPE,
+                               stdin=stdin,
```
knekrasov, line 120: Just for your information: perhaps the `Program` API from helper module may be less wordy and more convenient. 

Example how it is used in `configure` script:
https://github.com/Seagate/cortx-hare/blob/220d3202f74bb32be6479956d3cbce4b3c28a52e/hax/helper/configure.py#L70-L81

knekrasov: Moreover, it turns out that if you're trying to use subprocess to run piped commands (when one program's output becomes an input of the next one) is a [bit tricky](https://docs.python.org/3/library/subprocess.html#replacing-shell-pipeline): you need to forcibly close the output stream of the previous command. That is already made within Program/Executor API.

provisioning/miniprov/hare_mp/main.py
mssawant: Indeed! somehow when I was testing I didn't notice confd.xc getting created. probably  I made some mistake in the rpms that I installed.
We still have 1 problem, configure.py does not import consul kv due to which `update-consul-conf is not able to add services to `consul-server-conf.json`.
Let's abandon this PR. probably we will have to add consul kv import to configure.py but that will add dependency on consul to be running, which is true in the current situation but I am wondering if we should add this dependency on configure.py.

provisioning/miniprov/hare_mp/main.py
mssawant: Yes, agree. Abandoning this PR. will have to add consul import to configure.py or invoke update-consul-conf again.

mssawant:
Abandoning this PR as confd.xc is generated by configure script.

PR #1790, opened 2021-09-13
Problem: hare configure does not populate services information

mssawant:
Hare mini provisioner config stage generate Hare and Motr configuration, which
also comprises of information to register Hare and Motr services with Consul.
This information is not populated by hare configure utility.

Solution:
- Accept configuration directory path as an argument to Hare scripts.
- Fetch required services information from offline key values generated as
  part of Hare configuration.


Shreya-18 (approved):
Looks great!

knekrasov (approved):
Please resolve DCO problem

PR #1795, opened 2021-09-15
EOS-24972: Hare watcher scripts do not work in containers

mssawant:
Hare watchers scripts that monitor Hare, Motr and S3 servers processes use
systemd interface to fetch their respective statuses. Systemd interface is
typically not used in the container environments and thus the watcher scripts
fail and report motr processes as offline even though they are started.
This leads to hax reporting invalid failure statues causing further inconsistencies.

Solution:
- Configure Hare check-script to send curl requests on the Hare, Motr and S3 server
  ports to detect if the process is active or not.
- Use node name in hax http server url instead of hax endpoint.

### DIFF
--- a/utils/check-service
+++ b/utils/check-service
@@ -17,7 +17,7 @@
 # For any questions about this software or licensing,
 # please email opensource@seagate.com or cortx-questions@seagate.com.
 
-set -eu -o pipefail
+set -u -o pipefail
 
 prog=${0##*/}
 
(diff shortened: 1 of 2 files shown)


Shreya-18 (approved):
Looks good!

suvratjoshi (approved):

suvratjoshi:
@mssawant ,
Can you please add the JIRA id?
Thanks!

madhavemuri (approved):

PR #1797, opened 2021-09-17
cfgen: fix parity_units calc for dix and md pools

andriytk:
We must take into account the services configured at the enclosures, when we use them to calculate the `parity_units` parameter for the pools: CAS-services for the dix-pool and IOS services for md-pool. Otherwise, we can get the wrong values when the services are not present on the nodes. (Like, for example, in case of the pure client-nodes present in the cluster configuration.)

The same principle must be applied in case of calculating the `tolerance` vector for the pools, as well as the `mdredundancy` parameter.

### DIFF
--- a/cfgen/cfgen
+++ b/cfgen/cfgen
@@ -2145,7 +2145,7 @@ def build_cluster(cluster_desc: Dict[str, Any]) -> Cluster:
     if conf[root_id].mdpool is None:  # no Metadata pool defined in the CDF
         ConfPool.build(conf, root_id, {'type': 'md'})
 
-    conf[root_id].mdredundancy = len(cluster_svcs(conf, SvcT.M0_CST_CAS))
+    conf[root_id].mdredundancy = len(cluster_svcs(conf, SvcT.M0_CST_IOS))
 
     for prof in cluster_desc['profiles']:
         prof['pools'] += [aux_pool['name']
--- a/cfgen/cfgen
+++ b/cfgen/cfgen
@@ -1261,10 +1261,10 @@ def pool_drives(m0conf: Dict[Oid, Any],
 
 def cluster_encls(m0conf: Dict[Oid, Any], svc_t) -> List[Oid]:
     return list({encl_id
-            for encl_id in m0conf if encl_id.type is ObjT.enclosure
-            for proc_id in m0conf[m0conf[encl_id].node].processes
-            for svc_id in m0conf[proc_id].services
-            if m0conf[svc_id].type is svc_t})
+                 for encl_id in m0conf if encl_id.type is ObjT.enclosure
+                 for proc_id in m0conf[m0conf[encl_id].node].processes
+                 for svc_id in m0conf[proc_id].services
+                 if m0conf[svc_id].type is svc_t})
 
 
 def cluster_ctrls(m0conf: Dict[Oid, Any]) -> List[Oid]:
--- a/cfgen/cfgen
+++ b/cfgen/cfgen
@@ -1192,19 +1192,21 @@ def pool_type(pool_desc: Dict[str, Any]) -> PoolT:
     return PoolT[pool_desc.get('type', 'sns')]
 
 
+def cluster_svcs(m0conf: Dict[Oid, Any], svc_t) -> List[Oid]:
+    return [svc_id
+            for encl_id in m0conf if encl_id.type is ObjT.enclosure
+            for proc_id in m0conf[m0conf[encl_id].node].processes
+            for svc_id in m0conf[proc_id].services
+            if m0conf[svc_id].type is svc_t]
+
+
 def pool_drives(m0conf: Dict[Oid, Any],
                 pool_desc: Dict[str, Any]) -> List[Oid]:
     ptype = pool_type(pool_desc)
 
     if ptype is PoolT.dix:
-        cas = []
-        for item in m0conf:
-            if item.type is ObjT.enclosure:
-                encl_id = item
-                for proc_id in m0conf[m0conf[encl_id].node].processes:
-                    for svc_id in m0conf[proc_id].services:
-                        if m0conf[svc_id].type is SvcT.M0_CST_CAS:
-                            cas.append((svc_id, m0conf[svc_id].ctrl_id))
+        cas = [(svc_id, m0conf[svc_id].ctrl_id)
+               for svc_id in cluster_svcs(m0conf, SvcT.M0_CST_CAS)]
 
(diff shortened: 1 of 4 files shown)


cfgen/cfgen
```diff
@@ -1255,19 +1257,16 @@ def pool_drives(m0conf: Dict[Oid, Any],
 
 
 def cluster_encls(m0conf: Dict[Oid, Any]) -> List[Oid]:
```
mssawant, line 1259: May be this can be named better (may be `cluster_m0_servers`)?

mssawant (approved):

suvratjoshi (approved):

mssawant:
@andriytk, please update the commit message for mdrdundancy,
```
Calculate mdredundancy based on the number of CAS services
instead of number of nodes.
```

andriytk:
The number of IOS services was intended. Fixed the code.

cfgen/cfgen
```diff
@@ -720,7 +720,7 @@ class ConfRoot(ToDhall):
                   Downlink('profiles', ObjT.profile)}  # XXX + fdmi_flt_grps
 
     def __init__(self, nodes: List[Oid], sites: List[Oid], pools: List[Oid],
-                 profiles: List[Oid], mdpool: Oid = None,
+                 profiles: List[Oid], mdpool: Oid = None, mdredundancy: int = 0,
```
mssawant, line 723:
```
test/test_delivery_herald.py::TestDeliveryHeraldAny::test_exception_raised_by_timeout cfgen:723:80: E501 line too long (80 > 79 characters)
11:56:09  make[4]: *** [flake8] Error 1
11:56:09  make[4]: *** Waiting for unfinished jobs....
```

madhavemuri:
> We should count parity_units not by number of enclosures,
> but by number of cas services configured in case of dix
> pool, and by number of ios services configured in case of
> md pools.

I don't think so, it will add lot of overhead in case of large number of nodes and also with more than one ios/cas service per node.
Pool width can be equal to number of cas/ios services, but K needs to be equal to number of replicas (or node failures to be supported - 1)

cfgen/cfgen
```diff
@@ -1410,11 +1412,11 @@ class ConfPool(ToDhall):
                 # indivisible pieces of data, the whole CAS record is
                 # always stored on one node.
                 data_units = 1
-                parity_units = len(cluster_encls(m0conf)) - 1
+                parity_units = len(cluster_svcs(m0conf, SvcT.M0_CST_CAS)) - 1
```
mssawant, line 1415: I missed this somehow, good point by Madhav, idea to use number of enclosures is to have redundancy factor for metadata with respect to nodes. As 1 node can also run multiple ioservices, it this case with replication more that the nodes will lead to to much redundancy. We want to restrict replication factor with respect to the number of node failures. In worst case it is okay to have it equal to number of nodes instead of number of ioservices.

mssawant (commented):
Need to test this with node failure scenario.

cfgen/cfgen
andriytk: Good point. But we need to take into account server nodes, not just any nodes (which can be client-only ones).

cfgen/cfgen
andriytk: Fixed.

cfgen/cfgen
andriytk: Updated the patch, please check.

andriytk:
@madhavemuri good point! I've updated the description and the patch, please check. Thanks!

andriytk:
I was able to start the cluster with this patch:

```
$ hctl status
Data pool:
    # fid name
    0x6f00000000000001:0x2e 'the pool'
Profile:
    # fid name: pool(s)
    0x7000000000000001:0x48 'default': 'the pool' None None
Services:
    client 
    [started]  hax        0x7200000000000001:0x25  192.168.180.181@tcp:12345:1:1
    [unknown]  m0_client  0x7200000000000001:0x28  192.168.180.181@tcp:12345:4:1
    [unknown]  m0_client  0x7200000000000001:0x2b  192.168.180.181@tcp:12345:4:2
    centos79  (RC)
    [started]  hax        0x7200000000000001:0x6   192.168.180.171@tcp:12345:1:1
    [started]  confd      0x7200000000000001:0x9   192.168.180.171@tcp:12345:2:1
    [started]  ioservice  0x7200000000000001:0xc   192.168.180.171@tcp:12345:2:2
    [unknown]  m0_client  0x7200000000000001:0x1d  192.168.180.171@tcp:12345:4:1
    [unknown]  m0_client  0x7200000000000001:0x20  192.168.180.171@tcp:12345:4:2
```

cfgen/cfgen
```diff
@@ -1254,20 +1259,17 @@ def pool_drives(m0conf: Dict[Oid, Any],
     return [drive_id for drive_id, _ in drives]
 
 
-def cluster_encls(m0conf: Dict[Oid, Any]) -> List[Oid]:
-    encls: List[Oid] = []
-    for item in m0conf:
-        if item.type is ObjT.enclosure:
-            encls.append(item)
-    return encls
+def cluster_encls(m0conf: Dict[Oid, Any], svc_t) -> List[Oid]:
```
mssawant, line 1262: This will create multiple entries of the same enclosure in case of multiple ioservices (and CAS services) per enclosure.
Thus in case of a single enclosure, number of parity_units will still be 1 and tolerance vector for dix and md pools will be invalid  (`  , types.Obj.Pver { id = utils.zoid types.ObjT.Pver 79, data_units = 1, parity_units = 1, spare_units = 0, pool_width = 2, tolerance = [0, 0, 1, 1, 1], sitevs = [ utils.zoid types.ObjT.Objv 84 ] }`) as per this,
```
    def gen_md_tolerance(cls, m0conf: Dict[Oid, Any],
                         pool_desc: Dict[str, Any]) -> Failures:
        encls_nr = len(cluster_encls(m0conf, SvcT.M0_CST_IOS))
        ctrls_per_encl = floor(len(cluster_ctrls(m0conf)) / encls_nr)
        encl_failures = encls_nr - 1
        ctrl_failures = min(encl_failures, (ctrls_per_encl * encl_failures))
        return Failures(0, 0, encl_failures, ctrl_failures, encl_failures)
```

andriytk: Ouch, yes, good catch! Let me fix that...

andriytk: Fixed. Thanks!

andriytk:
@mssawant, when landing this patch, please squash all the commits and use the description and summary line for the commit msg from the PR above. Thank you.

PR #1798, opened 2021-09-17
Problem: hare cfgen considers pure m0clients in enclosures

mssawant:
Hare configuration generator calculates number of enclosures to identify
number of parity units for meta data pools (i.e. DIX and md). Number of
enclosures are also used to calculated mdredundancy value.
But a node may comprise of pure m0_clients as well (i.e. no m0_servers
but only motr client processes). Presently Hare cfgen considers m0_client
only nodes as well to calculate number of enclosures which results in
invalid value of redundancy factor.

Solution:
- Create Enclosure object only for m0_servers and not m0_clients.
- Set mdredundancy value explicitly

### DIFF
--- a/cfgen/cfgen
+++ b/cfgen/cfgen
@@ -2080,9 +2080,10 @@ def build_cluster(cluster_desc: Dict[str, Any]) -> Cluster:
 
     for node in cluster_desc['nodes']:
         node_id = ConfNode.build(conf, root_id, node)
-        encl_id = ConfEnclosure.build(conf, rack_id, node_id)
         ctrl_ids: List[Tuple[Optional[Oid], Any]] = []
         m0ds = node.get('m0_servers', [])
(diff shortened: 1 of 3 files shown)


cfgen/cfgen
```diff
@@ -2140,6 +2141,8 @@ def build_cluster(cluster_desc: Dict[str, Any]) -> Cluster:
     if conf[root_id].imeta_pver is None:  # no DIX pool defined in the CDF
         ConfPool.build(conf, root_id, {'type': 'dix'})
 
+    conf[root_id].mdredundancy = len(cluster_encls(conf)) - 1
```
andriytk, line 2144: What if there is only confd configured in `m0_servers` without any ios?

mssawant:
Closing the PR as the issue is addressed by #1797 .

PR #1801, opened 2021-09-22
Problem: no interface to start hax service without systemd

mssawant:
Presently without systemd, there is no other interface to start Hare hax service.
In-order to start hax process the corresponding environment and logging needs to
be configured explicitly in a non-systemd environment.

Solution:
- Add a sub command, start to hare_setup utility.
- Configure required environment, logging and start hax service in background
  as part of hare_setup start command.

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -32,6 +32,7 @@ from enum import Enum
 from sys import exit
 from time import sleep
 from typing import Any, Callable, Dict, List
+from threading import Event
 from urllib.parse import urlparse
 
 import yaml
@@ -45,6 +46,8 @@ from hare_mp.store import ConfStoreProvider
 from hare_mp.systemd import HaxUnitTransformer
 from hare_mp.validator import Validator
 from hare_mp.utils import execute, Utils
+from hare_mp.consul_starter import ConsulStarter
+from hare_mp.hax_starter import HaxStarter
 
 # Logger details
 LOG_DIR = "/var/log/seagate/hare/hare_deployment/"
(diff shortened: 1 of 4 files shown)


mssawant:
@knekrasov, @SwapnilGaonkar7, PR is about implementation of a start command for hare_setup utility, couple issues still need to be addressed,
1. Either need to document or also handle the LR case to use systemd interface as well.
2. Close the open loggin filehandlers, this also needs to be done for setup_logging() handlers.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -223,6 +223,36 @@ def init(args):
...
+            os.makedirs(var_hax_path)
+        logger = logging.getLogger('hax-log')
+        fhandler = logging.handlers.RotatingFileHandler(hare_log_file,
+                                                        maxBytes=LOG_FILE_SIZE,
+                                                        mode='a',
+                                                        backupCount=5,
+                                                        encoding=None,
+                                                        delay=False)
+        logger.addHandler(fhandler)
+        cmd = ['hax']
```
knekrasov, line 245: I think it is safer to execute hax script by the absolute path. In that case `PATH` env variable is not actually needed.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -223,6 +223,36 @@ def init(args):
         raise RuntimeError(f'Error while initializing cluster :key={error}')
 
 
+def start(args):
+    try:
+        path = os.getenv('PATH')
+        if path:
```
knekrasov, line 229: If path is None, what will happen then?

provisioning/miniprov/hare_mp/utils.py
```diff
@@ -128,3 +144,19 @@ def execute(cmd: List[str], env=None) -> str:
             f'Command output: {err}')
 
     return out
+
+
+def execute_no_communicate(cmd: List[str], env=None,
+                           working_dir: str = '/var/log/cortx/hare',
+                           out_file=subprocess.PIPE):
+    process = subprocess.Popen(cmd,
```
knekrasov, line 152: I don't quite get how is it supposed to work.

See, the end user will invoke `hare_setup start` to start hax. But in fact the first script spawns a new process that runs hax. Is the `hare_setup` process finishes keeping `hax` running in the background? 

If not, why [don't we block](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait) until `hax` process exits?

knekrasov (commented):
A few comments but the main question - why can't we simply use `hax` executable directly? Why do we need this play with CLI wrapper?

provisioning/miniprov/hare_mp/utils.py
mssawant: Yes, idea is to run hax in the background.

provisioning/miniprov/hare_mp/main.py
mssawant: Good point, fix by setting a default value to path if PATH is None. Removed the check.

provisioning/miniprov/hare_mp/main.py
mssawant: Fixed the path related issue.

mssawant:
>A few comments but the main question - why can't we simply use hax executable directly? Why do we need this play with CLI wrapper?
We need to start hax as part of Hare container startup but executing `hax` binary directly does not work as PYTHONPATH needs to be set.
```
[root@cortx-hare-motr-prov-pod ~]# /opt/seagate/cortx/hare/bin/hax
  File "/opt/seagate/cortx/hare/bin/hax", line 7, in <module>
    from hax.hax import main
ModuleNotFoundError: No module named 'hax'
```
Yes we can do that explicitly as part of container startup command but, instead why not have a single command to start Hare services (hax),
```
[root@cortx-hare-motr-prov-pod ~]# /opt/seagate/cortx/hare/bin/hare_setup start --config json:///tmp/hare.config.conf.1-node
[root@cortx-hare-motr-prov-pod ~]# ps -aux | grep hax
root      40992 96.3  1.3 3616316 224988 pts/0  Sl   20:35   0:08 /usr/bin/python3.6 /opt/seagate/cortx/hare/bin/hax
```
All the required environmental settings can be done behind the start command.

mssawant:
@knekrasov, @SwapnilGaonkar7, made hare_setup start command blocking. Also implemented consul starter to start consul along with hax. Idea is to use hare_setup start command as an entrypoint to the container and presently if either of the 2 processes (i.e. consul or hax) stop, hare container will stop.
Kindly review.

provisioning/miniprov/hare_mp/utils.py
mssawant: Blocking now after starting hax. To be used as a container entrypoint.

mssawant:
```
[root@cortx-io-pod-1 ~]# /opt/seagate/cortx/hare/bin/hare_setup start --config json:///tmp/hare.config.conf.n.cvgs.3-node


[root@cortx-io-pod-1 ~]# ps -aux | grep consul
root      43746  1.3  0.2 787188 40660 pts/5    Sl+  00:43   0:00 consul agent -bind=0.0.0.0 -client=0.0.0.0 -datacenter=dc1 -data-dir=/var/hare/consul/data -enable-script-checks -config-dir=/var/hare/consul/config -domain=consul -serf-lan-port=8301 -retry-join=consul-server:8301
root      43792  0.0  0.0  12532   980 pts/7    S+   00:43   0:00 grep --color=auto consul
[root@cortx-io-pod-1 ~]# ps -aux | grep hax
root      43745  1.1  0.1 165240 30620 pts/5    S+   00:43   0:00 /usr/bin/python3.6 /opt/seagate/cortx/hare/bin/hax
root      43794  0.0  0.0  12532   972 pts/7    S+   00:43   0:00 grep --color=auto hax
[root@cortx-io-pod-1 ~]# consul members
Node                          Address               Status  Type    Build   Protocol  DC   Segment
consul-server-0               192.168.148.235:8301  alive   server  1.10.0  2         dc1  <all>
consul-server-1               192.168.33.247:8301   alive   server  1.10.0  2         dc1  <all>
consul-server-2               192.168.220.152:8301  alive   server  1.10.0  2         dc1  <all>
cortx-io-pod-1                192.168.220.155:8301  alive   client  1.9.1   2         dc1  <default>
cortx-io-pod-2                192.168.148.238:8301  alive   client  1.9.1   2         dc1  <default>
cortx-io-pod-3                192.168.33.249:8301   alive   client  1.9.1   2         dc1  <default>
ssc-vm-3947.colo.seagate.com  192.168.220.151:8301  alive   client  1.10.0  2         dc1  <default>
ssc-vm-4161.colo.seagate.com  192.168.33.246:8301   alive   client  1.10.0  2         dc1  <default>
ssc-vm-4218.colo.seagate.com  192.168.148.234:8301  alive   client  1.10.0  2         dc1  <default>
```
```
2021-09-30 00:49:47,150 [ERROR] Stopping Hax
NoneType: None
2021-09-30 00:49:51,205 [ERROR] Stopping Consul
NoneType: None
[root@cortx-io-pod-1 ~]# ps -aux | grep hax
root      43918  0.0  0.0  12532   972 pts/5    S+   00:50   0:00 grep --color=auto hax
[root@cortx-io-pod-1 ~]# ps -aux | grep consul
root      43920  0.0  0.0  12532   980 pts/5    S+   00:50   0:00 grep --color=auto consul
[root@cortx-io-pod-1 ~]#
```

provisioning/miniprov/hare_mp/main.py
```diff
@@ -225,6 +228,104 @@ def init(args):
...
+    log_dir = f'{hare_local_dir}/consul/log'
+    if not os.path.exists(log_dir):
+        os.makedirs(log_dir)
+    data_dir = f'{hare_local_dir}/consul/data'
+    if not os.path.exists(data_dir):
+        os.makedirs(data_dir)
+    config_dir = f'{hare_local_dir}/consul/config'
+    if not os.path.exists(config_dir):
+        os.makedirs(config_dir)
+    peers = ['consul-server']
```
SwapnilGaonkar7, line 283: @mssawant  I guess this will be used in 'retry-join'. Are we not going to use endpoint provided by 'cortx>external>consul>endpoints' here.

mssawant: Yes, once we have merged all the new conf-store key changes, we can read the consul endpoint from conf-store. This can be done separately.

SwapnilGaonkar7 (approved):

suvratjoshi (approved):

PR #1813, opened 2021-10-02
EOS-24381 : Integration task with Motr, S3 for F-99A

suvratjoshi:
Accommodating recent key related changes for storage_set_id.
Key name storage_set_id chenged to storage_set.


SwapnilGaonkar7 (approved):
Looks good

PR #1814, opened 2021-10-03
Privisioner exiting incase of warning or error message

SwapnilGaonkar7:
Following changes in provisoner code causing provisioner to terminate in case
hare prints error or warning message.
Seagate/cortx-prvsnr#6209

So changing them to debug


mssawant:
Okay, @SwapnilGaonkar7, I think you missed adding the PR link to the provisioner change. We are directing the stderr nd stdout to log file but looks like the error and warning messages are still going to console. I think we need to fix it properly post demo.

mssawant (approved):
Looks good for now.

PR #1815, opened 2021-10-03
Problem: systemd check in hare_setup start may not work

mssawant:
hare_setup start command checks if systemd is enabled and accordigly
decides to start hax and consul without or with systemd. But,
hare_setup does not start consul in case systemd is enabled in the
containers.

Solution:
Instead of checking systemd internally in hare_setup start command,
take an explicit option to specifiy if hare_setp start is being executed
in environment with or without systemd.

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -316,51 +316,12 @@ def init_with_bootstrap(args):
         raise RuntimeError(f'Error while initializing cluster :key={error}')
 
 
-def __is_systemd_enabled() -> bool:
-    try:
-        cmd = ['systemctl']
-        execute(cmd)
-    except Exception:
-        return False
-    return True
-
-
 def start_hax_with_systemd():
     cmd = ['systemctl', 'start', 'hare-hax']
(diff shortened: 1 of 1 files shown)


mssawant:
```
[root@cortx-io-pod-1 ~]# /opt/seagate/cortx/hare/bin/hare_setup start --config json:///tmp/hare.config.conf.n.cvgs.3-node --nosystemd &
[1] 13168
[root@cortx-io-pod-1 ~]# ps -aux | grep hax
root      13187  9.5  0.1 165236 30604 pts/10   S    17:36   0:00 /usr/bin/python3.6 /opt/seagate/cortx/hare/bin/hax
```

madhavemuri:
@mssawant : Can you check for cluster type in confstore (common>environment_type) and if it is k8s, assume it to be non-systemd environment ?

mssawant:
@madhavemuri, Would prefer to have an explicit input, so let's have non systemd start as default and have an option as `--with-systemd`.

suvratjoshi (approved):

SwapnilGaonkar7 (approved):

PR #1818, opened 2021-10-04
EOS-24408: Integration of Hare Mini provisioner - support to F-87A 

SwapnilGaonkar7:
In provioner environment, currently UUID is being generated by
hare code before writing it into sysconfig files.
But according to motr requirement we should use 'machine_id' as UUID

As a solution :
Use machine_id as uuid, if specified as option
if not specified then generate it.

Also added code for copying confd.xc at $local/motr/sysconfig/<machine-id>/
For e.g. /etc/cortx/motr/sysconfig/<machine-id>/confd.xc

**Test**

Tested config method and sysconfig files are getting generated by using machine-id as UUID
Also tested bootstrap and in that case UUID is randomly generated

[root@ssc-vm-6327 hare]# /opt/seagate/cortx/hare/bin/hare_setup config --config yaml:///tmp/e24393/single_node_confstore.yaml
[root@ssc-vm-6327 hare]# cat /etc/machine-id
85fffa84fee1b12b46bcfc7cc491e602
[root@ssc-vm-6327 hare]# cat /etc/cortx/motr/sysconfig/85fffa84fee1b12b46bcfc7cc491e602/m0d-0x7200000000000001\:0x16
MOTR_M0D_EP='inet:tcp:ssc-vm-6327.colo.seagate.com@3002'
MOTR_HA_EP='inet:tcp:ssc-vm-6327.colo.seagate.com@2001'
MOTR_PROCESS_FID='0x7200000000000001:0x16'
MOTR_CONF_XC='/etc/motr/confd.xc'
MOTR_NODE_UUID='85fffa84fee1b12b46bcfc7cc491e602'

### DIFF
--- a/hax/helper/configure.py
+++ b/hax/helper/configure.py
@@ -36,6 +36,7 @@ class AppCtx:
     cdf_path: str
     conf_dir: str
     consul_server: bool
+    uuid: str
 
 
 def _setup_logging():
@@ -56,23 +57,36 @@ def _setup_logging():
               '-s',
               is_flag=True,
               help='Configure given node as a consul server.')
+@click.option('--uuid',
+              type=str,
+              help='UUID to be used',
+              show_default=True)
 @click.pass_context
-def parse_opts(ctx, cdf: str, conf_dir: str, consul_server: bool):
+def parse_opts(ctx, cdf: str,
+               conf_dir: str,
+               consul_server: bool,
+               uuid: str):
     """Generate Hare configuration according to the given CDF file.
 
     CDF   Full path to the Cluster Description File (CDF)."""
     ctx.ensure_object(dict)
-    ctx.obj['result'] = AppCtx(cdf_path=cdf, conf_dir=conf_dir,
-                               consul_server=consul_server)
+    ctx.obj['result'] = AppCtx(cdf_path=cdf,
+                               conf_dir=conf_dir,
+                               consul_server=consul_server,
+                               uuid=uuid)
     return ctx.obj
 
 
 class ConfGenerator:
-    def __init__(self, cdf_path: str, conf_dir: str,
-                 consul_server: bool):
+    def __init__(self,
+                 cdf_path: str,
+                 conf_dir: str,
+                 consul_server: bool,
+                 uuid: str):
         self.cdf_path = cdf_path
         self.conf_dir = conf_dir
         self.consul_server = consul_server
+        self.uuid = uuid
         self.executor = Executor()
 
     def generate(self) -> None:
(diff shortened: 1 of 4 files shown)


vaibhavparatwar (approved):

hax/helper/configure.py
```diff
@@ -135,9 +136,12 @@ def get_join_ip() -> str:
         ]),
                           env=self._get_pythonic_env())
 
+        machine_id = Conf.machine_id
+
         self.executor.run(Program([
             'update-consul-conf', '--conf-dir', f'{self.conf_dir}',
-            '--kv-file', f'{self.conf_dir}/consul-kv.json'
+            '--kv-file', f'{self.conf_dir}/consul-kv.json',
+            '--uuid', machine_id
```
mssawant, line 144: @SwapnilGaonkar7, in-order for --uuid to be optional, we need to add this optional to configure.py as well.

hax/helper/configure.py
```diff
@@ -135,9 +136,12 @@ def get_join_ip() -> str:
         ]),
                           env=self._get_pythonic_env())
 
+        machine_id = Conf.machine_id
```
knekrasov, line 139: Can we use our home-made ConfStoreProvider from `hare_mp` package instead?

Reasons:
1. It will make the code testable
2. It will make this code more coherent to other pieces that communicate to ConfStore
3. It works (in case it doesn't, we know how to fix it quick).

hax/helper/configure.py
```diff
@@ -135,11 +148,19 @@ def get_join_ip() -> str:
         ]),
                           env=self._get_pythonic_env())
 
-        self.executor.run(Program([
-            'update-consul-conf', '--conf-dir', f'{self.conf_dir}',
-            '--kv-file', f'{self.conf_dir}/consul-kv.json'
-        ]),
-                          env=self._get_pythonic_env())
+        if self.uuid:
```
mssawant, line 151: how about,
```
update_consul_conf_cmd = ['update-consul-conf', '--conf-dir', f'{self.conf_dir}','--kv-file', f'{self.conf_dir}/consul-kv.json']
if self.uuid:
    update_consul_conf_cmd.extend(['--uuid', self.uuid])
self.executor.run(Program(update_consul_conf_cmd), env=self._get_pythonic_env())
```

mssawant (approved):
Kindly address the comment(s). Review completed.

hax/helper/configure.py
SwapnilGaonkar7: Done

suvratjoshi (approved):

hax/helper/configure.py
```diff
@@ -51,20 +52,32 @@ def _setup_logging():
               help='Target folder where Hare-related configuration will '
               'be written to.',
               show_default=True)
+@click.option('--uuid',
+              type=str,
+              help='UUID to be used',
+              show_default=True)
```
knekrasov, line 58: [Minor] There is no default value anyway. Probably `show_default` makes no sense.

knekrasov (approved):

knekrasov:
@SwapnilGaonkar7 please resolve the conflicts so we can merge the PR

suvratjoshi (approved):

knekrasov (approved):

PR #1819, opened 2021-10-04
Problem: hctl interfaces fails in non-systemd environment

mssawant:
hctl interface fails with error,
```
[root@cortx-io-pod-1 ~]# hctl fetch-fids
Failed to create stream fd: No such file or directory
```
if hctl is executed in non-systemd environment.

Solution:
- check if systemd is enabled implicitly in hctl and accordingly
set log re-direction to journald or not.
```
[root@cortx-io-pod-1 ~]# hctl fetch-fids
[
{
"name": "ioservice",
"fid": "0x7200000000000001:0xa"
},
{
"name": "ioservice",
"fid": "0x7200000000000001:0x17"
},
{
"name": "confd",
"fid": "0x7200000000000001:0x24"
}
]
[root@cortx-io-pod-1 ~]#
```

### DIFF
--- a/hctl
+++ b/hctl
@@ -114,6 +114,14 @@ while true ; do
     shift
 done
 
+is_systemd_enabled() {
+   if [[ `ps --no-headers -o comm 1` == systemd ]]; then
+       true
+   else
+       false
+   fi
+}
+
 # process commands
 case $cmd in
     help) usage; exit ;;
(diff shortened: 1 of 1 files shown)


hctl
```diff
@@ -108,6 +111,7 @@ commands_usage() {
 while true ; do
     case $1 in
         -h|--help|help) usage; exit ;;
+        --systemd) systemd=true; ;;
```
knekrasov, line 114: From the UX perspective it looks like a hope for a smart user (i.e. the user understands if there is systemd available or not). Usually it is not the best assumption 😁 

Is it possible to avoid explicit CLI argument and make the script smarter instead?

mssawant: Hmm, let me see.

mssawant: Update the patch.

suvratjoshi (approved):

knekrasov (approved):

PR #1821, opened 2021-10-05
Problem: linters not run over helper module

knekrasov:
This means that even if there is a problem that mypy or flake8 could
have found and highlight during the build, these errors are not seen and
the build green even when there are obvious problems (like missing
positional argument).

Solution: add 'helper' module explicitly to `setup.py hax` CLI call. Fix
existing problems in helper module.


cortx-admin:

hessio:
Is this PR still relevant? I see the owner of this PR has left Seagate

shailesh-vaidya:
Closing as an obsolete

PR #1823, opened 2021-10-06
EOS-25251: locality thread is busy in hax code

knekrasov:
JIRA ticket: [EOS-25251](https://jts.seagate.com/browse/EOS-25251)

Entrypoint request is replied by Dummy thread (locality thread in libmotr) - that was done in a hope that this will speed up the replying process. In fact, since Dummy thread is external to Python interpreter, it holds GIL exclusively (making all other Pythonic threads to be frozen) until Dummy thread dives out of Python code. Another problem is that having this locality thread be busy with hax' business logic, can affect the logic of Motr.

Solution:
In any case, do reuse WorkPlanner abstraction for scheduling the Motr-related work in hax, so that entrypoint requests processing will happen in Pythonic threads only.

### DIFF
--- a/hax/hax/motr/__init__.py
+++ b/hax/hax/motr/__init__.py
@@ -19,7 +19,7 @@
 import ctypes as c
 import logging
 from errno import EAGAIN
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Type, Union
 
 from hax.exception import ConfdQuorumException, RepairRebalanceException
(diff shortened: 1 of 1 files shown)


mssawant:
@knekrasov, are we good to merge this PR?

cortx-admin:

vaibhavparatwar:
The bug for which this PR was raised is no more visible. @mssawant closin this PR. If these changes are needed, lets reopen and assess. cc @SunilSavanurSeagate

PR #1826, opened 2021-10-08
EOS-25289: nvec request processing takes too much time

mssawant:
Motr and s3server processes on startup requests cluster state from hare
via ha nvec requests. Hare saves an augmented state of the cluster objects
in Consul KV store. Number of cluster objects can easily go in hundreds,
Hare in-order to fetch the latest state of these objects sends multiple
requests to Consul, multilply the number of requests by the number of
motr and s3server processes running on the node. All this adds to the
delay of nvec processing. Nvec processing needs to be optimized to
prevent processes from timing out during initialization.
This interim fix that avoids nvec processing by returning always
online dummy state proves the problem in the nvec processing as the
cluster with multiple processes bootstraps successfully.

Solution:
As an interim solution, reply always online state of the requested
cluster configuration object.
Note that nvec requests are presently sent only during process startup
and runtime state of the cluster is not affected.
Also note that with this patch, dgread from a motr client that starts
post node failure will fail.

### DIFF
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -109,8 +109,8 @@ class ConsumerThread(StoppableThread):
                     if (self.consul.get_process_local_status(
                             state.fid) == 'M0_CONF_HA_PROCESS_STARTED'):
                         continue
-                if current_status in (ServiceHealth.FAILED,
-                                      ServiceHealth.STOPPED):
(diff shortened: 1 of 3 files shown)


mssawant:
@andriytk, possible, overall nvec path needs to be re-visited. We need to fix this because degraded IO post node failure will not work as mentioned in the comment.

PR #1839, opened 2021-10-20
Problem: incessant broadcast requests block entrypoint processing

mssawant:
Hax planner creates a separate group for entrypoint and broadcast requests.
Thus, one waits for the other to complete. Consul watcher probes are triggered
by 1 second interval thus there can be multiple broadcasts requests being
processed. This may block processing of entryoint reqeusts thus delaying the
overall motr process startup.
Also, broadcasting STOPPED on m0mkfs complete leads to `Motr panic:
(!pver->pv_is_dirty)` while setting up dix during motr ioservice startup.
This happens due to FAILED event of motr-mkfs (STOPPED) reaching motr m0d
(ioservice) startup due to delay. Failed event is mainly required to
disconnect halinks between Hare and Motr process. Motr-mkfs process anyway
sends STOPPED event on completion and Hare broadcasts Failed on that event.
Thus, subsequent FAILED event for motr-mkfs process is redundant which can
occur in context of consul notification.
Existing process statuses in Consul kv also lead to false notifications
processing during bootstrap.

Solution:
- Allow entrypoint and broadcast requests to be processed in parallel. If
one needs to wait for another, this needs to be done via an explicit wait,
e.g. FirstEntrypoint request. Implicit dependencies lead to performance
degradation.
- Cleanup existing process statuses in consul kv for a given node during
hax startup to avoid false failure notifications.
- Do not send FAILED event for motr-mkfs in context of consul notifications.

### DIFF
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -118,29 +118,17 @@ class ConsumerThread(StoppableThread):
                         # notifications, it may cause delay in cleanup
                         # activities.
                         continue
-                if current_status == ServiceHealth.UNKNOWN:
-                    # We got service status as UNKNOWN, that means hax was
-                    # notified about process failure but hax couldn't
-                    # confirm if the process is in failed state or have
-                    # failed and restarted. So, we will not loose the
-                    # event and try again to confirm the real time
-                    # process status by enqueing a broadcast event
-                    # specific to this process.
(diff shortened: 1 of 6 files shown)


vaibhavparatwar:
@mssawant  can we have a bug in JIRA for this issue and link it to this PR?

PR #1844, opened 2021-10-22
EOS-25582 entrypoint request is processed with delay

knekrasov:
When the cluster bootstraps, the health status picture changes very
frequently (especially given that the m0d processes in the cluster start
out of order). In this case a huge flow of BroadcastHAState events is
observed (reason: Consul notices the change in health statuses at
different nodes).

BroadcastHAState events by definition are processed sequentially, one by
one. If after a sequence of such events an entrypoint request comes, it
will start processing not earlier than the previous broadcasts are
fulfilled (and corresponding 'delivered' notifications are received).

This causes a significant delay of entrypoint request processing (up to
several minutes) and that becomes a reason of m0d processes hang.

## Solution
process entrypoint requests out of order. Entrypoint requests
are processed in the current group, regardless what other messages are
being processed in the moment.

### DIFF
--- a/hax/hax/motr/planner.py
+++ b/hax/hax/motr/planner.py
@@ -340,7 +340,8 @@ class WorkPlanner:
             self.state.next_group_id = self._get_increased_group(
                 self.state.next_group_id)
 
-        if isinstance(cmd, AnyEntrypointRequest):
+        if (isinstance(cmd, AnyEntrypointRequest)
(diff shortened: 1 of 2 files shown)


mssawant:
Testing was successful. @knekrasov can you please resolve the conflicts and add HaNvec requests as well to priority? cc @vaibhavparatwar

PR #1847, opened 2021-10-24
EOS-25502:Problem: hax still sends FAILED for motr mkfs process

mssawant:
When a motr-mkfs process reports STOPPED to hax, hax broadcasts FAILED
to all the other peer Motr processes and self in-order to disconnect the
ha links. In case of multiple ioservices per node, other ioservice, on
receiving this FAILED event for its fellow ioservice panics as during
its mkfs process as the corresponding configuration objects in the FAILED
mkfs process's heirarchy are marked FAILED and no clean pool version
can be found. In such a situation ioservice with mkfs in-progress fails
with a `!pv-dirty` assertion.
Also, sending a Failed event on entrypoint request also causes problem
if all the processes start in parallel.

Solution:
- Send OFFLINE event for mkfs STOPPED event
- Don't broadcast FAILED on receiving entrypoint request, this causes
inconsistencies when processes start in parallel.
- check consul kv status if node failure is reported, it could be mkfs
process completing.

### DIFF
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -30,7 +30,7 @@ from hax.motr import Motr
 from hax.motr.delivery import DeliveryHerald
 from hax.motr.planner import WorkPlanner
 from hax.queue.publish import EQPublisher
-from hax.types import (ConfHaProcess, HAState, MessageId, HaLinkMessagePromise,
+from hax.types import (ConfHaProcess, HAState, MessageId,
                        ObjT, ServiceHealth, StoppableThread, m0HaProcessEvent,
                        m0HaProcessType)
 from hax.util import ConsulUtil, dump_json, repeat_if_fails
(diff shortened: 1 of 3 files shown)


andriytk:
Can this be merged to the main branch also?

hax/hax/handler.py
```diff
@@ -167,17 +167,6 @@ def _do_work(self, planner: WorkPlanner, motr: Motr):
 
                     LOG.debug('Got %s message from planner', item)
                     if isinstance(item, FirstEntrypointRequest):
-                        LOG.debug('first entrypoint request, broadcast FAILED')
```
andriytk, line 170: Why did we do it in the 1st place, and what are we doing now instead?

mssawant: We were sending it because it was the safest place to report a process failure as hax received an entrypoint request. Consul notifications can be delayed and process can already be starting (entrypoint was already replied to). Hax cannot report a starting process as failed as that leads to other in-consistencies.
I still think reporting FAILED on receiving a first entrypoint request is the right place but due to out-of-order starting and with multiple ioservices on a node, Failed of one process may reach other process during its startup. This leads to the PV dirty issue where if no clean pool versions are found during dix setup motr process crashes.
Problem needs to be solved but we need some additional work there.

PR #1856, opened 2021-10-30
Problem: hax reports wrong confd quorum value

mssawant:
It is noted during testing that hax sometimes can report wrong quorum
value by calculating the active confds. This causes further
inconsistencies in motr processes startups.
This check in hax was added to avoid a hang situation during shutdown,
where hax's motr rconfc sends incessant entrypoint requests. Ideally,
reporting of confd failure must avoid this.

Solution:
Don't check for actve confds while sending entrypoint responses.

### DIFF
--- a/hax/hax/motr/__init__.py
+++ b/hax/hax/motr/__init__.py
@@ -224,12 +224,20 @@ class Motr:
             # case, so we wait until bootstrapping is done that is all the
             # motr services are up, we check the confd status and exclude
             # corresponding confd from the entrypoint reply.
-            active_confds = []
-            if self.spiel_ready:
-                for confd in confds:
-                    if not util.is_confd_failed(confd.fid):
-                        active_confds.append(confd)
-                confds = active_confds
+
+            # EOS-25726: It seems that the confds were reported as started
+            # and they failed later. This could be due to a Motr issue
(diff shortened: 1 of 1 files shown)


hax/hax/motr/__init__.py
```diff
@@ -224,12 +224,18 @@ def send_entrypoint_request_reply(self, message: EntrypointRequest):
             # case, so we wait until bootstrapping is done that is all the
             # motr services are up, we check the confd status and exclude
             # corresponding confd from the entrypoint reply.
-            active_confds = []
-            if self.spiel_ready:
-                for confd in confds:
-                    if not util.is_confd_failed(confd.fid):
-                        active_confds.append(confd)
-                confds = active_confds
+            # XXX It seems that the confds were reported as started and they
```
knekrasov, line 227: Is `XXX` reads "To be revisited later"?

If so, maybe it makes more sense to create the ticket so that this work will not be forgotten. I'm afraid that this way of notes in the code doesn't work in a large team.

mssawant: Sure, created https://jts.seagate.com/browse/EOS-25726

vaibhavparatwar (approved):

PR #1857, opened 2021-10-30
Add ARM64/AArch64 support

andriytk:
Add ARM64/AArch64 support:
1) `make rpm` works now on Arm64.
2) Fixed `libhax.cpython*.so` name for Arm64 at `HaxFFI::__init__`.

Testing: 1) rpm was built with `make rpm` and then installed successfully on the node; 2) singlenode cluster bootstrapped successfully:

```
$ hctl bootstrap --mkfs singlenode.yaml 
2021-10-30 13:14:43: Generating cluster configuration... OK
2021-10-30 13:15:25: Starting Consul server on this node..................................................................................... OK
2021-10-30 13:18:21: Importing configuration into the KV store... OK
2021-10-30 13:18:24: Starting Consul on other nodes...Consul ready on all nodes
2021-10-30 13:18:26: Updating Consul configuraton from the KV store... OK
2021-10-30 13:18:43: Waiting for the RC Leader to get elected...... OK
2021-10-30 13:19:00: Starting Motr (phase1, mkfs)... OK
2021-10-30 13:19:57: Starting Motr (phase1, m0d)... OK
2021-10-30 13:20:36: Starting Motr (phase2, mkfs)... OK
2021-10-30 13:21:28: Starting Motr (phase2, m0d)... OK
2021-10-30 13:22:24: Checking health of services... OK
```

Note: `sage-2.1` branch was used for testing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,9 +59,14 @@ SETUP_PY    := $(PY_VENV); $(PYTHON) setup.py
 PY3_VERSION := 36
 PY3_VERSION_MINOR := $(shell grep -o . <<<$(PY3_VERSION) | tail -n1)
 DHALL_VERSION := 1.26.1
-DHALL_URL     := https://github.com/dhall-lang/dhall-haskell/releases/download/$(DHALL_VERSION)/dhall-$(DHALL_VERSION)-x86_64-linux.tar.bz2
 DHALL_JSON_VERSION := 1.4.1
-DHALL_JSON_URL     := https://github.com/dhall-lang/dhall-haskell/releases/download/$(DHALL_VERSION)/dhall-json-$(DHALL_JSON_VERSION)-x86_64-linux.tar.bz2
+PLATFORM := $(shell uname -p)
+DHALL_URL      := https://github.com/dhall-lang/dhall-haskell/releases/download/$(DHALL_VERSION)/dhall-$(DHALL_VERSION)-$(PLATFORM)-linux.tar.bz2
+DHALL_JSON_URL := https://github.com/dhall-lang/dhall-haskell/releases/download/$(DHALL_VERSION)/dhall-json-$(DHALL_JSON_VERSION)-$(PLATFORM)-linux.tar.bz2
+ifeq ($(PLATFORM),aarch64)
+  DHALL_URL      := https://github.com/andriytk/dhall-haskell/releases/download/$(DHALL_VERSION)-el7-arm64/dhall-$(DHALL_VERSION)-$(PLATFORM).linux.tar.bz2
+  DHALL_JSON_URL := https://github.com/andriytk/dhall-haskell/releases/download/$(DHALL_VERSION)-el7-arm64/dhall-json-$(DHALL_JSON_VERSION)-$(PLATFORM).linux.tar.bz2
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -59,9 +59,14 @@ SETUP_PY    := $(PY_VENV); $(PYTHON) setup.py
...
-DHALL_JSON_URL     := https://github.com/dhall-lang/dhall-haskell/releases/download/$(DHALL_VERSION)/dhall-json-$(DHALL_JSON_VERSION)-x86_64-linux.tar.bz2
+PLATFORM := $(shell uname -p)
+DHALL_URL      := https://github.com/dhall-lang/dhall-haskell/releases/download/$(DHALL_VERSION)/dhall-$(DHALL_VERSION)-$(PLATFORM)-linux.tar.bz2
+DHALL_JSON_URL := https://github.com/dhall-lang/dhall-haskell/releases/download/$(DHALL_VERSION)/dhall-json-$(DHALL_JSON_VERSION)-$(PLATFORM)-linux.tar.bz2
+ifeq ($(PLATFORM),aarch64)
+  DHALL_URL      := https://github.com/andriytk/dhall-haskell/releases/download/$(DHALL_VERSION)-el7-arm64/dhall-$(DHALL_VERSION)-$(PLATFORM).linux.tar.bz2
+  DHALL_JSON_URL := https://github.com/andriytk/dhall-haskell/releases/download/$(DHALL_VERSION)-el7-arm64/dhall-json-$(DHALL_JSON_VERSION)-$(PLATFORM).linux.tar.bz2
+endif
 DHALL_PRELUDE_VERSION := 11.1.0
 DHALL_PRELUDE_URL     := https://github.com/dhall-lang/dhall-lang/archive/v$(DHALL_PRELUDE_VERSION).tar.gz
```
andriytk, line 71: I failed to see how the dhall-prelude is used. It seems like it isn't used at all. Should we drop it in another PR?

knekrasov: As far as I understand, Prelude is the basic set of types and functions available in Dhall language. So if you remove it, dhall binary will probably work but none of Dhall expressions will compile.

andriytk: Then, any idea how it works on Arm64 without it? Can you point out the code which uses it?

knekrasov: I think it is used here: 
https://github.com/Seagate/cortx-hare/blob/main/cfgen/dhall/Prelude.dhall#L23-L26

https://github.com/Seagate/cortx-hare/blob/3c517ed5b00930cbd8080cba69224d33ffcf8ea3/cfgen/dhall/Prelude.dhall#L23-L26

And if I understand the code right, the Prelude is taken from the filesystem but as a fallback it can also download the package from the internet.

andriytk: Well, I don't see where the `DHALL_PRELUDE` is used in the code at all. Looks like it doesn't.

knekrasov: It does. Otherwise `./dhall-lang-11.1.0/Prelude/package.dhall` file would simply not appear at the file system.

knekrasov (approved):

mssawant (approved):

PR #1858, opened 2021-10-31
hare-bootstrap: fix consul leader check

andriytk:
With bash `set -o pipefail`, the following construction may
give a false negative status result:

    cmd | grep -q something

if the `cmd` returns multiple lines.

The reason is that the grep command with `-q` option finishes
immediately as soon as it finds `something`.  But in case of
multiple lines output from `cmd`, the latter might get killed
by SIGPIPE (because grep has already exited). So the resulting
status of the whole command will be 141 instead of 0.

This is exactly what sometimes happens when we check the status
of internal Consul leader on bootstrap.  On slow virtual machines
(like arm64 in QEMU) the check in the loop might take a lot of
time returning false negative results, until it happens that
the grep process is still running, and the pipe is not closed,
by the time `cmd` finishes outputting into the pipe.

Solution: use the following construction instead:

    grep -q something <<< $(cmd)


mssawant:
@andriytk, does this mean that before the `cmd` puts data in the pipe, grep tries to read from it and fails?

andriytk:
> @andriytk, does this mean that before the `cmd` puts data in the pipe, grep tries to read from it and fails?

@mssawant no, not exactly. Grep does read the data and finds the match. And as soon as finds it, it finishes (because of `-q` option). But `cmd` is not finished yet. It still has some data to put into the pipe. But the pipe is not available any more (because the grep process has finished already). So the `cmd` is killed by SIGPIPE signal. As a result, the overall command status fails (because bash was `set -o pipefail`, so the resulting status is the status of the 1st failed cmd in the pipe).

knekrasov (approved):
The code is good.

[Optional] Suggestion: add an explanatory comment to the code to highlight why is it implemented like this (so when somebody tries to write a similar code, this issue will be understood).

Moverover, bash expressions are very hard to google, so this kind of change will be hard to analyze.

andriytk:
Thanks @knekrasov. Added. Please, take a look if it's clear enough.

mssawant (approved):

PR #1859, opened 2021-11-02
EOS-25703: hctl status takes too much time to complete in containers

mssawant:
hctl status takes too much time in containers and also report inacurate statuses
for Motr and s3servers. hctl status looks at the health of consul service corresponding
to the Motr and s3server processes. Consul service status uses consul watcher which
probes the process endpoint. Thus a motr process can be running but not yet functional.

Solution:
Read Motr and s3server statuses saved in Consul KV that is updated when Motr process
reports the same. Thus is accurate from functionally readiness perspective of Motr
or s3server processes.

### DIFF
--- a/hax/hax/util.py
+++ b/hax/hax/util.py
@@ -1328,8 +1328,10 @@ class ConsulUtil:
     @uses_consul_cache
     def get_process_status(self,
                            fid: Fid,
+                           proc_node=None,
                            kv_cache=None) -> MotrConsulProcInfo:
-        proc_node = self.get_process_node(fid, kv_cache=kv_cache)
+        if not proc_node:
+            proc_node = self.get_process_node(fid, kv_cache=kv_cache)
         key = f'{proc_node}/processes/{fid}'
         status = self.kv.kv_get(key, kv_cache=kv_cache)
         if status:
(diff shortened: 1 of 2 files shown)


SwapnilGaonkar7 (approved):

knekrasov (approved):
Although the changes (from the functionality point of view) are correct and good, do they really speed `hctl status` up?

If so, what was the reason of slowness? Extra queries to KV when we could learn that the node is offline (so the status of the services is automatically offline)?

suvratjoshi (approved):

mssawant:
>Although the changes (from the functionality point of view) are correct and good, do they really speed hctl status up?

If so, what was the reason of slowness? Extra queries to KV when we could learn that the node is offline (so the status of the services is automatically offline)?

- I can see more than 50% improvement, earlier, we were trying to fetch consul service's status which takes the results provided by the watcher check-service script. check service script probe may take some time. Although it is okay, this was not only taking time but the status reported was also in-accurate.
- Now we fetch the saved status from Consul KV instead. 
- We can improve further may be but that will need some additional changes where we fetch the statuses for a node recursively from consul kv and then parse it.

PR #1861, opened 2021-11-02
spec: fix make rpm on CentOS 8

andriytk:
`make rpm` is failing on CentOS 8, solution:
1) `python36-*` packages are called `python3-*` in CentOS 8 - update in requirements.
2) Fix `h_motr_version` for motr requirement: it is taking the VERSION and RELEASE values from Hare instead of Motr. Fix by adding one more escape `%` symbol before VERSION and RELEASE.

### DIFF
--- a/hare.spec
+++ b/hare.spec
@@ -41,10 +41,17 @@ BuildRequires: cortx-motr
 BuildRequires: cortx-motr-devel
 BuildRequires: cortx-py-utils
 BuildRequires: git
+%if %{rhel} < 8
 BuildRequires: python36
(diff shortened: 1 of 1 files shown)


mssawant:
@andriytk, the problem is not very clear, can you please elaborate the same in the commit message and the PR description?

andriytk:
@mssawant, is it better now?

suvratjoshi (approved):

knekrasov (approved):

andriytk:
retest this please

mssawant (approved):

PR #1862, opened 2021-11-02
Fix facter version check on CentOS 8

andriytk:
On CentOS 8 the standard facter rpm version is 3.14,
which fits our needs. So the puppet-agent rpm (which
provided the needed version of facter-3.14 to us on
CentOS 7) is no longer required on CentOS 8.

-----
[View rendered README.md](https://github.com/andriytk/cortx-hare/blob/fix-facter-check-centos8/README.md)

### DIFF
--- a/hare.spec
+++ b/hare.spec
@@ -47,7 +47,11 @@ BuildRequires: python36-pip
 BuildRequires: python36-setuptools
 
 Requires: consul >= 1.7.0, consul < 1.10.0
+%if %{rhel} < 8
 Requires: puppet-agent >= 6.13.0
+%else
+Requires: facter >= 3.14.2
+%endif
 Requires: jq
 Requires: cortx-motr = %{h_motr_version}
 Requires: cortx-py-utils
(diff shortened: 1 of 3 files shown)


mssawant (approved):

hare.spec
```diff
@@ -91,11 +95,13 @@ groupadd --force hare
 chgrp hare /var/lib/hare
 chmod --changes g+w /var/lib/hare
 
+%if %{rhel} < 8
 # puppet-agent provides a newer version of facter, but sometimes it might not be
```
suvratjoshi, line 99: Can this mentioned thing in the comment happen with CentOs ver =>8 (sometimes it might not be available at path ...)?
So I was wondering, can it be more risk free if we do not include this added `%if %{rhel} < 8` condition here.
Thanks!

knekrasov (approved):

hare.spec
andriytk: No, it cannot happen in CentOS 8.

suvratjoshi (approved):

PR #1867, opened 2021-11-03
fixup! spec: fix make rpm on CentOS 8

andriytk:
There is no need to check for CentOS 7 in defining `h_motr_version`: escaping `%` with one more `%` works fine in CentOS 7 as well.


mssawant:
Branch has conflicts, kindly resolve the same.

PR #1871, opened 2021-11-09
EOS-25863: BroadcastHAStates is slow at large clusters

knekrasov:
Solution:
1. Make sure that execution path of broadcast_ha_states reuses the same
kv_cache instance (to minimize repeated KV GET requests).
2. Refactor the KV-modifying functions so that actual kv_put()
operations can be postponed when possible (reason: kv_put() invalidates
the current kv_cache instance which may be an overkill if kv_put is
invoked too early).

### DIFF
--- a/hax/hax/motr/__init__.py
+++ b/hax/hax/motr/__init__.py
@@ -19,7 +19,7 @@
 import ctypes as c
 import logging
 from errno import EAGAIN
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Tuple
 
 from hax.consul.cache import supports_consul_cache, uses_consul_cache
 from hax.exception import ConfdQuorumException, RepairRebalanceException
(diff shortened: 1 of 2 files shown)


mssawant:
@knekrasov, so why don't we put process events as well at the head of the command queue as nvec and entrypoint requests? if not executing them in parallel?

mssawant:
Yeah, we cannot put process events are the head of the queue, but the current path is not helping much, we are still seeing a delay of 30+ mins in processing process events.

knekrasov:
@mssawant I think that it is too risky to allow ProcessEvent to be processed out of order of BroadcastHAStates. The idea of parallelized groups (partitions) was as follows:
1. There exist a group of events that don't depend on each other. in other words, there is no harm if events A and B are processed in any order A->B, B->A or (A, B) (at the same time).
2. But there are also events that should be processed one after another. This is the place when partial ordering helps.

Because both BroadcastHAStates and ProcessStates are about process states and since we all know that these events can be affect the health status of the processes (e.g. broadcasting FAILED event about an alive process is considered as a command to commit a suicide), I considered those two events as mutually exclusive and I guessed that the order of those events is critical.

If you think that it doesn't make any sense, then let's consider all events as always allowed for parallel execution (i.e. all commands will appear to be in one single group). I have a gut feeling that this will cause Heisenbugs because of race condition, but in a happy path everything will become super fast.

@nikitadanilov I think I need your opinion here. This discussion is based on mechanism described here: https://seagate-systems.atlassian.net/wiki/spaces/PRIVATECOR/pages/178421786/22+PARHAX+Parallel+message+processing+in+HaX

vaibhavparatwar (approved):

mssawant:
@knekrasov, agree, as mentioned above we cannot process ProcessEvents out of order. We need to process them independent of Broadcast but in order.

mssawant:
Tested the present change,  its good enough from caching broadcast events processing perspective, so going ahead with merging this PR.

PR #1873, opened 2021-11-11
EOS-25537: Improvement in aux pool enable/disable implementation

supriyachavan4398:
Presently there's an option to disable aux pool creation in the Hare configuration generator and there is disk repetition found in base pools. Logically, disk repetition is not allowed for base pools but only aux pools can have disks repeated. So, we don't allow overlapping pools, i.e. pools sharing physical devices between them.

Solution:
Depending on the provided configuration we should be able to generate correct combinations of pools. For this, We need to check specifically for the pool instead of pool versions. and if disks are repeated in base pools then it should be thrown AssertionError for that.

Signed-off-by: Supriya Yadav <supriya.s.chavan@seagate.com>

### DIFF
--- a/cfgen/cfgen
+++ b/cfgen/cfgen
@@ -1257,9 +1257,44 @@ def cluster_svcs(m0conf: Dict[Oid, Any], svc_t) -> List[Oid]:
             if m0conf[svc_id].type is svc_t]
 
 
+def pool_drives_with_nodes(m0conf: Dict[Oid, Any]) -> List[Tuple[Oid, Oid]]:
+    all_drives: List[Tuple[Oid, Oid]] = [
+        (drive_id, m0conf[encl_id].node)
+        for encl_id in m0conf if encl_id.type is ObjT.enclosure
+        for ctrl_id in m0conf[encl_id].ctrls
+        for drive_id in m0conf[ctrl_id].drives
+    ]
+    assert all_drives
+    assert all(drive_id.type is ObjT.drive and node_id.type is ObjT.node
+               for drive_id, node_id in all_drives)
+
+    return all_drives
+
+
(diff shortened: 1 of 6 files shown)


supriyachavan4398:
Added some required conditionals to the patch.
Tested with one node and three node clusters.
Test conditions used while testing:
          1. No presence of variable create_aux in the CDF (which defaults to False)
          2. create_aux: False
          3. create_aux: True
 
For create_aux=True, bootstrap fails for some node/layout combinations, which is expected.
For create_aux=False, bootstrap works well (as tested for 1 node and 3 node cluster).

mssawant:
@supriyachavan4398 the problem statement in the commit message and PR description is not very clear, kindly update the same to be more relevant.

mssawant:
>We are provided enable/disable aux pools functionality. In this, the function is
checking if a drive is referenced in multiple pools, every pool has a base pool
version and multiple virtual pools. So we are checking if a drive belongs to
multiple pool versions and thus to multiple pools. We don't want to allow a drive
to be part of multiple pools(in this case aux pools). Basically, disk repetition
is not allowed for base pools but only aux pools can have disks repeated.

Presently there's no option to disable aux pool creation in Hare configuration generator. In a fairly large cluster, the number of auxiliary pool generated can be huge. Thus a mechanism to enable/disable auxiliary pool generation is required. Also, we don't to allow overlapping pools, i.e. pools sharing physical devices between them.

mssawant (approved):
Looks good overall, kindly update the commit message and PR description.

supriyachavan4398:
All the related changes are addressed by https://github.com/Seagate/cortx-hare/pull/1888. So closing this.
cc. @mssawant @vaibhavparatwar

PR #1880, opened 2021-11-15
EOS-25916: mkfs halink is not disconnected

mssawant:
Hax relays self's and motr-mkfs process events to all the peer connected motr-mkfs
and to hax's motr land. Eventhough motr-mkfs processes event broadcasts are
restricted to the process states only and not to the entire motr process's
configuratio tree hierarchy (i.e. process's corresponding node, controller,
enclosure and drives), hax process events broadcasts happen to its corresponding
configuration tree hierarchy. Thus, when hax's motr land starts, the event is a
broadcast to all the connected motr-mkfs processes as well. As process, entrypoint
and nvec motr process events are given higher proirities, it may happen that
hax's process event broadcast may take time. It is observed that in a certain
situation, hax tries to send an event notification to itself and other connected
motr-mkfs processes, it may happen that although hax only starts to use the halink
after hax gets a `link_connected_cb()` call back from motr land (which should assure
that the halink is ready to use), the halink is not ready and hax crashes in Motr
land due to `Motr panic: (m0_ha_lq_invariant(lq)) at m0_ha_lq_enqueue() ha/lq.c:199`
invariant failure.
A related bug identified was that the halink saved in hax was not deactivated in
`link_disconnecting_cb()` received from hax's motr land, although, the invariant
failure was observed even after deactivating the halink.
All the evidence and the problem's intermittent occurance points to a race condition,
the suspect of the same is the halink being either not ready to use or being used
while it is getting disconnected. This needs to be further investigated.

Few other related fixes in hax tend to avoid this halink crash in the motr land.

Other fixes:
- Presently hax sends OFFLINE for mkfs M0_CONF_HA_PROCESS_STOPPED event to
avoid pv-dirty crashes during motr-mkfs to peer motr processes. But, this restricts
the motr-mkfs halink to be disconnected. Thus send FAILED notification for
motr-mkfs STOPPED event.
- Motr mkfs processes execute independent of each other and are short lived, thus,
they do not require a STOPPED/FAILED notification corresponding to a peer motr-mkfs
process, but, hax motr land requires the same in order to disconnect hax - motr-mkfs
halink corresponding to a motr-mkfs process, thus the broadcast of motr-mkfs events
need to be only to hax.
- Hax process events are a broadcast to all the connected motr processes which also
include hax's configuration tree hierarchy (i.e. hax's node, enclosure, controller,
etc.). Hax failure needs to be handled separately and such a broadcast is not
required as hax may share a node with other motr processes. This may lead to
inconsistencies, thus avoid broadcasting hax states to motr processes.

### DIFF
--- a/hax/hax/handler.py
+++ b/hax/hax/handler.py
@@ -76,7 +76,7 @@ class ConsumerThread(StoppableThread):
                     ServiceHealth.OK),
             (m0HaProcessType.M0_CONF_HA_PROCESS_M0MKFS,
                 m0HaProcessEvent.M0_CONF_HA_PROCESS_STOPPED): (
-                    ServiceHealth.OFFLINE),
+                    ServiceHealth.FAILED),
             (m0HaProcessType.M0_CONF_HA_PROCESS_M0D,
                 m0HaProcessEvent.M0_CONF_HA_PROCESS_STARTED): (
                     ServiceHealth.OK),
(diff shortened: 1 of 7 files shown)


knekrasov (approved):

vaibhavparatwar (approved):

mssawant:
Merging as the several deployments completed successfully with the custom build http://eos-jenkins.colo.seagate.com/job/GitHub-custom-ci-builds/job/centos-7.9/job/cortx-all-image-custom-ci/1117/.

https://jts.seagate.com/browse/EOS-25926?focusedCommentId=2241445&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2241445

PR #1882, opened 2021-11-17
EOS-25377: StobIoqError type not JSON-serializable

knekrasov:
This is a duplicate of #1833 for `kubernetes` branch.

Solution:
1. Implement custom serializer function for StobIoqError (for_json())
2. Implement integration tests that verify that the bug is not
reproduced anymore:
   - StobIoqError is serializable by simplejson library
   - Resulting payload is understood by the BQ processor

### DIFF
--- a/hax/hax/message.py
+++ b/hax/hax/message.py
@@ -17,7 +17,7 @@
 #
 
 import queue as q
-from dataclasses import dataclass, field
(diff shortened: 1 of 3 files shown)


vaibhavparatwar (approved):

knekrasov:
Closing because #1885 already has these changes.

PR #1886, opened 2021-11-21
EOS-25703: hctl status takes too much time in containers

Shreya-18:
### Description:
In the containers environment, consul kv calls take a longer time for execution in comparison to the LR environment.
An experiment was done which should a significant difference of 17 mins between the two environments. 

### Solution:
Reduced number of consul kv calls.
Initially, for every process, multiple separate kv calls were made. Now, we are fetching kv for all the processes under a node at once and parsing it further.

### Testing:
- Time taken with the final change ( fetching endpoint for every process function) is yet to be tested on 15 node.
- on 5node could see 33% reduction.
#### time taken in 15 node setup 
- with code changes in 2 functions -  _process_status()_ and _proc_fid2name()_ fetching at node level is:
```
real 1m5.397s
user 0m9.109s
sys 0m0.164s
```
- original hct status script took
```
real 4m5.153s
user 0m22.156s
sys 0m0.235s
```
Signed-off-by: Shreya Karmakar <shreya.karmakar@seagate.com>

### DIFF
--- a/utils/hare-status
+++ b/utils/hare-status
@@ -24,6 +24,7 @@ import argparse
 import inject
 import io
 import logging
+import re
 import simplejson
 import sys
 from subprocess import PIPE, Popen
@@ -124,26 +125,6 @@ def get_fs_stats(cns: Consul) -> Any:
     return {'stats': {}} if stats is None else json.loads(stats)
 
 
-def proc_id2name(cns: Consul, node: str, proc_id: int) -> str:
-    names = {
-        'confd': 'confd',
-        'ha': 'hax',
-        'ios': 'ioservice',
-        'm0_client_s3': 's3server'
-    }
-    services = kv_item(cns,
-                       f'm0conf/nodes/{node}/processes/{proc_id}/services/',
-                       recurse=True)
-    assert services
-    for svc in services:
-        # key looks like
-        # 'm0conf/nodes/<name>/processes/<process_fidk>/services/<svc_type>'
-        svc_type = svc['Key'].split('/')[-1]
-        if svc_type in names:
-            return names[svc_type]
-    return 'm0_client'
-
-
 def fid_key(fid: str) -> int:
     key = fid.split(':')[1]
     return int(key, 16)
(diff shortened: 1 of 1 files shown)


utils/hare-status
```diff
@@ -167,31 +148,91 @@ def processes(cns: Consul, consul_util: ConsulUtil,
...
+            state = 'started'
+        elif val['state'] in ('Unknown', 'UNKNOWN'):
+            state = 'unknown'
+        else:
+            state = 'offline'
+        process_states[Fid.parse(fid)] = state
+    return process_states
+
+
+def proc_id2endpoint(process_list: Any, proc_id: int) -> Optional[str]:
```
knekrasov, line 195: [optional] I think we should avoid using `Any` when possible.

```suggestion
def proc_id2endpoint(process_list: List[Dict[str, Any]], proc_id: int) -> Optional[str]:
```

knekrasov (commented):
The code looks good but probably we should improve the type hints.

utils/hare-status
Shreya-18: done

knekrasov (approved):

vaibhavparatwar (approved):

utils/hare-status
```diff
@@ -167,31 +148,91 @@ def processes(cns: Consul, consul_util: ConsulUtil,
...
+            state = 'started'
+        elif val['state'] in ('Unknown', 'UNKNOWN'):
+            state = 'unknown'
+        else:
+            state = 'offline'
+        process_states[fid] = state
+    return process_states
+
+
+def proc_id2endpoint(process_list: List[Dict[str, Any]],
```
mssawant, line 195: Here we are iterating on each consul kv, another option is to use, `consul_util.catalog.get_services()`, check `get_service_data_by_name(service_name)` this will give all the details for a given service type on a node including its status and endpoint. So we don't have to fetch endpoint individually for every process.
For process status we still need to fetch it from consul kv but for endpoint we can use this.

utils/hare-status
```diff
@@ -167,31 +148,91 @@ def processes(cns: Consul, consul_util: ConsulUtil,
...
-        return 'unknown'
+    if str(fid) in proc_states:
+        return proc_states[str(fid)]
     else:
-        return 'offline'
+        return 'unknown'
+
+
+def get_node_process_states(cns: Consul, node_name: str) -> Dict[str, str]:
+    key = f'{node_name}/processes/'
```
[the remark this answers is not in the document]
Shreya-18: this **key = f'{node_name}/processes/'** fetches below format key
```
cortx-data-headless-svc-ssc-vm-g3-rhev4-2387/processes/0x7200000000000001:0x27:{"state": "M0_CONF_HA_PROCESS_STARTED", "type": "M0_CONF_HA_PROCESS_M0D"}
cortx-data-headless-svc-ssc-vm-g3-rhev4-2387/processes/0x7200000000000001:0x2a:{"state": "M0_CONF_HA_PROCESS_STARTED", "type": "M0_CONF_HA_PROCESS_M0D"}
cortx-data-headless-svc-ssc-vm-g3-rhev4-2387/processes/0x7200000000000001:0x31:{"state": "M0_CONF_HA_PROCESS_STARTED", "type": "M0_CONF_HA_PROCESS_M0D"}
```

and `get_node_processes()` fetches
```
m0conf/nodes/cortx-data-headless-svc-ssc-vm-g3-rhev4-2387/processes/52/meta_data:/dev/sdc
m0conf/nodes/cortx-data-headless-svc-ssc-vm-g3-rhev4-2387/processes/52/services/addb2:61
m0conf/nodes/cortx-data-headless-svc-ssc-vm-g3-rhev4-2387/processes/52/services/cas:62
m0conf/nodes/cortx-data-headless-svc-ssc-vm-g3-rhev4-2387/processes/52/services/fdmi:64
```
Both the keys are different. So we cannot use **get_node_processes()** here.
Even the prefix isn't the same for both keys... So is there any other way to use it?

utils/hare-status
Shreya-18: `get_service_data_by_name(service_name)` This function indeed returns details for that **particular service name** but on **all nodes**.

That would help for multiple processes with the same name like ios and s3, only. For rest, we need to make individual consul kv calls. 
Had it would return all the services on a particular node, would be actually helpful.
This is what I understand, please correct me.

> Here we are iterating on each consul kv,

Yes, correct. But we are parsing the data iteratively from the recurse key of 'm0conf/nodes/{node_name}/processes/',  which is only fetched once.
Also, we could optimize it and store all fid-> endpoint keys mapping in a dictionary once and use it.

mssawant: Okay, we can also get a node level details,
```
`[root@cortx-data-headless-svc-ssc-vm-g2-rhev4-1473 /]# curl http://127.0.0.1:8500/v1/catalog/node/cortx-data-headless-svc-ssc-vm-g2-rhev4-1473
{"Node":{"ID":"f85ebb3b-5eba-1879-7976-b899c975fb17","Node":"cortx-data-headless-svc-ssc-vm-g2-rhev4-1473","Address":"172.16.84.98","Datacenter":"dc1","TaggedAddresses":{"lan":"172.16.84.98","lan_ipv4":"172.16.84.98","wan":"172.16.84.98","wan_ipv4":"172.16.84.98"},"Meta":{"consul-network-segment":""},"CreateIndex":8299,"ModifyIndex":8306},"Services":{"259":{"ID":"259","Service":"hax","Tags":[],"Address":"inet:tcp:cortx-data-headless-svc-ssc-vm-g2-rhev4-1473","Meta":null,"Port":2001,"Weights":{"Passing":1,"Warning":1},"EnableTagOverride":false,"Proxy":{"MeshGateway":{},"Expose":{}},"Connect":{},"CreateIndex":8310,"ModifyIndex":8310},"262":{"ID":"262","Service":"ios","Tags":[],"Address":"inet:tcp:cortx-data-headless-svc-ssc-vm-g2-rhev4-1473","Meta":null,"Port":3001,"Weights":{"Passing":1,"Warning":1},"EnableTagOverride":false,"Proxy":{"MeshGateway":{},"Expose":{}},"Connect":{},"CreateIndex":8315,"ModifyIndex":8315},"275":{"ID":"275","Service":"ios","Tags":[],"Address":"inet:tcp:cortx-data-headless-svc-ssc-vm-g2-rhev4-1473","Meta":null,"Port":3002,"Weights":{"Passing":1,"Warning":1},"EnableTagOverride":false,"Proxy":{"MeshGateway":{},"Expose":{}},"Connect":{},"CreateIndex":8316,"ModifyIndex":8316},"288":{"ID":"288","Service":"confd","Tags":[],"Address":"inet:tcp:cortx-data-headless-svc-ssc-vm-g2-rhev4-1473","Meta":null,"Port":3003,"Weights":{"Passing":1,"Warning":1},"EnableTagOverride":false,"Proxy":{"MeshGateway":{},"Expose":{}},"Connect":{},"CreateIndex":8313,"ModifyIndex":8313},"291":{"ID":"291","Service":"s3service","Tags":[],"Address":"inet:tcp:cortx-data-headless-svc-ssc-vm-g2-rhev4-1473","Meta":null,"Port":4001,"Weights":{"Passing":1,"Warning":1},"EnableTagOverride":false,"Proxy":{"MeshGateway":{},"Expose":{}},"Connect":{},"CreateIndex":8317,"ModifyIndex":8317},"294":{"ID":"294","Service":"s3service","Tags":[],"Address":"inet:tcp:cortx-data-headless-svc-ssc-vm-g2-rhev4-1473","Meta":null,"Port":4002,"Weights":{"Passing":1,"Warning":1},"EnableTagOverride":false,"Proxy":{"MeshGateway":{},"Expose":{}},"Connect":{},"CreateIndex":8318,"ModifyIndex":8318}}}[root@cortx-data-headless-svc-ssc-vm-g2-rhev4-1473 /]#
```
this will give you endpoints for all the services on a given node. Idea it to avoid pattern matching because that incurs multilevel string comparisons and in worst case the string we are trying to find can be at the bottom of the list. If we can have some kind of an associative array it may help.

But, on the second thought, let's take this as an improvement and complete this first version.

utils/hare-status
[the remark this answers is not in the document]
mssawant: Okay, It would be good to avoid pattern matching as the cluster and the number keys grow, the data for pattern matching will increase, pattern matching can be expensive as well. If we have the information in a hashed format (associative array) then it becomes easier.

mssawant (approved):
Let's complete this version and we can take improvements later.

utils/hare-status
vaibhavparatwar: @mssawant  do we want to take it to main branch as well?

PR #1893, opened 2021-12-01
EOS-26729: Modify consul endpoint handling to connect to endpoint with tcp protocol

SwapnilGaonkar7:
Solution: Added check for tcp protocol

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -234,6 +234,9 @@ def _start_consul(utils: Utils,
     peers = []
     for endpoint in consul_endpoints:
         key = endpoint.split('/')
+        # Considering tcp endpoints only. Ignoring all other endpoints.
+        if key[0] != 'tcp:':
+            continue
         peer = ('/'.join(key[2:]))
         peers.append(peer)
 


provisioning/miniprov/hare_mp/main.py
```diff
@@ -234,6 +234,9 @@ def _start_consul(utils: Utils,
     peers = []
     for endpoint in consul_endpoints:
         key = endpoint.split('/')
+        # Considering tcp endpoints only. Ignoring all other endpoints.
```
kupranay, line 237: we should not assume that 0th position will always have tcp protocol. Code should be written in a way that should search for tcp url and use it. Tomorrow we may end up adding 3rd entry here and that should not cause a change in code

mssawant: @kupranay, we are iterating over the endpoints here and I don't see hardcoding to a certain position.

mssawant: If you are referring to `key[0]` that is about 1st part of endpoint after splitting the entry. Ofcourse, that may fail if the format of the endpoint changes, but is that expected, if so then we need to finalize the format now.

mssawant:
@SwapnilGaonkar7, is this PR still a draft?

provisioning/miniprov/hare_mp/main.py
kupranay: Oh... Got it.. i missed the for loop mentioned above. This logic looks good.
One minor suggestion... should we split baed on "/" or ":" and search for tcp only instead of "tcp:"

By the way i have no concern with split with "/"

kupranay (approved):

SwapnilGaonkar7:
> @SwapnilGaonkar7, is this PR still a draft?

For this PR you tried deployment(1247) and faced csm agent container failure.
Let me check with Pavan if we can retry deployment

pavankrishnat:
@SwapnilGaonkar7, deployment with [custom-build-1247 ](http://cortx-storage.colo.seagate.com/releases/cortx/github/integration-custom-ci/centos-7.9.2009/custom-build-1247/cortx_iso/) is 2 times successful ran s3 IO's. no issue observed on 9N setup

provisioning/miniprov/hare_mp/main.py
mssawant: We are using `/` as a delimiter to extract the endpoint, as we use that but the protocol we don't use, its just required to identify the correct endpoint.

mssawant (approved):

PR #1899, opened 2021-12-08
EOS-26607: Skip hare init stage(No-op) in s3server pod

rahul27kumar:
Signed-off-by: Rahul Kumar <rahul.kumar@seagate.com>

# Problem Statement
- As per new design in hare for motr-s3 pod separation [here](https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/811696538/WIP+POC+for+more-s3+pod+separation). There are no devices attached and not even a motr services running in s3server pod so we don't need to run init as mkfs. It is not required.

# Design
-  Added only condition for server_node on init function to skip init stage for s3server pods.

# Coding
   Checklist for Author
-  [x] Coding conventions are followed and code is consistent

# Testing 
  Checklist for Author
- [ ] Unit and System Tests are added
- [ ] Test Cases cover Happy Path, Non-Happy Path and Scalability
- [x] Testing was performed with RPM

# Impact Analysis
  Checklist for Author/Reviewer/GateKeeper
- We are going to commit this change (and all the other changes related to pod separation) in 'K8s_pod_separation' branch.

# Review Checklist 
  Checklist for Author
- [ ] JIRA number/GitHub Issue added to PR
- [ ] PR is self reviewed
- [ ] Jira and state/status is updated and JIRA is updated with PR link
- [ ] Check if the description is clear and explained

# Documentation
  Checklist for Author
- [ ] Changes done to WIKI / Confluence page / Quick Start Guide

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -29,6 +29,7 @@ import os
 import shutil
 import subprocess
 import sys
+import re
 from enum import Enum
 from sys import exit
 from time import sleep, perf_counter
@@ -38,6 +39,7 @@ from urllib.parse import urlparse
 
 import yaml
 from cortx.utils.product_features import unsupported_features
+from cortx.utils.cortx import Const
 from hax.common import di_configuration
 from hax.types import KeyDelete, Fid
 from hax.util import ConsulUtil, repeat_if_fails, KVAdapter
@@ -143,6 +145,28 @@ def get_server_type(url: str) -> str:
     except Exception as error:
         logging.error('Cannot get server type (%s)', error)
         return 'unknown'
+
+
+def is_mkfs_required(url: str) -> bool:
+    try:
+        provider = ConfStoreProvider(url)
+        machine_id = provider.get_machine_id()
+        # Query results into a list of keys where, Const.SERVICE_MOTR_IO
+        # is found as one of the services under `services`.
+        # e.g.
+        # node>aaa120a9e051d103c164f605615c32a4>components[1]>services[0]
+        # node>bbb340f79047df9bb52fa460615c32a5>components[1]>services[0]
+        # node>ccc8700fe6797ed532e311b0615c32a7>components[1]>services[0]
+        # We try to find this node's machine id in the query result.
+        services = provider.search_val('node', 'services',
+                                       Const.SERVICE_MOTR_IO.value)
+        check_id = re.compile(f".*{machine_id}.*")
+        return any(check_id.match(service) for service in services)
+    except Exception as error:
+        logging.warn('Failed to get pod type (%s). Current stage will '
+                     'be assumed as not required by default', error)
+
+        return False
 
 
 def logrotate_generic(url: str):
@@ -482,6 +506,10 @@ def set_mkfs_done_for(node: str, cns_utils: ConsulUtil):
 def init(args):
     try:
         url = args.config[0]
+
+        if not is_mkfs_required(url):
+            return
+
         utils = Utils(ConfStoreProvider(url))
         cns_utils = ConsulUtil()
         stop_event = Event()


provisioning/miniprov/hare_mp/main.py
```diff
@@ -449,6 +449,10 @@ def init(args):
     try:
         url = args.config[0]
         utils = Utils(ConfStoreProvider(url))
+        provider = ConfStoreProvider(url)
```
knekrasov, line 452: Can this be refactored into a form of the kind?
```
if not stage_required(url):
    return
```

provisioning/miniprov/hare_mp/main.py
```diff
@@ -449,6 +449,10 @@ def init(args):
...
+        machine_id = provider.get_machine_id()
+        if provider.get(f'node>{machine_id}>type') == 'server_node':
+           exit(0)
```
knekrasov, line 455: Let's stop using `exit()` in functions outside of `main()`. One day we'll want to cover this code with unit tests and those voluntary `exit()` invocations will be a problem.
Please consider either just returning or raising an exception (which will turn into `sys.exit(error_code)` in `main` function).

knekrasov (changes requested):

provisioning/miniprov/hare_mp/main.py
```diff
@@ -144,6 +144,15 @@ def get_server_type(url: str) -> str:
         logging.error('Cannot get server type (%s)', error)
         return 'unknown'
 
+def is_stage_required(url: str) -> str:
+    try:
+        provider = ConfStoreProvider(url)
+        machine_id = provider.get_machine_id()
+        pod_type = provider.get(f'node>{machine_id}>type')
+        return pod_type == 'server_node'
```
mssawant, line 152: It is now decided that pod_type or node_type is a dynamic key and we need to use constants from https://github.com/Seagate/cortx-utils/pull/640/files.
We can use `SERVICE_MOTR_IO` macro to search for io services, thus, if ioservices are configured on a node then execute mkfs else not. We need to use conf-store's search api as below,
e.g.
```
Conf.search(self.default_index, parent_key, key_to_search, value_to_search)
```

provisioning/miniprov/hare_mp/main.py
```diff
@@ -144,6 +144,15 @@ def get_server_type(url: str) -> str:
         logging.error('Cannot get server type (%s)', error)
         return 'unknown'
 
+def is_stage_required(url: str) -> str:
```
knekrasov, line 147:
```suggestion
def is_stage_required(url: str) -> bool:
```

provisioning/miniprov/hare_mp/main.py
```diff
@@ -144,6 +144,15 @@ def get_server_type(url: str) -> str:
...
+    try:
+        provider = ConfStoreProvider(url)
+        machine_id = provider.get_machine_id()
+        pod_type = provider.get(f'node>{machine_id}>type')
+        return pod_type == 'server_node'
+    except Exception as error:
```
knekrasov, line 153: It doesn't make much sense to return 'unknown' string: 
1. It is very easy to forget to check the return code in the calling code
2. The function name gives a clue that the return type will be boolean (i.e. the return value is understood either as yes or no). Thus if somebody uses the function like this 
```python
if is_stage_required(url):
    do_smth()
``` 
then `do_smth()` *will be executed* in case of an exception. Reason: non-empty string (like `'unknown'` is) evaluates to True.

Suggestion:
Raise an exception when something unexpected happens, don't rely on return codes. In this case the code will be kept clean (no paranoid "if error" checks) but on the other hand if an error happens, it will not be swallowed because of a human mistake.

provisioning/miniprov/hare_mp/main.py
```diff
@@ -144,6 +147,17 @@ def get_server_type(url: str) -> str:
         logging.error('Cannot get server type (%s)', error)
         return 'unknown'
 
+def is_stage_required(url: str) -> bool:
```
mssawant, line 150: The function name is very generic, however, we are using it for a specific purpose to identify if mkfs must be executed on this node or not.
I think, let's rename the function accordingly, `is_mkfs_required()`

provisioning/miniprov/hare_mp/main.py
knekrasov: @mssawant Please look at my comment regarding this call [here](https://github.com/Seagate/cortx-hare/pull/1904#discussion_r769388348).

Summary: I suggest not using this API call. It doesn't help us much anyway (so @rahul27kumar has to filter out the resulting keys manually afterwards - and that's said to be "expected").

provisioning/miniprov/hare_mp/main.py
```diff
@@ -144,6 +147,24 @@ def get_server_type(url: str) -> str:
...
+        # ['node>aaa120a9e051d103c164f605615c32a4>components[1]>services[0]', 
+        # 'node>bbb340f79047df9bb52fa460615c32a5>components[1]>services[0]',
+        # 'node>ccc8700fe6797ed532e311b0615c32a7>components[1]>services[0]']
+        # We try to find this node's machine id in the query result.
+        services = provider.search_val('node' , 'services',
+                                          Const.SERVICE_MOTR_IO.value)
+        check_id = re.compile(f".*{machine_id}.*")
+        return any(check_id.match(service) for service in services)
+    except Exception as error:
+        logging.error('Cannot get stage type (%s)', error)
```
knekrasov, line 166: It's not an error rather a warning. 
```suggestion
        logging.warn('Failed to get pod type (%s). Current stage will '
                     'be assumed as not required by default', error)

```

provisioning/miniprov/hare_mp/main.py
```diff
@@ -449,18 +470,19 @@ def init(args):
...
+           log_dir = get_log_dir(url)
+           # Starting consul and hax
+           consul_starter = _start_consul(utils, stop_event,
+                                          config_dir, log_dir, url)
+           hax_starter = _start_hax(utils, stop_event, config_dir, log_dir)
+           hostname = utils.get_local_hostname()
+           start_mkfs_parallel(hostname, config_dir)
+           # Stopping hax and consul
+           hax_starter.stop()
+           consul_starter.stop()
```
knekrasov, line 485: [Optional]: if you invert the boolean expression, no nesting will be required (the code becomes simpler).

Update: while preparing this suggestion, I noticed that you use 3-space indentation. Please use 4 spaces as a single indent instead in your code.

```suggestion
        if not is_mkfs_required(url):
            return

        stop_event = Event()
        config_dir = get_config_dir(url)
        log_dir = get_log_dir(url)
        # Starting consul and hax
        consul_starter = _start_consul(utils, stop_event,
                                       config_dir, log_dir, url)
        hax_starter = _start_hax(utils, stop_event, config_dir, log_dir)
        hostname = utils.get_local_hostname()
        start_mkfs_parallel(hostname, config_dir)
        # Stopping hax and consul
        hax_starter.stop()
        consul_starter.stop()
```

knekrasov (changes requested):

mssawant:
Changes don't seem much, but it would be good to try deployment. cc @vaibhavparatwar

provisioning/miniprov/hare_mp/main.py
mssawant: @knekrasov, Problem is, as per the provisioning team, there are some keys that are volatile and some constant. Provisioning team has introduced a list of constants that we can use, so atleast we have a contract in some form.
So, it is suggested to search and extract the information from conf-store rather than accessing a keys directly.

vaibhavparatwar (approved):

mssawant:
@rahul27kumar,  kindly squash the commits and update accordingly.

mssawant:
@rahul27kumar, Is the deployment tested?

PR #1900, opened 2021-12-08
EOS-26779: Logging configuration in miniprovisioner needs

bdekvadiya:
Problem:
            By default the stream in setup_logging is sys.stderr
            which is populating the logs in the stderr. stderr
            indicates the log type as error. So all logs should
            be populated on stdout. Error can be indicated by
            return value, instead of log type as stderr.

Solution:
	Changed stream from sys.stderr to sys.stdout in setup_logging

Signed-off-by: Bhargav Dekivadiya <bhargav.dekivadiya@seagate.com>

# Problem Statement
- Problem statement

# Design
-  For Bug, Describe the fix here.
-  For Feature, Post the link for design

# Coding
   Checklist for Author
-  [ ] Coding conventions are followed and code is consistent

# Testing 
  Checklist for Author
- [ ] Unit and System Tests are added
- [ ] Test Cases cover Happy Path, Non-Happy Path and Scalability
- [ ] Testing was performed with RPM

# Impact Analysis
  Checklist for Author/Reviewer/GateKeeper
- [ ] Interface change (if any) are documented
- [ ] Side effects on other features (deployment/upgrade)
- [ ] Dependencies on other component(s)

# Review Checklist 
  Checklist for Author
- [ ] JIRA number/GitHub Issue added to PR
- [ ] PR is self reviewed
- [ ] Jira and state/status is updated and JIRA is updated with PR link
- [ ] Check if the description is clear and explained

# Documentation
  Checklist for Author
- [ ] Changes done to WIKI / Confluence page / Quick Start Guide

### DIFF
--- a/provisioning/miniprov/hare_mp/main.py
+++ b/provisioning/miniprov/hare_mp/main.py
@@ -78,10 +78,10 @@ def create_logger_directory(log_dir):
     if not os.path.isdir(log_dir):
         try:
             os.makedirs(log_dir)
-        except Exception:
+        except Exception as e:
             logging.exception(f"{log_dir} Could not be created")
             shutdown_cluster()
-            exit(-1)
+            raise RuntimeError("Failed to create log directory " + str(e))
 
 
 def setup_logging(url) -> None:
@@ -93,7 +93,7 @@ def setup_logging(url) -> None:
 
     create_logger_directory(log_dir)
 
-    console = logging.StreamHandler(stream=sys.stderr)
+    console = logging.StreamHandler(stream=sys.stdout)
     fhandler = logging.handlers.RotatingFileHandler(log_file,
                                                     maxBytes=LOG_FILE_SIZE,
                                                     mode='a',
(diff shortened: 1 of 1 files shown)


cortx-admin:

provisioning/miniprov/hare_mp/main.py
```diff
@@ -93,7 +93,7 @@ def setup_logging(url) -> None:
 
     create_logger_directory(log_dir)
 
-    console = logging.StreamHandler(stream=sys.stderr)
+    console = logging.StreamHandler(stream=sys.stdout)
```
mssawant, line 96: I think instead of just changing stderr to stdout, where all the error as well as regular logs will be re-directed to just one stream, we need to have a way to log errors/exceptions to stderr while regular logs to stdout.
This can be done by having 2 logging streams,  1 for stderr and other for stdout.
A single wrapper class around these logging streams can be created to log messages to the respective streams.

knekrasov: @mssawant I think it's an overkill.
As I had written earlier, there is no specific rule that writing to stderr means a failure. It means that the application is free to choose where to write. A failure is a failure if and only if the exit code is non-zero, everything else depends on use cases of the particular application.

What you suggest is a nice to have but I'm pretty sure that it requires 1-2 days of investigation how to tune the logging subsystem in that fine manner. Is it worthwhile?

knekrasov (approved):

provisioning/miniprov/hare_mp/main.py
mssawant: In that case we need to make sure that every exception is handled such that a non-zero error code is returned to the caller, it should not happen that an exception is raised but the operation was a success. If we guarantee that, then I am okay with the current implementation.

mssawant:
@bdekvadiya, deployment needs to be tested with this PR. cc @vaibhavparatwar

provisioning/miniprov/hare_mp/main.py
bdekvadiya: I have tested the deployment with all the changes and populated the result on this Jira : https://jts.seagate.com/browse/EOS-26779

bdekvadiya:
> @bdekvadiya, deployment needs to be tested with this PR. 

I have tested the PR changes and results are populated here : https://jts.seagate.com/browse/EOS-26779

SwapnilGaonkar7 (approved):

mssawant (approved):

Issue #1779, opened 2021-09-05
RuntimeError: Unsupported facter version found at node localhost: 2.4.1 . Please use 3.14.8 or higher.


hessiotest:
When I try to bootstrap my single node cluster I get this error:

`[root@localhost hare]# hctl bootstrap --mkfs ~/CDF.yaml
2021-09-05 10:32:53: Generating cluster configuration...Traceback (most recent call last):
  File "/opt/seagate/cortx/hare/bin/../bin/cfgen", line 2164, in <module>
    main()
  File "/opt/seagate/cortx/hare/bin/../bin/cfgen", line 216, in main
    enrich_cluster_desc(cluster_desc, opts.mock)
  File "/opt/seagate/cortx/hare/bin/../bin/cfgen", line 309, in enrich_cluster_desc
    ipaddr_key(node['data_iface']))
  File "/opt/seagate/cortx/hare/bin/../bin/cfgen", line 549, in get_facts
    validate_facter(hostname)
  File "/opt/seagate/cortx/hare/bin/../bin/cfgen", line 542, in validate_facter
    f'Unsupported facter version found at node {hostname}: {ver_str}. '
RuntimeError: Unsupported facter version found at node localhost: 2.4.1
. Please use 3.14.8 or higher.
`

vaibhavparatwar:
@hessiotest please refer README in which puppet agent installation (required version) steps are given which in turn will instal required facter version.. 
![image](https://user-images.githubusercontent.com/65211391/132158657-650d81db-5969-4dc7-b7cc-3b9297bce923.png)

https://github.com/Seagate/cortx-hare#readme
Please let me know if it helps

hessio:
Yes I followed the steps correctly and then I tried the previous commit before this one and the installation worked for me. I think we need to roll back this commit.

vaibhavparatwar:
@hessio This commit was made for some reason. I think one more change we need to do is add a step in readme which will check if already Facter is installed and if yes, what version. If its lesser than 3.14.8 then, we need to remove existing older version.

hessio:
Oh ok cool, we can do this, but maybe we revert this commit until it is working completely? Because if you follow the guide as is now you will get an error, what do you think?

vaibhavparatwar:
Actually the issue comes only on the systems where the facter is already installed .. so its not that it fails 100% of the time.. @pavankrishnat  already started working on making changes to README. let me know if that helps?

hessio:
[...]

PR #1932, opened 2022-01-04
README: update about facter installation on el7

andriytk:
`yum erase -y $(rpm -q --whatprovides $(readlink -f /usr/bin/facter))`
removes existing facter rpm along with the motr rpm, which depends on it.

Solution: don't erase existing facter rpm, the facter binary will be
overwritten by the following `ln -sf` command.

-----
[View rendered README.md](https://github.com/Seagate/cortx-hare/blob/andriytk-patch-1/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,6 @@ and health-checking mechanisms.
   On CentOS 7, if the facter version is less than 3.14 or not installed, then run the following commands to update/install it:
   
   ```sh
-  sudo yum erase -y $(rpm -q --whatprovides $(readlink -f /usr/bin/facter)) || sudo rm -fv /usr/bin/facter
   sudo yum localinstall -y https://yum.puppetlabs.com/puppet/el/7/x86_64/puppet-agent-7.0.0-1.el7.x86_64.rpm
   sudo ln -sf /opt/puppetlabs/bin/facter /usr/bin/facter
   ```


README.md
```diff
@@ -66,7 +66,6 @@ and health-checking mechanisms.
   On CentOS 7, if the facter version is less than 3.14 or not installed, then run the following commands to update/install it:
   
   ```sh
-  sudo yum erase -y $(rpm -q --whatprovides $(readlink -f /usr/bin/facter)) || sudo rm -fv /usr/bin/facter
```
suvratjoshi, line 69: I think this was done for specific case of existing ver less that 3.14.
I think @pavankrishnat , when was testing this particular case, he was facing some issue, that is why I guess we had added removal part here. @pavankrishnat can confirm if my understanding is correct.

@andriytk ,
Just to confirm, was this specific case tested, where we don't erase existing ver which is less than 3.14 and just run next steps?

pavankrishnat: @suvratjoshi/ @andriytk 
I added this line of code to fix #1779

andriytk: Yes, I tested it. The yum command deletes motr rpm! And if we don't do it, `ln -sf` cmd just replaces facter at `/usr/bin` with the version from puppet. Hare works fine then.

The problem with the puppet package is that it does not declare that it provides facter, so when we erase facter rpm - it affects motr, as motr rpm has a dependency on facter.

This ugly way of solving the problem will be gone in CentOS 8, which has 3.14 facter version rpm in the base repo - the one we need.

suvratjoshi (approved):

suvratjoshi:
pre-merge test failing is expected as it is already failing for all PRs currently
Not sure why deployment is failing here though changes are just in readme file.

andriytk:
Just FYI, I faced this issue while working on https://github.com/Seagate/cortx-motr/issues/1356.

andriytk:
Due to the inability to update the commit in the `andriytk-patch-1` branch, created another PR - https://github.com/Seagate/cortx-hare/pull/1937.

PR #1948, opened 2022-01-12
EOS-27301: Problem: Find and fix all URLs in hare code

supriyachavan4398:
Hyperlinks for the cortx-hare/rfc/ are broken

Solution:
Added correct absolute paths for every broken links.

Signed-off-by: Supriya Yadav <supriya.s.chavan@seagate.com>


mssawant (approved):

suvratjoshi:
Please use capital start in PR description and commit message
> EOS-27301: find and fix all URLs in hare code

rfc/12/README.md
```diff
@@ -81,4 +81,4 @@ There are 2 options for health checks integration into Pacemaker:
 
 ## See also
 
-* [Original list of LDR1 HA health checks. The spreadsheet has a lot of useful information](https://docs.google.com/spreadsheets/d/1xASlPnlFx1OmhKttbgOweHbmfbXMI6n3Pmp2TSO8ou8/edit#gid=1305517710)
+* [Original list of LDR1 HA health checks. The spreadsheet has a lot of useful information](https://seagatetechnology.sharepoint.com/:x:/r/sites/gteamdrv1/tdrive1224/LDR%20R1/HA/HA%20requirements%202020-01-14%20.xlsx?d=wdf76a255f0b84af5a771744ef00440d5&csf=1&web=1&e=gKXtZM)
```
suvratjoshi, line 84: I just checked :
These sharepoint links are not accessible from outside seagate devices.
Either sharing settings need to be changed or the docs are needed to be converted to publically accessible formats (github or confluence docs)

suvratjoshi, line 84: If the sharing path that is used currently is from its original location (which may not be in Hare project directory), then document can get vanished if somebody deletes it from original location.
Were the documents moved to sharepoint's Hare directory and then sharing link is created?

suvratjoshi, line 84: Can we try out techniques of embedding excel and ppt documents in confluence page, as shown in these videos?
https://www.youtube.com/watch?v=MhwKcD-Z83w
https://www.youtube.com/watch?v=1Ubm0DlnbVU

If it works then we can have our excel and ppt documents in confluence pages.
We can create a confluence page with title, what the sheet or ppt doc has, and embed the document in that confluence page.
Hopefully that will work.

supriyachavan4398: @suvratjoshi  As per yesterday's tech discussion, It discussed that we only create, confluence pages for design documents/Integration documents.

We don't need to convert these requirement-related documents. So as per discussion, I added it into this shared location https://seagatetechnology.sharepoint.com/sites/gteamdrv1/tdrive1224/LDR%20R1/Forms/AllItems.aspx?id=%2Fsites%2Fgteamdrv1%2Ftdrive1224%2FLDR%20R1%2FHA&viewid=c3aa8d53%2D081b%2D4b29%2D82ee%2D76108af397de
cc @mssawant @pavankrishnat

suvratjoshi: Thats ok.
Then we need to have method to share excel/ppt docs with outside world.

supriyachavan4398: Yes checking for that.

suvratjoshi: Other way to make docs accessible is by creating a directory named something like "ref_docs" and put excel and ppt like docs in that dir and link to those.

mssawant: @suvratjoshi, this is a different problem to solve. we need check if the linked documents must be made available to the community or not. If they are planning or estimation related documents then I don't think community needs to have access to it and they are internal. Let's hold on to it until it is required by the community.

suvratjoshi: Right @mssawant. Agree.
That is what even I was wondering.
For community those will not be needed mostly. But then in that case the link becomes invalid for community (as it shows Seagate login screen). If link still remains invalid for community, then we are still not fixing links in a way, fully.
Somebody will again file a github issue for that.

Should we 
1. remove these document refs from github pages and if required we can point those docs from our internal confluence pages where ever required, 
OR
2. just copy the documents to Hare sharepoint folders and no pointers to them for now from anywhere.

I think 2nd option is better.

vaibhavparatwar: Agree with @suvratjoshi . @supriyachavan4398 pleaes remove "See also" section as per our discussion so we get unblocked for this PR and task. cc @mssawant @SunilSavanurSeagate

supriyachavan4398: Addressed.

suvratjoshi (approved):

rfc/12/README.md
mssawant: This is not a Hare document but HA. We must not move other component documents in Hare directory. If documents are vanishing from the folders, then this needs to be reported to IT and to all the teams. May be such URLs can be removed from the document if not required. But this can figured out later. Let's just fix the links in this PR's scope.

PR #1952, opened 2022-01-17
cfgen: default ConfDrive object is not created for CAS service type

mssawant:
Hare cfgen create motr configuration object instances corresponding to
motr data and metadata devices associated with motr ioservice and CAS
service types. If an explicitly metadata device is not specified in Hare
CDF, then cfgen does not create a default metadata device for CAS service
type.

Solution:
Create a default ConfDrive object corresponding to `/dev/null` device for
motr CAS service type.

### DIFF
--- a/cfgen/cfgen
+++ b/cfgen/cfgen
@@ -1128,18 +1128,16 @@ class ConfProcess(ToDhall):
 
             if stype is SvcT.M0_CST_CAS:
                 assert proc_desc is not None
-                if meta_data is None:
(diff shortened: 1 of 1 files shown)


supriyachavan4398 (approved):
Looks good to me.

vaibhavparatwar (approved):

andriytk:
Shouldn't motr metadata device always be /dev/null? cc @madhavemuri .

PR #1966, opened 2022-01-31
EOS-27971:Problem: subsequent entrypoint requests are not received from motr

mssawant:
Hare, on receiving entrypoint request from a motr process fetches the
required information from Consul in-order to prepare the entrypoint
reply. If the required information is not available, Hare drops the
entrypoint requestsilently, expecting motr would timeout and resend
the entrypoint request. This mechanism was introduced to reduce the
number of entrypoint requests received from a motr process when Hare
replies with EAGAIN.

Solution:
Add a delay of 1 second before replying to entrypoint request in case
of a failure for now.

### DIFF
--- a/hax/hax/motr/__init__.py
+++ b/hax/hax/motr/__init__.py
@@ -17,8 +17,10 @@
 #
 
 import ctypes as c
+from errno import EAGAIN
 import logging
 from typing import Any, List, Optional, Tuple
+from time import sleep
 
 from hax.consul.cache import supports_consul_cache, uses_consul_cache
 from hax.exception import ConfdQuorumException, RepairRebalanceException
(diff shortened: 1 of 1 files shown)


vaibhavparatwar:
@mssawant  can we link it to any JIRA ID? and i see 1N deployment check is failing for this PR.

SwapnilGaonkar7 (approved):
Looks good.
I think we have already tested this approach.

suvratjoshi:
retest this please

suvratjoshi:
above comment was to auto trigger 1N deployment
no action needed from anybody

mssawant:
@vaibhavparatwar, 1N deployment is passing now.

SwapnilGaonkar7 (approved):

pragamj:
hctl status on 15
[hctl_status_15N.txt](https://github.com/Seagate/cortx-hare/files/8012919/hctl_status_15N.txt)
N

vaibhavparatwar (approved):

PR #1994, opened 2022-02-24
cfgen: cfgen execution fails outside of hctl

mssawant:
cfgen script imports some of the hax modules and thus require to have
hax modules path accessible if executed explicitly post installation.
e.g.
```
[root@ssc-vm-g4-rhev4-0554 ~]# /opt/seagate/cortx/hare/bin/cfgen
  File "/opt/seagate/cortx/hare/bin/cfgen", line 17, in <module>
    import psutil
ModuleNotFoundError: No module named 'psutil'
```

Solution:
Add hax module paths for imports to be available to cfgen script as
part of installation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -256,6 +256,7 @@ install: install-dirs install-cfgen install-hax install-miniprov install-systemd
 	@$(call _fix_hare_imports,configure)
 	@$(call _fix_hare_imports,m0ping)
 	@$(call _fix_hare_imports,update-conf)
+	@$(call _fix_hare_imports,cfgen)
 
 .PHONY: install-dirs
 install-dirs:
@@ -312,8 +313,9 @@ install-systemd: $(wildcard systemd/*)
 
 .PHONY: install-hax
 install-hax: HAX_INSTALL_CMD = $(PIP) install --ignore-installed --prefix $(DESTDIR)/$(PREFIX) $(HAX_WHL:hax/%=%)
-install-hax: export PYTHONPATH = $(DESTDIR)/$(PREFIX)/lib/python3.$(PY3_VERSION_MINOR)/site-packages
+install-hax: export PYTHONPATH = $(DESTDIR)/$(PREFIX)/lib64/python3.$(PY3_VERSION_MINOR)/site-packages
 install-hax: $(HAX_EXE)
+	@cd hax && $(SETUP_PY) install
 
 $(HAX_EGG_LINK) $(HAX_EXE): $(HAX_WHL)
 	@$(call _info,Installing hax with '$(HAX_INSTALL_CMD)')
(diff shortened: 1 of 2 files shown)


mssawant:
```
[root@ssc-vm-g4-rhev4-0554 ~]# /opt/seagate/cortx/hare/bin/cfgen
usage: cfgen [OPTION]... CDF
cfgen: error: the following arguments are required: CDF
[root@ssc-vm-g4-rhev4-0554 ~]# /opt/seagate/cortx/hare/bin/cfgen --help-schema
# Cluster Description is a YAML file with the following schema:
---  # start of the document (optional)
create_aux: <bool> # optional, defaults to false if not present in CDF
nodes:
  - hostname: <str>    # [user@]hostname; e.g., localhost, pod-c1
    data_iface: <str>  # name of network interface; e.g., eth1, eth1:c1
    data_iface_type: tcp|o2ib  # type of network interface;
                               # optional, defaults to "tcp"
    m0_servers:        # optional for client-only nodes
      - runs_confd: <bool>  # optional, defaults to false
        io_disks:
          meta_data: <str>  # device path for meta-data;
                            # optional, Motr will use "/var/motr/m0d-<FID>/"
                            # by default
          data: [ <str> ]   # e.g. [ "/dev/loop0", "/dev/loop1", "/dev/loop2" ]
                            # Empty list means no IO service.
    m0_clients:
        s3: <int>     # number of S3 servers to start
        other: <int>  # max quantity of other Motr clients this host may have
pools:
  - name: <str>
    type: sns|dix|md   # optional, defaults to "sns";
                       # "sns" - data pool, "dix" - KV, "md" - meta-data pool.
    disk_refs:  # optional section
      - path: <str>  # io_disks.data value
        node: <str>  # optional; 'hostname' of the corresponding node
    data_units: <int>
    parity_units: <int>
    spare_units: <int>
    allowed_failures:  # optional section; no failures will be allowed
                       # if this section is missing or all of its elements
                       # are zeroes
      site: <int>
      rack: <int>
      encl: <int>
      ctrl: <int>
      disk: <int>

# Profile is a reference to pools.  When Motr client is started, it receives
# profile fid from the command line.  Motr client can only use the pools
# referred to by its profile.
#
profiles:  # This section is optional.  If it is missing, a single "default"
           # profile referring to all pools will be created.
  - name: <str>
    pools: [ <str> ]

# FDMI filter defines if an FDMI record would be sent from FDMI source to FDMI
# plugin. Only M0_FDMI_FILTER_TYPE_KV_SUBSTRING filters are supported.
#
fdmi_filters:  # This section is optional.  If it is missing, no fdmi filters
               # would be defined
  - name: <str>  # Human-readable name of the filter
    node: <str>  # Hostname of the node that hosts FDMI plugin
                 # that would receive every FDMI record this filter matches
    client_index: <int>  # 0-based index of the client that would be
                         # the FDMI plugin. Only "other" clients could be FDMI
                         # plugins. "s3" clients are not used to for this
                         # index.
                         # The client has to be defined in m0_clients section
                         # already.
    substrings: [ <str> ]  # List of strings for FDMI plugin to match
                           # See m0_conf_fdmi_filter::ff_substrings
                           # for the reference. May be empty or absent.
...  # end of the document (optional)
[root@ssc-vm-g4-rhev4-0554 ~]#
```

SwapnilGaonkar7 (approved):

vaibhavparatwar (approved):

PR #1995, opened 2022-02-24
CORTX-27822: RGW-Modify ‘hctl fetch-fids’ to provide endpoints and fids for motr client

SwapnilGaonkar7:
**Solution**: Modified 'hctl fetch-fids' to include info present in sysconfig files

**Test:**
```
[root@cortx-data-headless-svc-ssc-vm-g4-rhev4-0635 temp]# hctl fetch-fids -c /etc/cortx/hare/config/2b334e7a459a4a07a155eceffed9bf80/ --all
[
  {
    "name": "cortx-data-headless-svc-ssc-vm-g4-rhev4-0635",
    "svcs": [
      {
        "name": "hax",
        "fid": "0x7200000000000001:0x7",
        "ep": "inet:tcp:cortx-data-headless-svc-ssc-vm-g4-rhev4-0635@2001"
      },
      {
        "name": "ioservice",
        "fid": "0x7200000000000001:0xa",
        "ep": "inet:tcp:cortx-data-headless-svc-ssc-vm-g4-rhev4-0635@3001",
        "ha_ep": "inet:tcp:cortx-data-headless-svc-ssc-vm-g4-rhev4-0635@2001",
        "uuid": "2f9b1e68-9625-11ec-82c0-a21cfcdd6585",
        "be_seg_path": "/dev/sdc"
      },
      {
        "name": "ioservice",
        "fid": "0x7200000000000001:0x19",
        "ep": "inet:tcp:cortx-data-headless-svc-ssc-vm-g4-rhev4-0635@3002",
        "ha_ep": "inet:tcp:cortx-data-headless-svc-ssc-vm-g4-rhev4-0635@2001",
        "uuid": "2f9b1e68-9625-11ec-82c0-a21cfcdd6585",
        "be_seg_path": "/dev/sdf"
      },
      {
        "name": "confd",
        "fid": "0x7200000000000001:0x28",
        "ep": "inet:tcp:cortx-data-headless-svc-ssc-vm-g4-rhev4-0635@3003",
        "ha_ep": "inet:tcp:cortx-data-headless-svc-ssc-vm-g4-rhev4-0635@2001",
        "conf_xc": "/etc/motr/confd.xc",
        "uuid": "2f9b1e68-9625-11ec-82c0-a21cfcdd6585"
      }
    ],
    "ha_ep": "inet:tcp:cortx-data-headless-svc-ssc-vm-g4-rhev4-0635@2001",
[...]

PR #2005, opened 2022-03-01
CORTX-27822: ‘hctl fetch-fids’ does not report motr client information

SwapnilGaonkar7:
Solution:
Added motr client support in 'hctl fetch-fids' to report motr client endpoints and fids.


SwapnilGaonkar7:
@mssawant I have addressed your review comments from https://github.com/Seagate/cortx-hare/pull/1995 here

Updated commit messages

`return type hint would be good.`
Done

`Not sure I understand, so if service is not hax and ioservices and confd then it is expected to be a motr_client service, we are then trying to fetch motr_client service information but get_motr_client_svc() returns ha endpoint and profile_fid, so as a reader I don't really understand exactly what information we are populating.`
I have modified code to check for clients(and not assuming).
The idea is that name, fid and endpoints are common for all services
`svc.update({'name': svc_name, 'fid': fid, 'ep': ep})`
And other parameters are apeended based on type thats why checking type calling respective function.
In future if more info for specific type of generic is required it can be added in there respective function or above svc,update

vaibhavparatwar (approved):

PR #2023, opened 2022-03-11
CORTX-28870: Alex scan generate false positive

venkuppu-chn:
Alex scan may report certain words to be prohibited from there use in various documentation or code. For example,

easy might be reported as insensitive. Thus we want to avoid such false positives.

**Solution:**

Add relevant words to the list to avoid such false positives as part of Alex scan configuration.

Signed-off-by: Venkatesh K <venkatesh.k@seagate.com>

### DIFF
--- a/.alexrc
+++ b/.alexrc
@@ -1,0 +1,3 @@
+{
+  "allow": ["execute", "host-hostess", "remains", "reject", "traps", "period", "invalid", "execution", "simple", "failure"]
+}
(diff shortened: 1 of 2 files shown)


vaibhavparatwar:
@venkuppu-chn Please add appropriate commit message from which it will be easy for reviewers to understand what this PR is about. cc @mssawant @SwapnilGaonkar7

mssawant:
okay, but I still don't see a proper commit message, an example would be good for the false positives that were generated for which you are adding the fix.

mssawant (approved):

PR #2049, opened 2022-04-07
Problem: CORTX-29866 Make motr server and client processes port range configurable per process type

SwapnilGaonkar7:
**Solution:**
Modified 'NetworkPorts' to include list of endpoints for clients as well as ios and confd.
cfgen will use endpoint according to type of process

```
network_ports:
  m0_server:
  - 29000
  - 22000
  m0_client_other:
  - name: rgw
    port: 21000
  m0_client_s3: null
  hax: 22000
```

**Testing in progress**

### DIFF
--- a/cfgen/cfgen
+++ b/cfgen/cfgen
@@ -4,7 +4,7 @@ import abc
 import argparse
 import ast
 import sys
-from enum import Enum, auto, IntEnum
+from enum import Enum, auto
 from functools import lru_cache, wraps
 import itertools
 import json
@@ -430,7 +430,7 @@ def enrich_cluster_desc(desc: Dict[str, Any], mock_p: bool) -> None:
                                  for port_type in portalgroup}
     else:
         for key, value in desc['network_ports'].items():
-            if value is None:
+            if not value:
                 desc['network_ports'][key] = portalgroup[key].value
         for srv in [port_type.name for port_type in portalgroup]:
             if srv not in desc['network_ports']:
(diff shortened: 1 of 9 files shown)


cfgen/cfgen
```diff
@@ -848,44 +848,58 @@ def oids_to_dhall(oids: List[Oid]) -> str:
 NetProtocol = Enum('NetProtocol', 'tcp o2ib')
 
 
-ProcT = Enum('ProcT', 'hax m0_server m0_client_s3 m0_client_other')
+ProcT = Enum('ProcT', 'hax m0_ios m0_confd m0_client_s3 m0_client_other')
```
mssawant, line 851: I was thinking of avoiding further classification of m0_server. Can we use m0ProcT instead?

SwapnilGaonkar7: Done

provisioning/miniprov/hare_mp/dhall/gencdf.dhall
```diff
@@ -26,6 +26,11 @@ let M0ClientDesc =
       , instances : Natural
       }
 
+let ClientPort =
```
mssawant, line 29: How about having something similar for m0server, name and port, instead of just port? We can then use names to map to the service types in Hare.

vaibhavparatwar: @SwapnilGaonkar7 is this comment take care?

vaibhavparatwar (approved):

provisioning/miniprov/hare_mp/dhall/gencdf.dhall
SwapnilGaonkar7: Done

mssawant (approved):
Looks good, thanks.

PR #2097, opened 2022-05-31
First phase changes to support various directories creation under use…

vimleshsgt:
…r home

	and systemctl services running under normal user session instead of root.

Signed-off-by: Vimlesh Kumar <vimlesh.kumar@seagate.com>


cortx-admin:

PR #2134, opened 2022-07-05
CORTX-32325: Remove version dependency between motr and hax

rkothiya:
Removing rpm level version dependency between motr and hax
which will help reduce build time.

Signed-off-by: Rinku Kothiya <rinku.kothiya@seagate.com>

### DIFF
--- a/hare.spec
+++ b/hare.spec
@@ -18,9 +18,6 @@
 # build number
 %define h_build_num  %(test -n "$build_number" && echo "$build_number" || echo 1)
 
-# motr version
-%define h_motr_version %(rpm -q --whatprovides cortx-motr | xargs rpm -q --queryformat '%%{VERSION}-%%{RELEASE}')
-
 # parallel build jobs
 %define h_build_jobs_opt  %(test -n "$build_jobs" && echo "-j$build_jobs" || echo '')
 
@@ -56,7 +53,7 @@ Requires: puppet-agent >= 6.13.0
 Requires: facter >= 3.14.2
 %endif
 Requires: jq
-Requires: cortx-motr = %{h_motr_version}
+Requires: cortx-motr
 Requires: cortx-py-utils
 %if %{rhel} < 8
 Requires: python36


rkothiya:
I created rpms for motr and hare and installed it. I was able to bootstrap the cluster : 

Below rpms were installed : 

```
# rpm -qa | grep cortx
cortx-py-utils-2.0.0-2_836c1a1.noarch
cortx-hare-2.0.0-1_gitc555bdb.el7.x86_64
cortx-motr-2.0.0-1_git106374f_3.10.0_1160.el7.x86_64
cortx-motr-debuginfo-2.0.0-1_git106374f_3.10.0_1160.el7.x86_64
cortx-hare-debuginfo-2.0.0-1_gitc555bdb.el7.x86_64
cortx-motr-devel-2.0.0-1_git106374f_3.10.0_1160.el7.x86_64
```
was able to start the cluster : 

```
# hctl bootstrap --mkfs /root/s.yaml
2022-07-05 03:58:23: Generating cluster configuration... OK
2022-07-05 03:58:24: Starting Consul server on this node........... OK
2022-07-05 03:58:33: Importing configuration into the KV store... OK
2022-07-05 03:58:34: Starting Consul on other nodes...Consul ready on all nodes
2022-07-05 03:58:34: Updating Consul configuration from the KV store... OK
2022-07-05 03:58:36: Waiting for the RC Leader to get elected........ OK
2022-07-05 03:58:42: Starting Motr (phase1, mkfs)... OK
2022-07-05 03:58:49: Starting Motr (phase1, m0d)... OK
2022-07-05 03:58:52: Starting Motr (phase2, mkfs)... OK
2022-07-05 03:59:04: Starting Motr (phase2, m0d)... OK
2022-07-05 03:59:08: Checking health of services... OK
```

hare.spec
```diff
@@ -56,7 +53,7 @@ Requires: puppet-agent >= 6.13.0
 Requires: facter >= 3.14.2
 %endif
 Requires: jq
-Requires: cortx-motr = %{h_motr_version}
+Requires: cortx-motr
```
SwapnilGaonkar7, line 56: So it means we can use any motr rpm versions with latest hare rpm.
How we will make sure whether motr rpm installed is compatible with latest Hare rpm?

vaibhavparatwar:
@SwapnilGaonkar7 @mssawant @d-nayak @Shreya-18 please review the PR

hare.spec
d-nayak: I have the same question that @SwapnilGaonkar7 does - this change removes all version checks between hare and motr and reduces it to checking whether motr is installed. Is that a safe assumption? Would it not break anything else?

mssawant: So with this change we can now build Hare with a different motr version and a different motr version can be installed. I understand this is okay for small changes in-order to save some build time. But If there are changes related to protocol or data structures for example, they could be different in Hare than underlying motr. Such cases, would further lead to failure. The version check was used to safe guard from the latter. Same is true for RGW. With disabling, now  it is possible that the we might hit some crash at some later point in time during execution.

vaibhavparatwar: @rkothiya @madhavemuri any comments on the concerns that @mssawant mentioned here?

madhavemuri: @mssawant : At ha interface level this check is already removed. If any API level changes happens we will update both hare and motr in the image.  This is not only for build purpose, we are removing dependency from both Hax and S3, assumption is that API's of libmotr will change rarely going forward and if that happens so which can be managed at release of CORTX image level.

d-nayak (approved):

